PR #18, opened 2021-10-03
fix: Allow VisualizerDifferences to work with OEIS sequences

### LINKED ISSUE #17: Differences visualizer does not work with OEIS sequences
To reproduce: With a running backend and frontend, go to the scope on frontend. Add any (working) OEIS sequence with the "Get Sequence" button, such as A000045 with 50 terms and modulus 1000. Bundle that with the Differences visualizer, and then click "Draw." The sequence values will display along the top of the page, but no differences will be shown.

On the other hand, if you bundle the "Naturals" sequence with the Differences visualizer, the differences are properly shown.

gwhitney:
The problem was that VisualizerDifferences was requesting far too many
  terms from the sequence, and so ran into trouble when there weren't more
  terms available (say if you hadn't loaded that many from the OEIS) and the
  unhandled errors terminated the operation of the visualizer.

  There is an underlying issue, however, that there is not a graceful
  convention for handling unavailable values in a sequence; that deserves
  a separate issue.

  Resolves #17.

### DIFF
--- a/src/visualizers/VisualizerDifferences.ts
+++ b/src/visualizers/VisualizerDifferences.ts
@@ -58,16 +58,11 @@ class VizDifferences extends VisualizerDefault implements VisualizerInterface {
 
 		const workingSequence = [];
 
-		const start = Number(this.settings.n) - Number(levels);
-		const end = Number(this.settings.n);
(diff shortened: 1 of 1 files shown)


liam-mulhall:
This fixes the problem described in #17. Merging.

PR #21, opened 2021-10-04
chore: Sequence factor

### LINKED ISSUE #13: Chore: code factoring in Sequence hierarchy
The code in src/sequences defines an interface, a SequenceClassDefault which is the most bare-bones implementation, stated as intending to be a a base class for other sequences, and then three implementations: SequenceConstant, SequenceNaturals, OEISSequenceTemplate. But none of these three actually derive from SequenceClassDefault, and there is duplicated caching code in the latter two. So there should be a SequenceCached which inherits from SequenceClassDefault, and the latter two should derive from that. This will make the code base more "DRY" (Don't Repeat Yourself), i.e. prevent us from having replicated code where a bug or enhancement might be made in one place but then missed in another duplicate place.

gwhitney:
Clean up the Sequence class hierarchy, so that all Sequence classes extend the SequenceClassDefault, and so that there is a single implementation of caching which all cached sequences share. This code factoring should make the code for future Sequence classes significantly more compact.

Resolves #13


src/sequences/SequenceConstant.ts
```diff
@@ -1,63 +1,53 @@
...
+    constructor(ID: number) {
+        super(ID);
     }
 
     validate(): ValidationStatus{
-		this.params.forEach(param => {
-            this.settings[param.name] = param.value;
-		});
+        this.settings['name'] = 'Constant';
+	const svs = super.validate();
```
liam-mulhall: Is `svs` a common abbreviation? I generally try to avoid short variables unless they're really common like `req` for request or `e` for error.

src/visualizers/visualizers.ts
```diff
@@ -1,5 +1,5 @@
 const vizFiles = require.context('.', false, /\.ts$/)
-const vizMODULES: { [key: string]: Function } = {};
+const vizMODULES: { [key: string]: __WebpackModuleApi.RequireContext } = {};
```
liam-mulhall, line 2: I don't know what's going on here. Please explain, @gwhitney.

src/sequences/SequenceConstant.ts
gwhitney: It was short for "super validation status" but your feedback is excellent. I will fix in a further commit shortly.

src/visualizers/visualizers.ts
gwhitney: I will comment this in the upcoming commit. Again, thanks for the excellent feedback.

gwhitney:
OK, the revised PR should address all of the comments above. Thanks again for the feedback!

liam-mulhall (approved):
@gwhitney, I have no further comments. I did some manual testing, and I think this should be merged.

PR #37, opened 2021-10-23
feat: automatically activate a sequence when added

### LINKED ISSUE #35: When you do "get sequence" in the main tool, it should become the active sequence
Currently, you have to click on "Get sequence," enter the OEIS number, and then a button for the sequence will pop up, which you then have to click to make it the active sequence. But it's completely just more convenient for the user to immediately make the new sequence the active sequence: it saves a click, and there's no downside because users can always just click on some other sequence if for some reason they don't want it to be active (which they probably do).

gwhitney:
Resolves #35.

### DIFF
--- a/src/views/ToolMain.vue
+++ b/src/views/ToolMain.vue
@@ -71,6 +71,7 @@ export default Vue.extend({
         true
       )
       this.sequences.push(seqBundle);
+      this.setActiveSeq(newOeisSeq);
     },
     bundleSeqVizPair: function() {
         const bundle = {


liam-mulhall:
@gwhitney, I can merge this PR, but like you said at last week's meeting, if there's anyone else who wants some practice, please feel free to merge it. If anyone needs a primer, [this video](https://www.youtube.com/watch?v=rgbCcBNZcdQ) provides an overview of creating and merging a PR.

HristopoulosSt (approved):
This change looks good, apologies for the late review.

Issue #23, opened 2021-10-08
Allow formulas to invoke OEIS sequences


gwhitney:
The idea is to be able to visualize something like A000045(n) - A000045(n-1) - A000045(n-2) (and see in this case that it's all zeros ;-) in a relatively straightforward way.

Depends on #14.

gwhitney:
Although #14 is closed, I feel this now also depends on #42.

gwhitney:
To summarize the current plan: based on the discussion in #48, sequences need to return their first and last valid indices. And as noted in the discussion on #74, since that's potentially very difficult (i.e., insoluble in general like the halting problem) to figure out by formula analysis, we will do the computation in the backend. So the plan is to use math.js just to parse the formulas into a canonical string form (RPN with subexpressions in lexicographic order, I guess) and pass that to the backend, which will actually do the computation. An advantage there is that Python already does arbitrary-precision integer arithmetic, so that is a win. So we shouldn't really have to worry about adding bigints to math.js, after all. My tentative implementation plan is to:
1) Make a version of frontscope that produces canonical RPN formulas, that I won't commit (as it won't really add any functionality). Note the formulas need to use only URL-acceptable characters.
2) Make a version of backscope that can accept such formulas and produce output like the sequence endpoint does, and commit that to main.
3) Extend (1) to use the results from backscope to fully implement an OEIS-with-formulas Sequence class.

katestange:
One comment:  I don't know if it saves us doing bigints in math.js, as this is something visualizers will want, potentially (they may do all kinds of processing on the sequences they get in bigint form).

gwhitney:
That's a good point, Kate. I guess how that's handled is partly up to whoever implements #48 and how that person goes about it.

While I am commenting, I am planning (eventually) to allow a formula writer optionally to indicate sequence limits via an inequality expression in the formula, something like:
`floor(sqrt(A000001(2*n)); 2 <= n <= 512`
I am leaning toward this for a couple of reasons:
* I would like formulas to remain self-contained; i.e, a formula-based sequence should be completely describable by a single string, without multiple entry boxes. This desire is mainly motivated by the possibility that the sequences area could be replaced  just by a panel of text boxes, each containing a formula. (And eventually those formulas could display in typeset math notation!)
* It also provides a nice place to specify the index variable, in case we ever want to get away from requiring that it be n.
[...]

PR #43, opened 2021-11-15
refactor: Delineate three kinds of Sequence items

gwhitney:
This code is a step toward resolving #23 and does not add new functionality. 

  Rather, it makes explicit the fact that there are three kinds of Sequence items:
  Sequence "Getter"s that have a button at the bottom of the sequence menu,
  which when used produces a new button for the specific sequence created;
  Sequence "Family"s that have a button in the top of the sequence menu, but
  need parameters to create a specific sequence each time their button is
  pressed; and Sequence "Instance"s that were created by a "Getter" and which
  select that one specific sequence when their button is pressed.

  Formerly, which one of these kinds a Sequence functioned as was an indirect
  result of whether it had a SequenceExportModule and whether or not "isOeis"
  was true in that module; also, the button for the only "Getter" in practice,
  the OEIS Sequence Template, was hard-coded.

  This PR creates an explicit enumeration of the three kinds, and requires that
  every Sequence object has a SequenceExportModule to show up in the Menu, and
  the kind must be explicitly set in that module.

  The motivation for this refactoring (besides code clarity and the removal
  of hard-coding for the OEIS template) is to allow for additional Getters
  besides the OEIS template, whose buttons will be automatically displayed
  alongside the OEIS one at the bottom of the Sequence menu. The idea is that
  the Sequence type to come that allows mixed formulas and OEIS sequences
  will be such a Getter.

### DIFF
--- a/src/components/SeqGetter.vue
+++ b/src/components/SeqGetter.vue
@@ -1,6 +1,6 @@
 <template>
     <a v-on:click="$emit('load-seq')">
-        <li class="list-group-item">Get Sequence</li>
+        <li class="list-group-item">{{ title }}</li>
     </a>
 </template>
 
@@ -8,7 +8,8 @@
 export default {
     name: 'SeqGetter',
     props: {
-        oeisId: String
+        title: String,
+        instanceId: String
     }
 }
 </script>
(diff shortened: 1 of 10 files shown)


katestange:
This sounds very sensible to me.

liam-mulhall (approved):
@gwhitney, I reviewed the files and did some manual QA testing. The code looks reasonable and I didn't see any issues during testing.

Issue #48, opened 2021-11-21
Capability for inbuilt math functions, with test being `mathematician's modulus'


katestange:
We should create a place in the codebase where we can build a library of inbuilt math functions available to visualizers.  So, housed in src/visualizers, e.g. src/visualizer/math.ts.  (Glen notes: the file name starts with a lower case by the conventions of this project because it is not defining a class.)  This file should export various useful functions.  The first test case, to include when setting this up, is a ''mathematician's modulus''.  The `%` operator in javascript will return a negative value when the dividend is negative.  Ideally, (A mod B) should return the smallest non-negative integer which differs from A by a multiple of B.  So, (-13 mod 5) should be 2 (not -3 as javascript would have it).  In particular, the result should be in the range 0 to B-1 inclusive.  The function itself would be called in the visualizer by importing modulus from math.ts, and then have syntax like modulus(A,B).

gwhitney:
Note that the return type of a Sequence element has turned out to be `bigint|undefined` (so that missing terms can be flagged) so these utility functions need to handle inputs of that union, and just return undefined if any necessary input is undefined.

gwhitney:
After some reflection on the experience of creating a module like this in #50, I would like to suggest that ultimately we fold these utilities into math.js, which we are already using. That way, anything we add can be uniformly used in formulas by users in the front-end and by visualizers in the back end, for example the powmod() function or whatever we decide to call it that computes a^k mod n. That doesn't mean we shouldn't put things in a separate math utility module for now, but it does suggest that we should as soon as feasible start using math.js for computations in visualizers. That will make things less ad hoc. (In particular, math.js modulus follows mathematicians' rules, so that solves that.) But getting that to work depends on extending math.js to handle items that are `bigint|undefined` so that will be my next effort as soon as #50 is resolved.

gwhitney:
More details on the above comment, quoted here from Slack (because of the discussion in https://numberscopef21.slack.com/archives/C02D6G6TCF8/p1637619803031500 to move significant design conversations to Github):

[...]

PR #50, opened 2021-11-22
WIP refactor proposal A: Systematize Sequence objects to return BigInt | undefined

### LINKED ISSUE #42: Inconsistent numeric types in certain sequences
The SequenceInterface says that getElement returns "number," but on the other hand, OEISSequenceTemplate fills the cache which produces values for getElement with "BigInt"s. (I am not sure what TypeScript's exact typing rules are for numbers here, so I am unclear on why TypeScript does not report any type error or warning for the code that puts BigInts into a cache declared to contain numbers.)

Anyhow, this mismatch causes some visualizers, e.g. ShiftCompare, to fail on OEIS sequences; they contain code that performs computations (such as the modulo operator) which does not allow mixing of BigInts and numbers.

We need to decide on a policy of how to represent numbers for sequences in Numberscope, and then use it uniformly.

gwhitney:
Resolves #42.

  Note this also creates the math.ts collection of functions for use by visualizer writers as called for in #48. It turns out Sequences in fact have to return the union `bigint | undefined` because otherwise there is no way to flag nonexistent sequence values. That makes arithmetic on the return values kind of a pain. Since we are going to have a math module anyway, though, we should just put in there utility functions that handle undefined values in the straightforward way of making results undefined whenever any inputs are undefined.

### DIFF
--- a/src/sequences/SequenceCached.ts
+++ b/src/sequences/SequenceCached.ts
@@ -12,7 +12,7 @@ export class SequenceCached extends SequenceClassDefault {
     name = 'Cached Base';
     description = 'A base class for cached sequences';
     entries = 0;
-    protected cache: number[];
+    protected cache: bigint[];
     protected newSize = 1;
 
     /**
@@ -60,15 +60,17 @@ export class SequenceCached extends SequenceClassDefault {
 
     fillCache(): void {
         for (let i: number = this.cache.length; i < this.newSize; i++) {
-            this.cache[i] = this.calculate(i);
+            const val = this.calculate(i);
+            /* Would be type error to store in cache if undefined */
+            if (typeof val !== 'undefined') this.cache[i] = val;
         }
     }
 
-    getElement(n: number): number {
+    getElement(n: number): bigint | undefined {
         if (this.entries > 0 && n >= this.entries) {
-            return Number.NaN;
+            return;
         }
-	if (n >= this.cache.length) {
+        if (n >= this.cache.length) {
             this.resizeCache(n);
             if (this.newSize > this.cache.length) {
                 this.fillCache();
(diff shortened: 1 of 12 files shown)


gwhitney:
In particular, all visualizers should now work with all sequences, but beware that ShiftCompare can be very slow -- try it on a sequence that does not have very many terms.

gwhitney:
Actually, if anyone is looking to review this, please first look at the latest discussion on slack in thread https://numberscopef21.slack.com/archives/C02H8SRCDBM/p1637456846026000 and provide your thoughts -- there's a potential alternate plan that might be better, which would lead me to retract this for an alternate PR. So don't review this until I comment again based on the responses in that Slack thread.

liam-mulhall:
Does our Slack plan persist conversations? Or will they go away at some point? If they do eventually go away, we might want to rethink having potentially illuminating conversations on Slack and have them on GitHub instead.

gwhitney:
Excellent question -- you might want to ask Kate on the slack #general channel, since presumably she set Slack up.

gwhitney:
OK, the conversation on this has been moved to #48, please comment there (rather than on Slack).

gwhitney:
Note that for comparison's sake, I have submitted PR #51 to show how things would work out in the alternate plan described in #48. Neither this PR or that one should be reviewed/merged until a consensus has been reached in #48 as to which way to go. Please see and comment on the discussion there. Thanks!

katestange:
Glen pointed out a sequence could have a very complex formula involving several OEIS tags, some math functions, etc.  It's possible this sequence itself could also take a long time to compute, right?  So we encounter something a bit like factoring sequences:  we may want data that takes a while to compute.  If we go compute the whole sequence and store it in the database, then have the visualizer pull from the database-stored version, we won't have any issues with valid range of indices, and could implement #51 (unless the sequence is "spotty" and has gaps?).  So it's a larger question:  how do we want to treat formulaic sequences?  Do we want to precompute them or attempt to compute them as indices are requested?  There are problems with computing on the fly anyway, irregardless of #50 vs. #51.

gwhitney:
See the discussion in #48; decided against this implementation route. Closing.

PR #53, opened 2021-12-03
feat(Sequences): add random sequence generator

### LINKED ISSUE #52: Random integer sequence
A sequence module that generates a sequence of random integers chosen uniformly from the range from n to m, where n and m can be specified by the user.  The sequence shouldn't change while a visualizer is using it, i.e. if the 9th term is 4, it should stay 4 if that term is called again.  But each time the user initializes such a sequence, it is independent from any previous such sequence created.  This turns out to be useful in some of the visualizers I'm thinking about or writing, e.g. the chaos game, where you want to be able to compare to what "random" looks like.

katestange:
closes #52

### DIFF
--- a/src/sequences/SequenceRandom.ts
+++ b/src/sequences/SequenceRandom.ts
@@ -5,79 +5,73 @@ import { SequenceCached } from './SequenceCached';
 
 /**
  *
- * @class SequenceClassNaturals
- * extends SequenceCached with a very simple calculate, mostly by way of an
- * example of using SequenceCached.
+ * @class SequenceClassRandom
+ * Creates a sequence of random integers in a specified range.
+ *
  */
 class SequenceRandom extends SequenceCached {
-    name = "Random Integers in Range";
-    description = "A sequence of integers chosen indepenently uniformly from n to m inclusive.";
-    params: SequenceParamsSchema[] = [new SequenceParamsSchema(
-        'min',
-        'number',
-        'Minimum value attainable',
-        false,
-        0 
-    ),
-    new SequenceParamsSchema(
-        'max',
-        'number',
-        'Maximum value attainable',
-        false,
-        9
-    )];
+	name = "Random Integers in Range";
+	description = "A sequence of integers chosen indepenently uniformly from n to m inclusive.";
+	params: SequenceParamsSchema[] = [new SequenceParamsSchema(
+		'min',
+		'number',
+		'Minimum value attainable',
+		false,
+		0 
+	),
+	new SequenceParamsSchema(
+		'max',
+		'number',
+		'Maximum value attainable',
+		false,
+		9
+	)];
 
-    private minimum = 0;
-    private maximum = 0;
-    private servedTerms: number[] = [];
+	private minimum = 0;
+	private maximum = 0;
 
-    /**
-     *Creates an instance of SequenceRandom
-     * @param {*} ID the ID of the sequence
-     */
-    constructor (ID: number) {
-        super(ID);
-    }
-
-    validate(){
-        this.settings['name'] = 'Constant';
-        const superStatus = super.validate();
-        if (!superStatus.isValid) {
-            return superStatus;
-        }
-
-	if (this.settings['min'] == undefined || this.settings['max'] == undefined ) {
-		this.isValid = false;
-		return new ValidationStatus(false, ["The min or max parameter is missing."]);
+	/**
+     	*Creates an instance of SequenceRandom
+     	* @param {*} ID the ID of the sequence
+	*/
+	constructor (ID: number) {
+		super(ID);
 	}
 
-	if ( !Number.isInteger( Number(this.settings.min) ) ) return new ValidationStatus(false, ["The min value must be an integer."]);
-	if ( !Number.isInteger( Number(this.settings.max) ) ) return new ValidationStatus(false, ["The max value must be an integer."]);
+	validate(){
+		this.settings['name'] = 'Constant';
+		const superStatus = super.validate();
+		if (!superStatus.isValid) {
+			return superStatus;
+		}
 
-	this.isValid = true;
-	this.name = "Random integers from " + String(Number(this.settings.min)) + " to " + String(Number(this.settings.max));
+		if (this.settings['min'] == undefined || this.settings['max'] == undefined ) {
+			this.isValid = false;
+			return new ValidationStatus(false, ["The min or max parameter is missing."]);
+		}
 
-	return new ValidationStatus(true);
+		if ( !Number.isInteger( Number(this.settings.min) ) ) return new ValidationStatus(false, ["The min value must be an integer."]);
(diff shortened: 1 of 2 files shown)


gwhitney:
I'll look at this as soon as I can if nobody else gets to it first. One tiny note: pull requests are generally by convention titled with what change they make, in the present tense, rather than by what "happened" in the past tense. Also, I find the "semantic commit messages" convention very helpful (see https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716). With those points in mind, I took the liberty of mildly renaming this PR.

katestange:
Thanks!

gwhitney:
OK, I've done the code review and have some suggested changes above. I'll continue the review/check-in process when those things are resolved. But definitely a nice addition to Numberscope!

gwhitney:
Oops, sorry, didn't realize I had to hit the "submit review" button for you to see the comments. They should be there now.

src/sequences/SequenceRandom.ts
```diff
@@ -0,0 +1,83 @@
...
+	if (this.settings['min'] == undefined || this.settings['max'] == undefined ) {
+		this.isValid = false;
+		return new ValidationStatus(false, ["The min or max parameter is missing."]);
+	}
+
+	if ( !Number.isInteger( Number(this.settings.min) ) ) return new ValidationStatus(false, ["The min value must be an integer."]);
+	if ( !Number.isInteger( Number(this.settings.max) ) ) return new ValidationStatus(false, ["The max value must be an integer."]);
+
+	this.isValid = true;
+	this.name = "Random integers from " + String(Number(this.settings.min)) + " to " + String(Number(this.settings.max));
```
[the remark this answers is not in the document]
katestange: It fits ok in a browser (one line, within the provided box).  When we revisit UI it may need revisiting.

src/sequences/SequenceRandom.ts
```diff
@@ -0,0 +1,83 @@
...
+	return new ValidationStatus(true);
+
+    }
+
+    calculate(n: number) {
+	this.minimum = Number(this.settings.min);
+	this.maximum = Number(this.settings.max);
+
+	console.log( "term", this.servedTerms[n] );
+	if ( isNaN(this.servedTerms[n])  ) {
```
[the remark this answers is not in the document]
katestange: It seems to work, yes.  But it requires a specific guarantee from the caching functionality that it won't recalculate terms ever.  So I've added an issue to add that as a test once we have a testing framework.

katestange:
Ok, ready for review again.

src/sequences/SequenceRandom.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		if (this.settings['min'] == undefined || this.settings['max'] == undefined ) {
+			this.isValid = false;
+			return new ValidationStatus(false, ["The min or max parameter is missing."]);
+		}
+
+		if ( !Number.isInteger( Number(this.settings.min) ) ) return new ValidationStatus(false, ["The min value must be an integer."]);
+		if ( !Number.isInteger( Number(this.settings.max) ) ) return new ValidationStatus(false, ["The max value must be an integer."]);
+
```
gwhitney, line 55: It's an unfortunate design flaw in the implementation of SequenceClassDefault that if either of the above cases are hit, `this.isValid` will be true: it's set to true in SequenceClassDefault.validate() and not reset to false in either of these cases, unlike in the cases where 'min' or 'max' is undefined. I'll create an issue for this design flaw, but in the meantime, the expedient thing to do for this PR would be to set this.isValid to false immediately after checking the return from super.validate() like at line 47. Then you can go through the bad conditions, with the isValid flag safely false if any of them are hit (i.e. you can then remove line 49) , and then you already set it back to true below once you've gotten through all the bad conditions, so it will be OK if there are no bad conditions. Hope I am making sense.

src/sequences/SequenceRandom.ts
```diff
@@ -0,0 +1,77 @@
...
+		this.name = "Random integers " + String(Number(this.settings.min)) + " to " + String(Number(this.settings.max));
+
+		return new ValidationStatus(true);
+
+	}
+
+	calculate(n: number) {
+		this.minimum = Number(this.settings.min);
+		this.maximum = Number(this.settings.max);
+
```
gwhitney, line 66: You should set `this.minimum` and `this.maximum` once in validate(), where in fact you are already doing the conversion to Number anyway, rather than re-doing the conversion on every call to calculate().

src/sequences/SequenceRandom.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ */
+class SequenceRandom extends SequenceCached {
+	name = "Random Integers in Range";
+	description = "A sequence of integers chosen indepenently uniformly from n to m inclusive.";
+	params: SequenceParamsSchema[] = [new SequenceParamsSchema(
+		'min',
+		'number',
+		'Minimum value attainable',
+		false,
```
gwhitney, line 19: This is the argument that specifies whether the parameter `min` is required. Looking at validate() below, clearly it is required. Hence this should be true, and the same goes for `max` below. I will also file an issue that if a parameter has been marked as required, the Sequence writer should not also have to check it manually in validate(). (Right now this parameter and the code in validate() are partially redundant.)

gwhitney (commented):
My apologies that my first review wasn't thorough enough. On closer inspection, uncovered a couple more concerns, most of which actually stem from minor design flaws in the Sequence hierarchy that I will file issues about. But they do need fixing here for now for this to go in, since I don't think it makes sense to wait for them to be fixed in general.

src/sequences/SequenceRandom.ts
katestange: Makes sense. General question:  where does one do general setup for a more general sequence?  validate() seems too special-purpose, at least based on its name, but sequences don't have a general setup() function that would only get called once.  Should that be an issue to add?

katestange:
Just pushed again.  Fixed the above, also caught a naming error in line 42, added something to suppress a lint error because n is not used in calculate() (line 65).

src/sequences/SequenceRandom.ts
gwhitney: There is already an initialize() method in SequenceInterface that is guaranteed to be called once per sequence. But it is called _after_ validate(), and by default checks that validation succeeded. But in the case of your two parameters, they already have to be parsed in validate(), so there is no reason not to set the object properties right in validate() as you have done. On the other hand, I agree, if you had to say pre-compute some big matrix or something like that one time before calculate() (or getElement(), in case of an uncached Sequence) was called, you would do that in initialize() rather than validate().

gwhitney:
(Just got rid of trailing whitespace, mixed tabs/spaces, and a line with only whitespace. All of these things are highlighted by "git diff main", which you should always do before submitting a PR.)

gwhitney:
Since the integer 0 is a "falsy" value, the default for the min of the range was not showing up pre-filled in the sequence parameter dialog. Fixed this by changing it to the one-character string `'0'` which is a "truthy" value.

Issue #71, opened 2021-12-18
Adopt a framework for documentation


gwhitney:
From the conversation on #69, there's clearly a need for a place to put such information as the code style guide, which is of documentary interest both to the core team and to potential new contributors. We also have a stub of a user guide integrated into numberscope itself, which includes, for example, usage documentation on each Visualizer (except some are stub or missing). To satisfy these and other documentation needs, the project should adopt a framework for where documentation will be kept and how it will be organized.

gwhitney:
I absolutely agree with the stub design that (at least a portionf of) the documentation should be available right from the Numberscope itself. So one basic question: (A) should the documentation so accessible comprise all documentation, including for example the contributing guidelines and coding style guide? Or should there be a user guide limited just to how to make nice pictures with Numberscope and maybe the theory behind the various visualizers, with more technical/code information available "somewhere else"?

gwhitney:
The next basic question is (B) how should we capture that documentation? Should it be in the main repository itself, or (some or all of it) be in a place like a GitHub wiki? In the former case, what format should the documentation be in?

gwhitney:
And the third question is (C) how should the documentation be provided to interested parties? Again, the answer(s) could depend on the answers to (A) and (B). For example, if some goes into a wiki, then that content would be served by GitHub's wiki feature.

gwhitney:
[...]

Issue #69, opened 2021-12-18
Make Code Style Guide for Numberscope


liammulh:
@katestange, @gwhitney, before I add the Prettier code formatter, it would be useful to get your opinions on what our TypeScript code style / code format should be going forward.

First, I highly recommend skimming Idiomatic JavaScript (https://github.com/rwaldron/idiomatic.js/), a fairly popular guide for JavaScript code style. We obviously don't need to agree with all of their rules, but I think they have a good approach/philosophy when it comes to code style and code format.

Second, would you please comment as to whether your text editor or IDE has a plugin for Prettier? See their supported text editors and IDE's here: https://prettier.io/docs/en/editors.html.

If one of you uses a text editor that isn't supported, we could try using a combination of https://editorconfig.org/ and ESLint rules. Otherwise, to familiarize yourselves with some of the options Prettier provides, please click this link: https://prettier.io/playground/.

Some of the big things we need to decide on:

1. still going with 80-character limit?
2. tabs or spaces? width?
3. single quote or double quote?
4. trailing comma?
```js
const myTrailingCommaObj = {
  a: 1,
  b: 2,
  c: 3,
}

const myRegularObj = {
  a: 1,
  b: 2,
  c: 3
}
```
5. semicolons?
6. conditional style?
```js
if(condition) doSomething();

if(condition) {
  doSomething();
}
```
7. do we want to abbreviate variable names or be verbose?

I'm sure there are other things as well. If you can think of any, please write them in this issue. If we can't use Prettier to enforce some things, I bet we can find an ESLint rule.

Alternatively, we could just use ESLint rules for formatting, but the downside to this is that you wouldn't be able to use your editor to "autoformat" your code, which I find saves a lot of time.

Finally, I think we should create a living code style guide that we can add to / subtract from as needed.

liam-mulhall:
We also need to agree on a style for conditionals. For example, [see this PR conversation](https://github.com/numberscope/frontscope/pull/51/files/ab053262e5030fab31fcfb9613f6bf3476d0bba6#diff-52110c64f50a9009e3b18d7c6507ba770c193c70c7ee04cfe3b7ad1b08ee8d6c) between me and Glen.

gwhitney:
I use emacs, which has a prettier mode, so we are cool there. I will read your recommended references and weigh in on the issues you raise. Where should the code style guide live? In the docs within the repository, or here in the repository manager (github), say as part of the PR guidelines?

[...]

PR #38, opened 2021-10-29
[WIP] Set Up Jest

liam-mulhall:
In this PR, we install Jest, configure it for TypeScript, and add docs for unit tests and integration tests.

PR #72, opened 2021-12-19
Add Vue Test Utils

liam-mulhall:
This commit adds Vue Test Utils, which is "a set of utility
functions aimed to simplify testing Vue.js components". It
allows us to unit test Vue components.

I was originally trying to set up the Jest testing framework for general testing purposes, but I was running into lots of issues. My plan is to further configure Jest for general testing purposes now that I've got Vue Test Utils working.

To test this PR:

1. `npm run test:unit` should output something like the following
```
> newnumberscope@0.2.0 test:unit
> vue-cli-service test:unit

 PASS  tests/unit/SeqGetter.spec.ts
  ✓ displays dummy title (21ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        3.828s
Ran all test suites.
```
2. `npm run serve` should compile successfully
3. you should be able to interact with Numberscope in your browser as usual


gwhitney:
A few comments/questions:
1) How does this interact with PR #38? At the very least, if they were both merged, we would have `npm run test` running whatever TypeScript tests we set up in jest and `npm run test:unit` running Vue tests. Those two commands don't seem parallel or mnemonic for the two alternatives. Should this one be `npm run test:vue` and the other be modified to `npm runt test:ts` or something like that? Please just make sure you are heading to a rational collection of test directives, thanks.
2) I see that you've (whitespace-)reformatted the list of parameters in .eslintrc.js, as well as added the new "overrides" parameter. Some projects prefer to minimize the lines touched in functional PRs, and leave such reformatting to format-only PRs. I don't myself much care especially for a fairly lean PR like this one, but if you would prefer that approach that separates functional PRs from format PRs, you might want to leave the spacing alone in this PR.
3) I see the one test for SeqGetter just checks that it displays the title. But somehow the "essence" of a SeqGetter seems more that it emits a `load-seq` signal when it is clicked. Is that a sort of thing that can be tested in this framework? If so, could you add a test for that -- mostly because it would make your one example test a better guide for folks who will have to write tests for other Components (which will presumably become mandatory for future PRs).
4) Your Jest PR #38 in progress added a documentation file about testing. Even though we haven't resolved #71, should this PR have something like that for Vue testing, in some reasonable place that can be rearranged depending on how #71 turns out? If so, that file should likely also mention that providing a working, reasonably complete unit test for all affected Vue components is mandatory for all future PRs that touch a Vue component.
5) What about testing views? Is that doable with this framework, or does it only test components? Some mention of that point would be nice somewhere in this PR, for example in the doc file if you add one, or in some other reasonable place if not.
I am going to unassign myself because really once there are responses to these questions, anyone could move the PR on from there.

liam-mulhall:
1. This morning, I initially set out to work on https://github.com/numberscope/frontscope/pull/38, but as I mentioned, I was running into too many issues. The plan is to essentially scrap the work done for that PR and instead build off of this PR. I can delete https://github.com/numberscope/frontscope/pull/38 once I'm sure there isn't anything useful in it. As you suggest, I was planning on doing something like `npm test:vue` and `npm test:ts` for the two different types of testing. I think the exact names for the test directives can be decided upon later. I'm just trying to get things working at this point.
2. That formatting wasn't added by me; it was done automatically when installing Vue Test Utils. I can remove it.
3. I can add a more comprehensive test for `SeqGetter`, but that wasn't really the point of this PR. The point of this PR is to add Vue Test Utils. I only added the test so that we'd have a means to ensure Vue Test Utils is working properly. I just wanted to add a "Hello, World!" for Vue Test Utils. Would you like me to add a more comprehensive test for `SeqGetter` in this PR, @gwhitney?
4. I'd like to add a file `doc/tests.md` that goes over how to add a test for Vue components and how to add a test for TS once we're able to do both of those things and once things like the names for the test directives have been decided upon. In other words, once things are more settled, I will add that doc.
5. Yes. See https://github.com/liam-mulhall/frontscope/commit/8bda358c6ffa0edf1816a4a53a8549087238488f. I can add this to the doc.

gwhitney:
State of the items below. As far as I am concerned, all that remains -- except I feel a little stronger than that about the new (6) -- are optional suggestions. Just let me know whether or not you're going to take any action with respect to them, and then I'll wrap up this PR and merge it.
1) Thanks for explaining that #38 is to be superseded. As to the name of the directive to run the test, I totally agree that we can't expect - and shouldn't attempt to force - the specifics of any PR to be the permanent state of affairs. That said, I think it's good practice all else being equal to have as many aspects of a PR as practical be your current best guess of the final state of things. So if the name "test:unit" seems unlikely to be the final directive, I'd suggest you change it in this PR to your best current guess of what that final directive should be.

2) Fully resolved.

3) I understand the point of this PR wasn't to test SeqGetter in particular, but of course adding a testing framework has to include at least a test or two. The addition of the Home test is nice, and that test is pretty much comprehensive; not sure what else you would test about the Home view. So it seems to me that it would make a better introduction, and example for future PRs to follow, to start with each of your initial tests being essentially comprehensive for the components you chose as "guinea pigs." That will have the side effect of demonstrating a slightly larger range of testing techniques, as well. And it doesn't look like much more work, since as far as I can see, there's really only one other thing to test about a SeqGetter. And if it isn't done at this point, when the SeqGetter test is being introduced, when will that test be made comprehensive? Only when somebody touches SeqGetter.vue, and then only if the PR writer or reviewer notice/remember at that juncture that the SeqGetter test isn't comprehensive. So I'd suggest that this is the time to make a comprehensive SeqGetter test; but as I said it's just a suggestion and once you respond I will follow through with this PR either way.

4) You might want to file an issue immediately about the need for some doc on the testing framework, if you're not including even a stub of that in this PR.

5) Fully resolved.

6) Why did you delete the final newline of tsconfig.json? It doesn't seem relevant to the content changes you made to the file, and as far as I am concerned every line of a source file should end in newline, and in any case we definitely need to not get into a cycle of going back and forth on final newlines -- this is an item that should be standardized and be in the eventual document produced by #69. I will add a comment there, and for now I'd recommend you leave alone the final newline that's already in this file.

liam-mulhall:
1. My best guess right now is `npm run test:vue`. I will rename the script. (The command `vue-cli-service test:unit` can't be easily renamed, so I'll leave that as is.) Done in https://github.com/numberscope/frontscope/pull/72/commits/98a41622df6ece4a1ae8efd2925061bcfe0c32f0.
2. Resolved.
3. Added additional test for `SeqGetter` that ensures a `load-seq` is emitted when clicked in https://github.com/numberscope/frontscope/pull/72/commits/4bf71ae0dc60387f72386e157068be2221933613.
4. Done in https://github.com/numberscope/frontscope/issues/73.
5. Resolved.
6. I've never faced an issue where not having a newline at the end of a file has presented issues in the web dev projects I've worked on, but I'm happy to follow best practices here. Newline added in https://github.com/numberscope/frontscope/pull/72/commits/704d1a35d8219a83ae66d05987704d4b659733cf.

gwhitney:
All looks good. I will merge, and consider supplying tests for Vue components and views that are touched to be mandatory going forward (that should of course go in our submission documentation).

On (6), I fear you mistook the source of the strength of my recommendation. I am not terribly concerned about final newlines; note I did not even previously comment on the new source file you added that lacks one. I am just concerned about going back and forth on their presence. (My editor automatically adds them, yours clearly doesn't, and we don't want git history muddled up with lots of changes back and forth on this.) Hence, the thing I am concerned with is that at some point fairly soon, we pick one convention or the other, do a one-time update to uniformize, and then stick with that convention. In the meantime, we shouldn't unnecessarily change any particular source file in this respect.

liam-mulhall:
Hey, no worries. I like to pick a convention and stick with it. I'm sure I can get at least an ESLint rule to enforce newlines at the end of files. Maybe Prettier will have an option. I can also configure my various editors to do it as well. Anyway, we have lots of tools at our disposal.

liam-mulhall:
I think making tests for a touched file mandatory is a good idea.

PR #77, opened 2021-12-20
style(Home): Remove prototype Sequence Playground page.

### LINKED ISSUE #33: Remove "Sequence Playground" from UI?
At a group meeting a couple of weeks ago, the consensus seemed to be that this was sort of a prototype to see if the front end could read sequences before there were visualizers to look at them with. So probably this section of the UI, which doesn't really seem to add any functionality, should seemly be retired and removed. But this should be explicitly agreed upon by the group before proceeding.

gwhitney:
Also add a test the NavBar view, which must be modified to remove
  Sequence Playground.

  Resolves #33.


liam-mulhall (approved):
Looks good, @gwhitney!

PR #80, opened 2021-12-22
fix(OEISSequenceTemplate): respect OEIS offset

### LINKED ISSUE #20: How to handle OEIS sequence "offsets"?
Of course, every OEIS sequence boils down to a list of numbers. The current numberscope code as of this filing treats every such sequence _s_ as "zero-based"; in other words _s(0)_ is the first number in the list, _s(1)_ the next, and so on.

However, the OEIS lists an "offset" _t_ for each sequence such that nominally, the first item in the list is conceptually _s(t)_, the second item in the list is conceptually _s(t+1)_, etc. Should numberscope respect these offsets? If so, what would that look like? if _t>0_ should the elements of index less than _t_ show up as "unavailable" or "nonexistent"? How will that affect visualizers? (For these reasons, design work on this issue should like also address #19.) Similarly, for negative offsets, should we end up ignoring initial elements of the sequence? Or should there be a way to allow visualizers to take action on elements at negative indices as well?

This is a longer-term design issue, without a clear quick fix.

### LINKED ISSUE #70: Refactor Code to Use async/await
There are plenty of instances of old promise syntax throughout the code base. This is the old way of doing asynchronous JS. It's harder to read than the new async/await syntax. We should be getting rid of the promise syntax and using async/await as we move forward.

If you aren't familiar with the various ways to do asynchronous JS, check out [this excellent blog](https://medium.com/front-end-weekly/callbacks-promises-and-async-await-ad4756e01d90). It does a good job of summarizing the evolution of asynchronous JS.

gwhitney:
Changes the route to backend used to obtain OEIS sequence values to one
  that includes the original OEIS index information based on the offset.
  Sets the first and last indices of the resulting Sequence object to
  match the data from the OEIS.
  Also strengthens the error checking on the modulo and numElements parameters
  of the OEISSequenceTemplate.

  Resolves #20.
  Resolves #70.

### DIFF
--- a/src/sequences/OEISSequenceTemplate.ts
+++ b/src/sequences/OEISSequenceTemplate.ts
@@ -16,6 +16,9 @@ export default class OEISSequenceTemplate extends SequenceCached {
     name = 'OEIS Sequence Template';
     description = 'Factory for obtaining sequences from the OEIS';
     oeisSeq = true;
+    modulo = 0n;
+    cacheBlock = 1000;
+
     params: SequenceParamsSchema[] = [
         new SequenceParamsSchema('oeisId', 'text', 'OEIS ID', true, false),
         new SequenceParamsSchema('name', 'text', 'Name', false, false),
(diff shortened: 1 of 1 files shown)


liam-mulhall:
I tested locally and it seems to be working.

PR #88, opened 2022-01-07
chore: Add Our Own Favicon

liam-mulhall:
Add our own favicon (the little icon you can see in browser tabs). It used to be the default Vue favicon. To test that it's working properly, `npm run serve`.

You shouldn't see:
old

You should see:
new

This was just something that was bugging me a bit that I wanted to change.


gwhitney:
test:vue passes, numberscope runs, favicon is changed. The new favicon is a bit dark/hard to make out much detail from, but anything is better than using the generic Vue favicon, so merging.

PR #89, opened 2022-01-07
chore: Rename Title

liam-mulhall:
We rename the title from "newnumberscope" to "Numberscope". The old
title was obtained from the "name" field in `package.json`. I'm not a
fan of this method since the "name" field in `package.json` is supposed
to be lowercase. Thus, we change the "name" in `package.json` to be
"frontscope" to match the repo's name.

Related to https://github.com/numberscope/frontscope/pull/88 in that
it's a minor cosmetic thing that will make the frontend look a bit
nicer.

Here's how the title should look now:
new-title

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "newnumberscope",
+  "name": "frontscope",
   "version": "0.2.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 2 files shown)


gwhitney:
npm run test:vue passes, npm run serve executes without errors or warnings, tab is now titled Numberscope, and the numberscope seems to work properly. Definitely "Numberscope" is an improved tab title as compared to "newnumberscope". I fear that in the future if the name of the app appears many places we will have to change to one authoritative place where it can be controlled, but at the moment hardcoding it in two places in the same file seems fine. Merging.

Issue #26, opened 2021-10-08
Add Code Formatter


liam-mulhall:
To ensure consistent code style, we should add a code formatter. I think [Prettier](https://prettier.io/) will suit our needs.

PR #90, opened 2022-01-08
chore: Add Code Formatter

liam-mulhall:
Install and configure the Prettier code formatter. We also install the
eslint-config-prettier package to get ESLint and Prettier to play nicely
together, as recommended by the Prettier docs.

For
https://github.com/numberscope/frontscope/issues/26.
Related to https://github.com/numberscope/frontscope/issues/69.

Prettier is working for me in Visual Studio Code. I also use Neovim and Emacs, so if either @katestange or @gwhitney need help setting up Prettier, I might be able to help. (I've used most of the popular Vim/Neovim plugin managers, but I've only used [use-package](https://github.com/jwiegley/use-package) for Emacs.)

As for actually formatting the files, it can be done in one fell swoop, but I think I'll format each file individually and make sure nothing breaks as they are formatted, however I think this should be done in another PR.

Eventually, I'd like to add complementary ESLint rules and add some Git hooks so that every commit gets formatted and linted.

Regarding numbered items in https://github.com/numberscope/frontscope/issues/69:
| item | what was done in this commit |
|-|-|
| 1. 80-character limit | configured in this commit |
| 2. spaces for indentation | configured in this commit |
| 3. single quotes | configured in this commit |
| 4. trailing comma | configured in this commit |
| 5. no semicolon | configured in this commit |
| 6. conditional style | not addressed in this commit (Prettier doesn't add or enforce curly braces, so I think you can write them however you want) |
| 7. variable naming | not addressed in this commit |
| 8. brace spacing | Prettier's default spacing seems to be similar to what @gwhitney describes in https://github.com/numberscope/frontscope/issues/69 |
| 9. arrow parens | set to "always" (the options are either "always" or "avoid") |
| 10. newline at the end of a file | not addressed in this commit |

### DIFF
--- a/doc/formatting-code.md
+++ b/doc/formatting-code.md
@@ -36,9 +36,20 @@ For example, if you want to run `format-write`, you enter:
 npm run format-write
 ```
 
+Note also that `npm run lint` runs `eslint` in a non-fixing mode, to check if
+there are any violations of non-fixable rules and generate messages about them
+if so.
+
 ## Text Editor / IDE Integration
 
 How exactly you integrate these scripts into your text editor or IDE will depend
 on the text editor or IDE you use. A cursory Google search for "how to run a
-command on save in Vim" yields a few useful-looking results. I would imagine
-there are ways to do this for most text editors and IDE's.
+command on save in Vim" yields a few useful-looking results. There are
+presumably ways to do this for most text editors and IDE's.
+
+### Emacs
+
+Sample configuration code for the (Gnu) Emacs editor which sets up
+`prettier-eslint` to be run whenever a `.ts` or `.vue` file is saved is
+included, with instructions for how to enable that code, in
+`tools/editor/autoformat.el`.
--- a/doc/formatting-code.md
+++ b/doc/formatting-code.md
@@ -24,13 +24,13 @@ capabilities of ESLint.
 * `format-check` runs both `format-check-ts` and `format-check-vue`.
 * `format-write` runs both `format-write-ts` and `format-write-vue`.
 
-To run these scripts at the command line, you simply to enter:
+To run these scripts at the command line, you enter:
 
 ```sh
 npm run <script-name>
 ```
 
-For example, if you want to run `format-write`, you would enter:
+For example, if you want to run `format-write`, you enter:
 
 ```sh
 npm run format-write
--- a/doc/formatting-code.md
+++ b/doc/formatting-code.md
@@ -1,0 +1,55 @@
+# Formatting Code
+
+As of this writing, there are NPM scripts in package.json that automatically
+format your code for you using Prettier and ESLint. These scripts use
+prettier-eslint, which is a package that "formats your code via `prettier`, and
+then passes the result of that to `eslint --fix`." This way we get the benefits
+of Prettier's superior formatting capabilities and the configuration
+capabilities of ESLint.
+
+The following are the NPM scripts:
+
+```json
+    "format-check-ts": "prettier-eslint \"src/**/*.ts\"",
+    "format-write-ts": "prettier-eslint \"src/**/*.ts\" --write",
+    "format-check-vue": "prettier-eslint \"src/**/*.vue\"",
+    "format-write-vue": "prettier-eslint \"src/**/*.vue\" --write",
+    "format-check": "npm run format-check-ts && npm run format-check-vue",
+    "format-write": "npm run format-write-ts && npm run format-write-vue"
+```
+
+* `format-check-ts` runs `prettier-eslint` on all the TypeScript files it can
+  find in the `src` directory. This outputs the changes `prettier-eslint` would
+  make to standard output, i.e. when you enter the command at the command line,
+  it will vomit out its proposed changes.
+* `format-write-ts` runs `prettier-eslint --write` on all the TypeScript files
+  it can find in the `src` directory. This writes `prettier-eslint`'s changes,
+  i.e. files will be automatically formatted.
+* `format-check-vue` runs `prettier-eslint` on all the Vue files it can find in
+  the `src` directory. This outputs the changes `prettier-eslint` would make to
+  standard output, i.e. when you enter the command at the command line, it will
+  vomit out its proposed changes.
+* `format-write-vue` runs `prettier-eslint --write` on all the Vue files it can
+  find in the `src` directory. This writes `prettier-eslint`'s changes, i.e.
+  files will be automatically formatted.
+* `format-check` runs both `format-check-ts` and `format-check-vue`.
+* `format-write` runs both `format-write-ts` and `format-write-vue`.
+
+To run these scripts at the command line, you simply to enter:
+
+```sh
+npm run <script-name>
+```
+
+For example, if you want to run `format-write`, you would enter:
+
+```sh
+npm run format-write
+```
+
+## Text Editor / IDE Integration
+
+How exactly you integrate these scripts into your text editor or IDE will depend
+on the text editor or IDE you use. A cursory Google search for "how to run a
+command on save in Vim" yields a few useful-looking results. I would imagine
+there are ways to do this for most text editors and IDE's.
--- a/doc/formatting-code.md
+++ b/doc/formatting-code.md
@@ -1,6 +1,6 @@
 # Formatting Code
 
-As of this writing, there are NPM scripts in package.json that automatically
+As of this writing, there are NPM scripts in `package.json` that automatically
 format your code for you using Prettier and ESLint. These scripts use
 prettier-eslint, which is a package that "formats your code via `prettier`, and
 then passes the result of that to `eslint --fix`." This way we get the benefits
--- a/doc/formatting-code.md
+++ b/doc/formatting-code.md
@@ -7,17 +7,6 @@ then passes the result of that to `eslint --fix`." This way we get the benefits
 of Prettier's superior formatting capabilities and the configuration
 capabilities of ESLint.
 
-The following are the NPM scripts:
-
-```json
-    "format-check-ts": "prettier-eslint \"src/**/*.ts\"",
-    "format-write-ts": "prettier-eslint \"src/**/*.ts\" --write",
-    "format-check-vue": "prettier-eslint \"src/**/*.vue\"",
-    "format-write-vue": "prettier-eslint \"src/**/*.vue\" --write",
-    "format-check": "npm run format-check-ts && npm run format-check-vue",
-    "format-write": "npm run format-write-ts && npm run format-write-vue"
-```
-
 * `format-check-ts` runs `prettier-eslint` on all the TypeScript files it can
   find in the `src` directory. This outputs the changes `prettier-eslint` would
   make to standard output, i.e. when you enter the command at the command line,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
         'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
         'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
         'operator-linebreak': [2, 'before'],
+        'max-len': ['error', { 'code': 80, 'comments': 80, }]
     },
     overrides: [
         {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,30 +1,31 @@
 module.exports = {
-  root: true,
-  env: {
-    node: true
-  },
-  'extends': [
-    'plugin:vue/essential',
-    'eslint:recommended',
-    '@vue/typescript/recommended',
-    'prettier'
-  ],
-  parserOptions: {
-    ecmaVersion: 2020
-  },
-  rules: {
-    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
-    'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off'
-  },
-  overrides: [
-    {
-      files: [
-        '**/__tests__/*.{j,t}s?(x)',
-        '**/tests/unit/**/*.spec.{j,t}s?(x)'
-      ],
-      env: {
-        jest: true
-      }
-    }
-  ]
+    root: true,
+    env: {
+        node: true,
+    },
+    extends: [
+        'plugin:vue/essential',
+        'eslint:recommended',
+        '@vue/typescript/recommended',
+        'prettier',
+    ],
+    parserOptions: {
+        ecmaVersion: 2020,
+    },
+    rules: {
+        'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+        'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+        'operator-linebreak': [2, 'before'],
+    },
+    overrides: [
+        {
+            files: [
+                '**/__tests__/*.{j,t}s?(x)',
+                '**/tests/unit/**/*.spec.{j,t}s?(x)',
+            ],
+            env: {
+                jest: true,
+            },
+        },
+    ],
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,8 @@ module.exports = {
   'extends': [
     'plugin:vue/essential',
     'eslint:recommended',
-    '@vue/typescript/recommended'
+    '@vue/typescript/recommended',
+    'prettier'
   ],
   parserOptions: {
     ecmaVersion: 2020
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+dist
+node_modules
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,5 +1,5 @@
 {
-  "arrowParens": "always",
+  "arrowParens": "avoid",
   "bracketSameLine": true,
   "bracketSpacing": false,
   "embeddedLanguageFormatting": "auto",
@@ -12,7 +12,7 @@
   "requirePragma": false,
   "semi": false,
   "singleQuote": true,
-  "tabWidth": 2,
+  "tabWidth": 4,
   "trailingComma": "es5",
   "useTabs": false,
   "vueIndentScriptAndStyle": true
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,19 @@
+{
+  "arrowParens": "always",
+  "bracketSameLine": true,
+  "bracketSpacing": false,
+  "embeddedLanguageFormatting": "auto",
+  "htmlWhitespaceSensitivity": "css",
+  "insertPragma": false,
+  "jsxSingleQuote": false,
+  "printWidth": 80,
+  "proseWrap": "preserve",
+  "quoteProps": "as-needed",
+  "requirePragma": false,
+  "semi": false,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "es5",
+  "useTabs": false,
+  "vueIndentScriptAndStyle": true
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,8 @@
         "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-vue": "^7.18.0",
         "prettier": "^2.5.1",
+        "prettier-eslint": "^13.0.0",
+        "prettier-eslint-cli": "^5.0.1",
         "typescript": "~3.8.3",
         "vue-loader": "^15.9.1",
         "vue-template-compiler": "^2.6.11"
@@ -1855,6 +1857,12 @@
       "integrity": "sha512-rr+OQyAjxze7GgWrSaJwydHStIhHq2lvY3BOC2Mj7KnzI7XK0Uw1TOOdI9lDoajEbSWLiYgoo4f1R51erQfhPQ==",
       "dev": true
     },
+    "node_modules/@types/eslint-visitor-keys": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@types/eslint-visitor-keys/-/eslint-visitor-keys-1.0.0.tgz",
+      "integrity": "sha512-OCutwjDZ4aFS6PB1UZ988C4YgwlBHJd6wCeQqaLdmadZ/7e+w79+hbMUFC1QXDNCmdyoRfAFdm0RypzwR+Qpag==",
+      "dev": true
+    },
     "node_modules/@types/events": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/@types/events/-/events-3.0.0.tgz",
@@ -3533,6 +3541,15 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/arrify": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/arrify/-/arrify-2.0.1.tgz",
+      "integrity": "sha512-3duEwti880xqi4eAMN8AyR4a0ByT90zoYdLlevfrvU43vb0YZwZVfxOgxWrLXXXpyugL0hNZc9G6BiB5B3nUug==",
+      "dev": true,
+      "engines": {
+        "node": ">=8"
+      }
+    },
     "node_modules/asn1": {
       "version": "0.2.4",
       "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.4.tgz",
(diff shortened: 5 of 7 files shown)


liam-mulhall:
Had to fix some issues with the commit, so I amended the commit and force pushed it.

gwhitney:
The only setting that I don't think we reached consensus on among the ones you list is (9) arrow parens. To me, `list.map(x => x+1);` reads very clearly, but it's nice to have the parens in a heavyweight thing like
```
doThingWithCallback((submission) => {
  validateSubmission(submission);
  analyzeSubmission(submission);
  implement(submission);
})
```
Is it possible to make prettier just not care about parens in fat-arrow functions and rely on our mutual implementation of a style guide? Or does the group want to reach a consensus on one of prettier's options before merging this PR? If the only options are "always" and "avoid," then I guess I'd vote for "avoid" since in general fewer characters and clutter when that doesn't compromise clarity seems better, and I think cases like the short little anonymous function for a map are much more frequent than big heavy callbacks.

On another matter, however you get there in terms of your process, I strongly suggest a single PR for bringing the codebase into compliance with whatever prettier rules are adopted.

liam-mulhall:
> Is it possible to make prettier just not care about parens in fat-arrow functions and rely on our mutual implementation of a style guide?

I don't think there's an easy way to do it. If you omit the "arrowParens" field from `.prettierrc.json`, it sets "arrowParens" to the default "always" setting. (I just tested this in VS Code.)

Here's my proposed solution to this arrow parens issue:
1. We change "always" to "avoid". Then we'll be able to write `list.map(x => x+1)`.
2. If we have a block of code like (1), we can change it to (2).

```js
// (1)
doThingWithCallback((submission) => {
  validateSubmission(submission);
  analyzeSubmission(submission);
  implement(submission);
})

// (2)
// Prettier ignore
doThingWithCallback((submission) => {
  validateSubmission(submission);
  analyzeSubmission(submission);
  implement(submission);
})
```

From Prettier docs:
> A JavaScript comment of // prettier-ignore will exclude the next node in the abstract syntax tree from formatting.

I think blocks of code like (1) are probably rarer than `list.map(x => x+1)`. Also, if you have multiple parameters in a block of code like (1), you get to keep your parentheses. (Otherwise I think it would be invalid JS/TS.)

@gwhitney, if this proposed solution is acceptable, let me know and I'll commit the necessary changes.

As for formatting the files, what you described is my plan. I meant that I'd format the files individually to make sure nothing breaks and then add the reformats to one big PR. (I wasn't planning on creating a separate PR for each file reformat; that would be a lot of PR's!)

gwhitney:
OK, I can confirm that `npm run test:vue` succeeds on this PR as it currently stands. I ran prettier on SequenceInterface.ts and have a couple of comments:

1) It issued two warnings:
```
[warn] Ignored unknown option { bracketSameLine: true }.
[warn] Ignored unknown option { embeddedLanguageFormatting: "auto" }.
```
It should not produce any warnings as configured, please fix, thanks.

2) It removed all of the semicolons and added a trailing comma to a list. All fine. But it also changed all of the indentation from 4 spaces per level to 2 spaces. This didn't come up in our conversation in #69 (I will add it), but this entire code base has 4 spaces per level at the moment, *and* I find 3 or 4 spaces much easier to read. I realize Prettier has a big attitude about not being configurable, but can the indentation per level be set? If not, I'd vote for finding a code formatter in the same style that can deal with 3 or 4 spaces, to use instead of Prettier. Presuming to mandate 2 spaces per indent level for all code projects seems just a tad too arrogant to me... Anyhow, happy to have everyone's thoughts on this topic.

liam-mulhall:
1. I'll look into those warnings.
2. We can set indent to 3 or 4 spaces. I vote for 4.

gwhitney:
I am cool with `arrowParens: avoid` and then playing it by ear with an ignore comment or whatever if we want to format differently for some particular big case; I agree the little map function case is more common.

I am totally fine with sticking with the current 4-space indent.

Will test more when you've pushed the adjustments based on the conversation so far.

gwhitney:
By the way, if anyone else is wanting to try this out before it goes into place, you can add Liam's repo as a remote (e.g. `git add remote liam https://github.com/liam-mulhall/frontscope.git`), grab his stuff (e.g. `git fetch liam`), check out this submission (`git checkout add-code-formatter`), and then try prettier on a file with `npx prettier <PATH-TO-SOURCE-FILE>` e.g. if you are in the top-level directory `npx prettier src/main.ts`.

I recommend that once this PR is merged, the code base be frozen until Liam submits a PR bringing the code into compliance with the prettier policies -- so please Liam when this is ready for merging, let us know times when it can be merged when you will relatively promptly (within a few days, say) be able to submit the follow-on PR. Once that second PR is in place, we can close #26.

liam-mulhall:
- [x] investigate warnings
- [x] change arrow parens "always" to "avoid" (done in https://github.com/numberscope/frontscope/pull/90/commits/646fb601561bca2c93ebfa45c839461a968a6fbb)
- [x] set indent to 4 spaces (done in https://github.com/numberscope/frontscope/pull/90/commits/646fb601561bca2c93ebfa45c839461a968a6fbb)

Am I missing anything?

@gwhitney, I wasn't able to reproduce the warnings you wrote about in [this comment](https://github.com/numberscope/frontscope/pull/90#issuecomment-1007852363). Here's what I did:

```sh
~/Code/liam/frontscope add-code-formatter* ≡
❯ npx prettier -w src/sequences/SequenceInterface.ts
src/sequences/SequenceInterface.ts 85ms
~/Code/liam/frontscope add-code-formatter* ≡
❯ 
```

As you can see, I'm not getting any warnings. What exact command did you issue to get the warnings?

gwhitney:
Aha, I neglected to run `npm install` after pulling your branch, so I was stuck on an old version of prettier. Once I did, the warnings went away. Continuing testing.

gwhitney:
OK, the 'scope runs fine with the changes as far as I can tell. The indentation is looking good. I am going to poke around more, but so far the only concern I have found is that prettier is breaking long expressions immediately after operators, as in:
```
        if (
            !Number.isInteger(this.gradientLength) ||
            this.gradientLength <= 0
        ) {
```
as opposed to starting the continuation lines with the next operator, as in
```
        if (
            !Number.isInteger(this.gradientLength)
            || this.gradientLength <= 0
        ) {
```
(this example is from prettier run on Kate's VisualizerChaos). Most projects I have worked on have favored the latter format, with the rationale (that I buy into) that it is much easier to see what operations are being applied in the latter format than in the former. And indeed, the standard in mathematical typesetting is to place operations in long formulas that must be broken across lines at the beginning of continuation lines rather than just before a line break. Is this configurable?  I note that if I manually break a long expression just before an operator, prettier reflows the expression so the line break occurs just after the operator. I get the point of consistency, but it feels pretty invasive when it seems like the authors of prettier made a poor choice in a particular aspect. Thoughts? What is everyone's preference on where operators should be relative to linebreaks?
We can't infer a preference from @katestange's code, because she simply never broke a complex expression across lines...

gwhitney:
See for example https://stackoverflow.com/questions/3730002/code-line-wrapping-how-to-handle-long-lines for commentary supporting breaking code lines before operators. (The question specifically concerns Java, but the principles appear to be universal.) Frankly I am surprised that Prettier is breaking after operators...

gwhitney:
Hmm, eslint has a rule for linebreaks with respect to operators (see https://eslint.org/docs/2.0.0/rules/operator-linebreak), which can be configured either way. Surprisingly to me, the default for eslint is to break after operators, even though most all discussions I can find of this on the web, including style guides for a number of major software companies, recommend breaking before operators. Anyhow, this gives me some small hope that maybe if we configure eslint to want the line breaks to be before operators, then prettier will do it that way? Do prettier and eslint interact in that kind of way at all? If not, it seems as though eslint itself can be used as the code formatter, looking at its homepage... not to try to second-guess, but given that we are already using eslint, why not just use its code formatter rather than adding a tool?  Thanks for working through these questions, Liam -- it's very much appreciated!

gwhitney:
One more comment and then I'll await all your thoughts/responses: Here's an article (https://www.benmvp.com/blog/prettier-eslint/) that recommends configuring eslint to run prettier as a rule in lieu of running prettier directly. Apparently it can do this so that then when you use eslint to format your code, it will use prettier to do the formatting for all of the prettier violations. Maybe if we do it this way we can then in addition configure eslint to want the line breaks before operators (which it can already by itself enforce in its built-in code formatter) and get the best of both worlds. (But I am slightly unclear on what will happen in eslint if prettier-as-a-rule contradicts one of the other eslint rules, in terms of which will take precedence and how code will end up formatted when you ask eslint to format it.)

OK, so now I will await responses, but if any of you would like me to explore the prettier-via-eslint option I am happy to. Just let me know.

gwhitney:
@liam-mulhall when you get back to looking at this further, please also rebase on main (even though there are no conflicts, it's just nicest to be working on top of current main when possible).

liam-mulhall:
Thanks for all of the feedback, @gwhitney.

Here are the options as I understand them:

1. "Original Plan" (use Prettier with the recommended library that makes it play nicely with ESLint and then configure ESLint separately)
2. "No Prettier" (just use ESLint for everything)
3. "Prettier-via-ESLint" (as described in [this comment](https://github.com/numberscope/frontscope/pull/90#issuecomment-1007897075))

I surmise from your comments that you'd prefer 3, but I wonder if that's just because you think I want to use Prettier. If I told you that I would have no issue with abandoning Prettier in favor of just using ESLint, would you prefer 2?

I'd prefer 1 or 2 rather than the more off-the-beaten-path 3 for the sake of ease of maintenance. As I said, I have no issue with just using ESLint and abandoning the notion of using Prettier.

gwhitney:
My main objection to (1) is that as far as I can tell as of right now, it will force all of our expressions that need to be broken across lines to have the line breaks after operators, which seems undesirable (and against most recommendations I can find), with no way to fix until such unknown date as prettier v3 is released.

So I would prefer either (2) or (3), and I am fine with either. Option (2) is clearly simpler and a smaller change from where we are, given that we are already using eslint, just not using it as a formatter. The primary advantage of (3) over (2) is that eslint by itself has no provision to reflow lines to the specified limit for you; we simply have to enable the mandatory line-length rule and then violations will fail in eslint but it's up to the coder to manually fix them. By involving prettier, you gain the ability for lines automatically to be broken to the specification. (I think there are a few other violations that eslint will not automatically reformat to fix, but line-length is the big one.)  That does sound convenient, so I am willing to help deal with the extra complexity of (3) if others on the team would prefer that.

So those are my thoughts/comments, and Liam if you are so inclined to just choose either (2) or (3) at your preference, I am happy to support either option.

gwhitney:
And just as a pointer, if I am reading the docs correctly, to pursue option (2) we would need to set up editors to do something like call "eslint --fix" on save.

liam-mulhall:
I've added the package recommended in https://www.benmvp.com/blog/prettier-eslint/ to make it possible to run Prettier as an ESLint rule in https://github.com/numberscope/frontscope/pull/90/commits/61d9984b1b0b24afd56ef1e56e719c29ac5f4db0.

liam-mulhall:
I've reached a point of confusion. In [this comment](https://github.com/numberscope/frontscope/pull/90#issuecomment-1007897075), you (@gwhitney) recommend a blog post which recommends a package (eslint-plugin-prettier) to run Prettier as an ESLint rule. So far so good. But then later in the same comment (probably in an edit), you point to a thread that recommends another package, prettier-eslint.

* eslint-plugin-prettier runs Prettier through ESLint.
* prettier-eslint formats code using Prettier and then passes the result of that to `eslint --fix`.

Do we want to add both of these packages? I'm very unclear on how prettier-eslint actually works in practice. It seems like prettier-eslint operates on strings. There's another package that provides a CLI for prettier-eslint: https://github.com/prettier/prettier-eslint-cli. Do we want to add that package as well?

gwhitney:
Right. I agree my comment ended up being somewhat confusing due to multiple edits. As far as I can tell, eslint-prettier-plugin can't allow your eslint configuration to override non-configurable prettier conventions, whereas prettier-eslint can, as it runs prettier on its own first, not as an eslint rule, and then runs eslint after that, so eslint "wins".  So if you are OK with indulging my strong feeling that binary operators "belong" at the beginning of continuation lines (rather than just before linebreaks), the only option seems prettier-eslint (and thus _not_ eslint-prettier-plugin).  I can't think of any circumstance in which one would use both.

Of course, I will accept the group consensus on these issues.

liam-mulhall:
Thanks for the clarification! I'll try to get prettier-eslint working.

liam-mulhall:
In https://github.com/numberscope/frontscope/pull/90/commits/ae439ee1962aabddd7c5cec7d2c3524e7bc417b8, I overwrote https://github.com/numberscope/frontscope/commit/61d9984b1b0b24afd56ef1e56e719c29ac5f4db0 and installed prettier-eslint and prettier-eslint-cli. I also added @gwhitney's desired operator-linebreak rule. I tested these packages out by formatting `.eslintrc.js` and they seemed to work well. I added some dummy code to the end of `.eslintrc.js` to test whether they would play nicely with the new operator-linebreak rule. (I removed the dummy code before committing.)

This

```js
if (somethingLong1111111111111111111111111111111111111111111111111111111111111 || somethingLong2222222222222222222222222222222222222222222222222222222222222) {
    console.log('hi')
}
```

got turned into

```js
if (
    somethingLong1111111111111111111111111111111111111111111111111111111111111
    || somethingLong2222222222222222222222222222222222222222222222222222222222222
) {
    console.log('hi')
}
```

when I ran `prettier-eslint .eslintrc.js --write`. (However, there was a warning `jsxBracketSameLine is deprecated.`, but this isn't set in `.prettierrc.json`, so I don't know why I'm seeing this warning.)

In summary, the packages and the operator-linebreak rule seem to be working well in tandem.

liam-mulhall:
I think the next steps are:

- [ ] investigate `jsxBracketSameLine is deprecated.` warning
- [ ] solicit comments/questions from others
- [ ] add a doc about formatting and linting code

@gwhitney, do you have any feedback before I proceed with next steps?

gwhitney:
As soon as I am able (not sure exactly when because I have upcoming deadlines in my role as editor of two different problem columns) I will review the functionality of the latest commit (and of course anyone else who wants to take up review/merging of this PR in the meantime is welcome to). But I don't see any reason that you shouldn't proceed with the action items you list immediately.

My other feedback about this PR is that generally PRs that introduce new code/packages/software (as opposed to documentation-only PRs or refactoring PRs) should have some functionality outcome. What's that outcome for this PR? I'd suggest it should be an "npm" directive (like part of `npm test`, say) that verifies each file matches what would be produced by prettier-eslint (and that there are no errors or warnings in the process). Then, once this PR is merged, it can become policy that this directive succeeds. And either in this PR or a subsequent one we can document how contributors can set up their environments so that prettier-eslint is run automatically to ensure their code is ready to succeed on that test.

If you think there already is a functional outcome as this PR stands or if there's something else that should be the outcome, I am totally open to hearing about that.

gwhitney:
Excellent, those scripts seem like they are what we need. I'll get back to reviewing this PR as soon as I can. Of course, anyone else, feel free to jump in here. What's the status on the warning? As far as the documentation goes, that could be in this PR, or a subsequent one, at your discretion, Liam.

gwhitney:
The 'jsxBracketSameLine is deprecated warning is https://github.com/prettier/prettier-eslint-cli/issues/428

gwhitney:
OK, I have just filed a PR (https://github.com/prettier/prettier-eslint-cli/pull/430) which would address the problem at the prettier-eslint level. Unclear how active prettier-eslint-cli is (the last release was a year ago), so not sure when/if that will be integrated. It's a one-line fix, so I am open to suggestions on how to proceed in our project in the potentially long meantime. Brainstorming:
1) Just ignore the warning (and document that we are doing so)
2) write our own little wrapper script that filters out the warning and use that instead
3) temporarily check into our repository a file with the one-line fix and a slight variant of prettier-eslint that uses the fixed version, and use those
4) temporarily check into our repository a script that patches prettier-eslint (so we don't have to duplicate a bunch of code in our repository) and then document that new installations have to run that script and use the patched version it creates

Definitely open to other options if anyone can think of any.
1)

gwhitney:
Perhaps better than any of the above would be either of the 2nd or 3rd options in https://www.mariokandut.com/how-to-patch-an-npm-dependency/

Opinions on how we should go here are welcome.

liam-mulhall:
I think the second option in https://www.mariokandut.com/how-to-patch-an-npm-dependency/ where we use the patch-package package is preferable to maintaining a fork of prettier-eslint-cli. However, I wonder if that's overkill for a warning like this. I'm not a big fan of seeing weird warnings like that. I'm also not a big fan of adding lots of packages that we have to know about and maintain. In this case, we'd be adding a workaround (patch-package) for a workaround (prettier-eslint-cli) for a workaround (prettier-eslint). My vote is for "Just ignore the warning (and document that we are doing so)".

gwhitney:
Yes I can hold my nose on that warning too, so long as the return code from `prettier-eslint` is OK (so that automated tools once we have them in place won't think the formatting has failed) and as long as the extra output doesn't disrupt the use of `prettier-eslint` for automated formatting in our respective editors.

doc/formatting-code.md
```diff
@@ -0,0 +1,55 @@
...
+
+```json
+    "format-check-ts": "prettier-eslint \"src/**/*.ts\"",
+    "format-write-ts": "prettier-eslint \"src/**/*.ts\" --write",
+    "format-check-vue": "prettier-eslint \"src/**/*.vue\"",
+    "format-write-vue": "prettier-eslint \"src/**/*.vue\" --write",
+    "format-check": "npm run format-check-ts && npm run format-check-vue",
+    "format-write": "npm run format-write-ts && npm run format-write-vue"
+```
+
```
gwhitney, line 20: I would mildly recommend quoting all of this code in the docs here because (a) you've described the behavior of each script in detail below, and (b) it just becomes another thing to sync manually when the relevant code changes. If we adopt a doc framework (see #71) that allows automated code snippeting (that's automatically kept up to date), I wouldn't object to such code quotes in the docs when useful. If you feel strongly that the docs should have the code here because it will be helpful to people trying to use the tools, then please in the relevant code section (in package.json, I guess) comment that any changes there need to be propagated to this doc. Otherwise they will absolutely fall out of sync eventually. Thanks.

gwhitney:
Other than my one comment, all is looking good to me, but I or someone still needs to review the functionality of this version of the formatting before merging. If nobody else gets to it, I can mid-week or so, once _The Prisoner's Dilemma_ is safely submitted to my editors.

doc/formatting-code.md
liam-mulhall: The snippet was removed in https://github.com/numberscope/frontscope/pull/90/commits/45a1d13411b7e0681d276678db36656bd017c48f.

gwhitney:
OK, I ran `npx prettier-eslint src/visualizers/VisualizerChaos.ts` and the output looked very nice. Thanks for all of the effort you put into setting this up! The only slightly odd thing I noticed is that in-line comments `// here is some stuff I gotta tell you` in several places ran over our 80-character limit. I would have expected them to be broken into blocks of `//`-comments, keeping each individual line to the limit. Is that just how prettier does it, or is it configurable? If it's configurable, how would we want it? I know this is up to prettier, since I know eslint does not have the ability to fix linebreaks for line length, unlike prettier, so we just have to go with how prettier does things on linebreaking. I am definitely not going to make any fuss about this formatting; I just thought that if it is configurable, we should decide what option we want; and personally in any case I'd suggest we put in the project conventions that adhering to the line-length limit is recommended even for comments.

gwhitney:
OK, I have reviewed prettier-eslint's output on a variety of files and am OK with everything I saw, the only questionable thing being the lengths of comment lines. Thoughts?

gwhitney:
And ok, I now have this running automatically on save in Emacs. To arrange that, I had to use some custom code that can be put in Emacs' configuration file (`.emacs`, usually). It's roughly 60 lines of code. Do you think I should (a) check that Emacs code into our repository (say in `tools/autoformat.el`) and document how to use it in `docs/formatting-code.el`), (b) quote the whole 60 lines in `docs/formatting-code.el` since after all it's not actually _used_ by this package, or (c) something else? And in any of these cases, should I make the change as a further commit folded into this PR, or file a separate PR when I merge this?

Anyhow, given all this, as soon as I hear back about the comment line lengths and we reach consensus on that and we decide where/how/whether to give the details on running this from Emacs, I am ready to merge this (finally, sorry it was circuitous).

liam-mulhall:
### 80-Character Line Limit for Comments

We can add an ESLint rule to enforce this for comments! See https://eslint.org/docs/rules/max-len. I'll add this rule in another commit. We'll have to disable the rule _if_ we need a comment longer than 80 characters, e.g. a long URL.

### Config Files for Text Editors

@gwhitney, please add your Emacs config to `/tools/editor/frontscope-config.el` as a commit in this PR. Instructions for other editors should be placed in the same directory as either config files or markdown files with instructions.

### Unrelated: Emacs Config Location

> To arrange that, I had to use some custom code that can be put in Emacs' configuration file (.emacs, usually).

If you want, you can put your Emacs config in `~/.config/emacs`. See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html. Just FYI. It helps keep your home directory nice and tidy.

liam-mulhall:
80-character ESLint rule added in https://github.com/numberscope/frontscope/pull/90/commits/826078c5b9ab2836a17cffb4576359a3e1e5955e. Testable by running `npm run lint`.

liam-mulhall:
Unfortunately, I don't think ESLint will auto fix long lines. See https://github.com/eslint/eslint/issues/11325.

gwhitney:
Right, I agree. That's one of the main reasons we're using prettier, since it can re-flow the code to 80 characters; it just doesn't touch the comments, and eslint doesn't change them either, which is why enforcement is mandatory.

But I am confused; when I run prettier-eslint (on say VisualizerChaos.ts) it does wrap the code, and it then uses eslint --fix to get the binary operators at the beginning of continuation lines, but it doesn't mention anything about there still being lines >80 characters long (because of comments).  Whereas if I save the result of prettier-eslint and do `npm run lint`, then eslint does complain about long lines. Why the difference? Are there two different configuration files for eslint, one used for prettier-eslint and one used when eslint is run stand-alone? That seems non-ideal; it seems like there should just be one universal setting for eslint for this project. Or, if there is only one configuration, are the errors being suppressed somehow when run through prettier-eslint? If so, how/why? It would seem to me that when I apply prettier-eslint, it would be good to see the problems right then, rather than waiting for another step...

I will commit the emacs configuration and some comments on it in the doc file you added, and hopefully by then we can get clear on the different behaviors of eslint and whether they should be made uniform, and then we can get this merged.

liam-mulhall:
> Why the difference?

See [this comment](https://github.com/numberscope/frontscope/pull/90#issuecomment-1032019567). ESLint can't auto fix long lines. prettier-eslint uses Prettier to format, then passes the result of that to `eslint --fix`. I'm pretty sure `eslint --fix` doesn't produce errors, it just fixes files. (I can verify this later if you're curious.)

> Are there two different configuration files for eslint, one used for prettier-eslint and one used when eslint is run stand-alone?

No. There's a Prettier config file and an ESLint config file. prettier-eslint doesn't have a config file.

gwhitney:
> I'm pretty sure eslint --fix doesn't produce errors, it just fixes files.
That's not it, see the below transcript of an eslint --fix run that moved an operator to the beginning of the following line, while
simultaneously producing errors:
```
frontscope> npx eslint --fix src/visualizers/VisualizerChaos.ts

/home/glen/code/frontscope/src/visualizers/VisualizerChaos.ts
    3:1  error  This line has a length of 87. Maximum allowed is 80           max-len
   13:1  error  This line has a length of 97. Maximum allowed is 80           max-len
   30:1  error  This line has a length of 103. Maximum allowed is 80          max-len
[remainder of messages cut]
✖ 42 problems (42 errors, 0 warnings)

frontscope>
```
(Note that without the --fix option, it reports 43 errors, one potentially fixable.)

So I guess prettier-eslint must be turning off the error reporting. I'll take a look and see if there's a switch or config to turn it back on (presuming that's OK with you if there is?)  If not, then we'll just have to rely on `npm run lint` after the fact as a separate step -- we should probably start building a code review checklist, to become the basis for automated PR processing when we get to that.

gwhitney:
Ah yes, the `prettier-eslint` module (that `prettier-eslint-cli` uses to provide the command we are running) notes in comments in its main source file (https://github.com/prettier/prettier-eslint/blob/master/src/index.js) that it is stripping all errors and warnings. That's a pity; the emacs code I am posting would pop up anything written to standard error during the fixing process. I was looking forward to seeing any lint errors pop up when I save the file. I will file an issue on the github projects (but I don't think they are being actively maintained). And in the meantime we shall move forward, we have spent much time on this already. I just need to check in the emacs stuff and I will merge.

gwhitney:
OK, I will merge later this evening (hate to do a bunch of stuff all together, seems like that more often leads to mistakes).

PR #101, opened 2022-02-09
chore: Comply with eslint and prettier-eslint

### LINKED ISSUE #100: Convert code to be eslint-compliant
Right now the code base generates many eslint warnings and/or errors. We should perform a one-time checkin of (nearly every) file so that all code is compliant with the project's standards (see #69). Thereafter, we should enforce "lint compliance" on all merges into main thereafter.

gwhitney:
Note that the two are not quite the same. The line length in the .prettierrc
  is not an absolute ceiling, so this commit reduces it to 78 so that the output
  will all comply with the eslint limit of 80.

  Resolves #100.

  Note that I had to manually move a number of style parameters from the template section to the style section in a few .vue files. For reasons I do not entirely understand, these parameters were no longer being picked up in the browser after the prettier-eslint transformation, but they work OK in the style section, which is probably a better place for them anyway.


gwhitney:
Unfortunately this commit will pretty much destroy the the meaningfulness of the "blame" view on Github, but better to do that now rather than wait even longer. Now that formatting will be consistent, there should be very few changes caused by reformatting.

gwhitney:
Please test as thoroughly as you have the stomach for, considering the magnitude of the changes (even though they are almost all whitespace or punctuation changes). Thanks!

liam-mulhall:
This passes our minimal test suite. When I tested manually, I noticed that our footer is gone.

### In bfd2e2da8b89e7d0ff6956d1290fa3bdb10d493e:

Screenshot from 2022-02-09 10-52-32

### In this PR:

Screenshot from 2022-02-09 10-55-20

I will add a commit that fixes this momentarily.

liam-mulhall:
@gwhitney, take a look at https://github.com/numberscope/frontscope/commit/fa713301a2d464aed28f19bb6eaf394781b9931f. If all looks good to you, I think we should merge.

gwhitney:
Good catch, Liam! Sorry I missed that. I should have done before-and-after screenshots like you did, once I noticed something was wrong. Anyhow, the additional commit looks good to me, and the extra functionality in the footer and the test for it are a +1. So as far as I am concerned, you can merge when you are comfortable.

liam-mulhall:
Hmm... I think my fix didn't get merged. I will open a PR with the fix in just a second.

Issue #58, opened 2021-12-05
Validation errors not cleared upon "Close"


katestange:
Load the scope.  Click "Mod Fill".  Click "Save changes".  A validation error will be shown, "Mod dimension must be positive."  Click "Close".  Click "Differences."  The same validation error is still displayed in the Differences pop-up, even though it is inappropriate here.

gwhitney:
This is resolved in ui2. Closing.

Issue #68, opened 2021-12-17
"hot-modifying" the parameters of a visualizer


katestange:
Have the parameters of the visualizer accessible (side panel?) during the main "draw" view.  If parameters are changed, immediately interrupt and redraw (no submit button).

khaledallen:
[Also an item on the todo.md]

katestange:
I'm adding the "important" tag to this because I think this is the main obstruction to "having fun" with numberscope at the moment (besides adding more visualizers).

gwhitney:
In light of #108, there was more discussion of this issue today. In particular, the question came up of whether a visualizer in progress should "just keep drawing" with the new parameters or clear itself and start again from the beginning with the new parameters. The consensus was that ideally it should be up to the visualizer -- when it is "notified" of a change, it should be able to analyze what's changed and decide which action to take. If it keeps on going but the user wanted to restart, there should also be a way to do that. It is not currently clear what the protocol should be for the UI notifying the visualizer of a change, though. So that piece of this issue could use a solid proposal.

gwhitney:
Also in the discussion today, @liam-mulhall mentioned that there were Vue libraries for state management that might be relevant to this issue. (Any links you might have, Liam, would be welcome.)

liam-mulhall:
Pinia for state management: https://pinia.vuejs.org/.

liam-mulhall:
For discussion at weekly meeting: Do we want this before we run Numberscope in Fall 2022?

liam-mulhall:
It would be nice to have this for the Fall 2022 undergrads. KS: "Ideally, but not necessarily."

liammulh:
We discussed this in today's (10/31) dev meeting.

- KS: We need some way of re-running `setup()`.
- GW: We can't tell p5 to call `setup()` again?

- KS & GW: We need to distinguish between restart and hot modify.
    - restart is new instance, i.e. new `setup()`
    - hot modify is modify `draw()`
- KS & GW: Programmer should know there are two ways to change, they need to be careful about this.

- GW: What if sequence changes?
  - Do we completely re-draw?
  - Or do we continue with new sequence?
- KS: Why not both functionalities? (Allow user to decide.)
    - If you want it to hot-modify, then do that.
    - If you want it to re-setup, then do that.

liammulh:
We also talked about the UI for params today (10/31).

- GW: The first step would be to use a side panel (or something similar) instead of a modal.
[...]

Issue #45, opened 2021-11-19
Don't `console.log` in Production


liammulh:
It's generally considered bad practice to use `console.log` in production. When we deploy, we need to make sure we don't have any `console.log`'s.

It might be that `npm run build` strips out `console.log`'s. We can address this when we're close to deploy.

gwhitney:
I recommend this issue remain until, and be closed when, there are no console.log commands anywhere in the code base that could be hit in production operation. (I am not actually clear on how far or near we are to that state.)

PR #108, opened 2022-02-15
refactor: Overhaul param system

### LINKED ISSUE #55: Sequence.isValid may be out of whack with return from validate()
The difficulty is a design flaw in SequenceClassDefault: both the return value from validate() and the Sequence object have an isValid flag, but there is nothing keeping them consistent with each other. So they could contradict each other, leading to unpredictable behavior. The code should be modified so that the return value from validate() is the sole authoritative source for whether a Sequence is valid. And in particular as part of this, the code in SequenceClassDefault.ts at the end of the validate() implementation there that ends up with isValid false in the object but true in the return value of validate() needs to be fixed.

### LINKED ISSUE #56: SequenceParam 'required' flag ends up redundant with much code in validate() method.
In many Sequence implementations, there are params that have the 'required' flag set to true in their SequenceParam structure, but then the implementation checks to make sure the param was supplied in the validate() method. While one might appreciate the abundance of caution, this is actually a case of the same information being specified twice. The code should be enhanced to provide a guarantee that if the param is marked as required in the SequenceParam structure, then the sequence will not be constructed without that param; then individual implementations can stop checking that required parameters were in fact supplied (at least as long as they call their super.validate() method, at least, say).

### LINKED ISSUE #60: OEIS sequence generator fails silently on malformed OEIS IDs
If you try malformed "000100" or "A00100" instead of the properly formed "A000100" when generating a new OEIS sequence, it claims to have generated it but then the numbers are apparently "NaN"s (judging by what happens if you use the generated sequence with the Differences visualizer).  Instead of failing silently like this, it ought to complain to the user.

### LINKED ISSUE #62: new param types for sequences and visualizers
Right now, `src/shared/ParamType.ts` lists only three types of possible parameters for visualizers and sequences: number, text and boolean.  In #59, I find that I would like to have a "multiple choice" type that will allow me to choose from 4 different colouring schemes.  As a workaround, I used an integer that I interpret as such a choice.  It would be nice to have a general "multiple choice" functionality.  (And then use it in the Chaos Visualizer #59.)

### LINKED ISSUE #82: Parameters should have expandable subcategories
It would be nice to have the ability to group visualizer/sequence parameters, and make groups collapsible/expandable.  For example, in the chaos visualizer, you could have a bunch of parameters for handling the case of multiple walkers.  Users who only want one walker would not need to bother with these.  They could be grouped, and collapsed by default, but expand on a click.  This would make the parameters panel more inviting and easier to navigate.  Groups should have a title.

gwhitney:
To allow params to be of any type (rather than just `string | number | boolean`
  and to allow params to correspond to correctly-typed top-level properties,
  this commit creates a common base class Paramable for Sequences and
  Visualizers that handles communication of params with the UI.

  Basically, a class just derives from Paramable, defines a `params` property
  with objects describing the top-level properties that should be params,
  and defines a checkParameters() function that determines if a collection
  of candidate values (as `value` properties in the `params` object) are
  acceptable. If so, they are copied automatically into the implementing
  object's top-level properties. If the implementation changes any of these
  properties and wants the change to be reflected in the UI, it must
  call `this.refreshProperties()` manually.

  Also, extends SeqVizParamModal.vue to present reasonable UI interfaces
  for a much larger variety of param types, as well as adds features such as
  hidable params, indentation, and so on.

  Resolves #55.
  Resolves #56.
  Resolves #60.
  Resolves #62.
  Resolves #82.
  (Although the graphic presentation for collapsible subgroups could be
  more polished.)

  Also takes care of #45 in all sequences and visualizers (there are still
  a few gratuitous console.logs in components, though).
  Lays groundwork for #68, in that the channel for communication between
  the UI and the sequence/visualizer is now very clear and as close to
  direct as possible (I don't think we want it any more direct; if it were,
  then in the midst of typing the visualizer could be left in an invalid
  state.)


src/sequences/OEISSequenceTemplate.ts
```diff
@@ -19,37 +15,38 @@ export default class OEISSequenceTemplate extends SequenceCached {
...
 
     async fillCache(): Promise<void> {
-        const resp = await axios.get(
-            `http://${process.env.VUE_APP_API_URL}/api/`
-                + `get_oeis_values/${this.settings.oeisId}/${this.cacheBlock}`
-        )
+        const backendUrl
+            = `http://${process.env.VUE_APP_API_URL}/api/`
+            + `get_oeis_values/${this.oeisId}/${this.cacheBlock}`
+        const resp = await axios.get(backendUrl)
```
liam-mulhall, line 49: I don't care whether we use `resp` or `res` or something else, but it should be consistent across the code base. I think whatever abbreviation we want to use should be documented somehow.

liam-mulhall: Right now, this is the only response variable in frontscope, so there isn't an inconsistency issue.

src/sequences/SequenceFormula.ts
```diff
@@ -18,40 +13,40 @@ import * as math from 'mathjs'
...
+    private evaluator: math.EvalFunction
 
     /**
      *Creates an instance of SequenceFormula
      * @param {*} sequenceID the sequence identifer of the sequence
      */
     constructor(sequenceID: number) {
         super(sequenceID)
-        this.formula = math.compile('n') // tide us over until validate()
+        // tide us over until checkParameters():
```
liam-mulhall, line 33: This comment is vague.

src/shared/Paramable.ts
```diff
@@ -0,0 +1,140 @@
+import {ValidationStatus} from './ValidationStatus'
```
liam-mulhall, line 1: Comments in this file are excellent!

src/shared/Paramable.ts
```diff
@@ -0,0 +1,140 @@
...
+     * refreshParams() should copy the current working values of all of the
+     * params back into the value properties in the params object, in case
+     * they have changed. This way the true current values can be
+     * represented in the UI presenting the params.
+     */
+    refreshParams(): void
+}
+
+/**
+ * @class Paramable
```
liam-mulhall, line 62: I don't understand why this class/file is called "Paramable".

src/sequences/SequenceFormula.ts
gwhitney: I will clarify

gwhitney:
Have verbally responded to all review comments but may or may not be able to supply the commit implementing them for a few days, thanks for understanding.

src/components/SeqVizParamsModal.vue
```diff
@@ -1,118 +1,271 @@
...
+                    retval += ` indent${indent}`
+                }
+                return retval
+            },
+            setBigint(e: Event, p: {value: BigInt}) {
+                const target = e.target as HTMLInputElement
+                try {
+                    p.value = BigInt(target.value)
+                } catch {
+                    // Continue with old value
```
liam-mulhall, line 168: I don't understand this. From what I understand `p.value` is already a `BigInt` and then in a try block you try to use the `BigInt()` constructor to make it a `BigInt`. You then have an empty catch block, so presumably if you can't convert `p.value` to `BigInt`, this method will still work? Why? Shouldn't a `setBigInt` do what it's supposed to do or error out?

src/components/SeqVizParamsModal.vue
```diff
@@ -1,118 +1,271 @@
...
+            groupClass(indent: number): string {
+                let retval = 'form-group'
+                if (indent) {
+                    retval += ` indent${indent}`
+                }
+                return retval
+            },
+            setBigint(e: Event, p: {value: BigInt}) {
+                const target = e.target as HTMLInputElement
+                try {
```
liam-mulhall, line 165: I don't understand this. Why is the catch block empty? This method will work even if `target` can't be converted to `BigInt`? What is `p` in this context? Why is `p.value` a substitute for `target`?

gwhitney: There's nothing to prevent someone from typing, say, letters in a BigInt input area, in which case this handler has to do something -- so what it does is use the old value (what else could it do?) Hope I'm making sense?

katestange: Sorry if I'm just out of the loop, but as far as the user experience, they should get an error if they type garbage, and not be allowed to proceed thinking that was ok.  How will that work if the code just uses the default value?

gwhitney: Sorry if I wasn't clear. This handler responds on (virtually) every keystroke. So if the field is empty, and you type a 1, the parameter becomes 1. Then you type 0, it becomes 10. Then you type a 5, it becomes 105. Then you type an x say. 105x doesn't parse as a bigint, so the last value is used, and the field stays 105, which has the effect of making the x disappear after a beat. So effectively you are prevented from entering anything but a BigInt, which I think is generally considered a more effective UI than accepting ill formed stuff and then telling the user "sorry no good change that". Try it-- I think you will like the "feel"

katestange: Oh that clears it up, thank you!  For bigints, any initial substring of a bigint is again a bigint, so this approach works fine.  Is there any situation where that's a problem?

gwhitney: Well, there are custom handlers for a variety of types because of that sort of thing; this one is just for BigInt. We may need to add more handlers if we want to deal with other types in the future, but I am pretty sure I got all of the current ones.

gwhitney:
OK I think that takes care of everything outstanding

src/components/SeqVizParamsModal.vue
liam-mulhall: > So effectively you are prevented from entering anything but a BigInt, which I think is generally considered a more effective UI than accepting ill formed stuff and then telling the user "sorry no good change that". Try it-- I think you will like the "feel"

This isn't a hill I'm willing to die on, but I think an input should bark at you if you don't supply the correct input. Silently using the old value of 105 when the input is 105x seems very wrong to me. It should be the user's job to provide correct input. It should be the code's job to parse the input, not interpret the user's input. If the input is 105x, the user should see "Can't parse 105x.". That being said, I can't imagine this will be an issue all that often, so having aired my qualms, I'm happy to hold my nose and merge.

liam-mulhall:
@katestange, I'm done with my review. Feel free to merge if/when you are done with yours.

src/components/SeqVizParamsModal.vue
katestange: What about having it show in red when the input is no good, so as soon as you type 105x, it shows it in red?  I haven't tried it yet, so this comment may be inapplicable somehow.  I'd like to pull it and play with it before we merge; I'll try to do that ASAP and maybe comment again here.

gwhitney: > Silently using the old value of 105 when the input is 105x seems very wrong to me. It should be the user's job to provide correct input. It should be the code's job to parse the input, not interpret the user's input.

Again, that's not how it works. You type 105x, you see the 'x' briefly, and then it disappears because it's invalid. What you see is always what you get. There is no silent modification of what you type; there's constant validation so it only lets you type valid input. Have you tried it? What did you think of the user experience? From what I've read, earlier validation and an avoidance of the possibility of incorrect input is generally considered better UX.

gwhitney: OK, Liam looked at this during our meeting and was comfortable with it, so I am re-resolving this conversation, although definitely I recommend @katestange also try it in practice _after_ I take care of the items below and _before_ this is merged.

gwhitney:
In joint discussion/review at the weekly meeting, Liam and I came up with this task list, so I am marking this as WIP not to be merged until the task list is done:

- [ ] Document in Paramable how parameter typing works and what forceType is for and the special values like 'color' that forceType can have.
- [ ] Add a new 'integer' forceType that stores its value in a JavaScript `number` but ensures at the UI level that the value is an integer, and use that type for existing parameters where appropriate.
- [ ] Add more explanation of what the `visibleValue` property of a parameter is and how to use it in comments of Paramable

gwhitney:
Also the following **question** came up in the discussion at the meeting, but Liam and I did not settle on a answer: Should the panel showing parameters with their input boxes and explanations indicate the desired type for the parameter? If so, how and where on the panel should it be indicated? Or is that too much information that will just clutter the panel, since mostly the designations are pretty intuitive (like the "modulus" should be an integer, etc.)? Please post with your thoughts on this, thanks.

katestange:
I've been playing with it and I have a suggestion.  I'm using the Chaos visualizer.  When I click on "Show additional parameters for dots" I get a new section appearing, but it is indented.  When I choose "Index" or "Highlight_one_walker" in "Colour dots by", I also get new options appearing, but the block is not indented.  I think this behaviour should be the same:  both indented blocks, probably.   I would even enjoy a light grey background shading (or similar) to help indicate a chunk of newly appearing parameters.  This would make the user aware what they have to pay attention to.

katestange:
> Also the following **question** came up in the discussion at the meeting, but Liam and I did not settle on a answer: Should the panel showing parameters with their input boxes and explanations indicate the desired type for the parameter? If so, how and where on the panel should it be indicated? Or is that too much information that will just clutter the panel, since mostly the designations are pretty intuitive (like the "modulus" should be an integer, etc.)? Please post with your thoughts on this, thanks.

Interesting question.  Here's a possible solution:  the programmer of the visualizer has the option to specify in the parameter prompt if they feel it is necessary, so we don't need to change anything in the way the box pops up originally.  *But* when something doesn't work, e.g. when the system is deleting your errant "x" in an integer input box, (and only then, to avoid advance cluttering), the type could be indicated by the system in a small font immediately below and flush left with the input field.  This would be a sort of "just in time" approach to the issue.  Is it feasible?

katestange:
Another suggestion I have, but which may not be in the scope of this PR, is that when validation errors occur, it would be nice if the instructions appeared not up at the top, but next to the offending input boxes (before or after).  Then at the top, there could be a generic message to "see errors below" and the errors below could be highlighted with a red background highlight.  Or something along those lines.

gwhitney:
> I think this behaviour should be the same: both indented blocks, probably. I would even enjoy a light grey background shading (or similar) to help indicate a chunk of newly appearing parameters.

These are nice ideas. Right now, that one block that's indented is indented because I set a parameter on the params, not specifically because they are appearing/disappearing. So your suggestion begs the question: Should I add a "background" parameter, and specify "indent" and "background: lightgrey" on all of the parameters you list? Or, on the other hand, should I actually _remove_ the "indent" parameter and instead make it so that the web CSS makes appearing/disappearing parameters always be indented and light grey background?

gwhitney:
> when the system is deleting your errant "x" in an integer input box, (and only then, to avoid advance cluttering), the type could be indicated by the system in a small font immediately below and flush left with the input field. This would be a sort of "just in time" approach to the issue. Is it feasible?

_Something_ along these lines should be doable. I will try to add a commit implementing something like this in the next roughly 1 week.

gwhitney:
> but which may not be in the scope of this PR, is that when validation errors occur, it would be nice if the instructions appeared not up at the top, but next to the offending input boxes (before or after).

That does seem like a good idea, but it would require an extension to the ValidationStatus interface -- basically there would need to be a place to communicate the information of _what_ the offending parameter is. So I think it is actually out of scope for this PR -- it would be great if you can file a separate issue that raises this. Also, I think any solution to this new point will have to resolve #58.

gwhitney:
This: https://github.com/numberscope/frontscope/pull/108#issuecomment-1067466824 is the comment that still needs a bit more feedback, thanks!

katestange:
> This: [#108 (comment)](https://github.com/numberscope/frontscope/pull/108#issuecomment-1067466824) is the comment that still needs a bit more feedback, thanks!

I'm not really sure... maybe Liam has a suggestion on this point?

gwhitney:
OK I will wait to hear the thoughts of @liam-mulhall.

gwhitney:
OK, let me know how that is on the type reminders for integers. Waiting to remove the draft status until we reach consensus on the api for conditionally visible parameters vs. explicit indentation/background properties and I implement it.

katestange:
> > This: [#108 (comment)](https://github.com/numberscope/frontscope/pull/108#issuecomment-1067466824) is the comment that still needs a bit more feedback, thanks!
> 
> I'm not really sure... maybe Liam has a suggestion on this point

I think maybe I just don't know what the ramifications of the two choices are, or pros/cons?  I'm just not enough of a programmer. :)

gwhitney:
> I think maybe I just don't know what the ramifications of the two choices are, or pros/cons? I'm just not enough of a programmer. :)

Well, the "define indent and background and just specify them on whichever parameters you want to get the look you want" gives you more layout possibilities, but it imposes less structure. Right now, the only examples we have where we want a different background and indenting are exactly the parameters that are sometimes visible and sometimes not. If we have a hunch that for the medium-term future those are the only parameters that we're going to want to have those characteristics, and that we will always want sometimes-visible parameters to have those characteristics, then it would be better to drop the properties and just do the styling to make those characteristics a consequence of being hide-able. On the other hand, if having indent and background opens up the imagination of various useful/attractive/interesting ways to lay out parameter panels,then I'd say we should go the first route. I'm not really sure, which is why I was looking for input from others involved. Happy to have your thoughts.

katestange:
About indent and background.  Your description helps a lot, thank you.  I don't feel strongly -- you can do what you think is best, as author here.  I think the philosophy so far has been to have the parameter UI part of frontscope and not under the control of individual visualizers.  And right now, our purpose with it is closely tied to hide-ability.  So until we have some bright ideas about future designs, or visualizers ask for more functionality, I guess I would go with the option that ties it to hide-ability?  I could certainly be swayed by a description of how the other option will help us in future.

gwhitney:
OK, that's enough for me -- I will eliminate the indent parameter and make indentation and grey background an effect of hide-ability. (That's the way I was leaning but didn't want to unduly influence others' views.) @liam-mulhall  if you think the other direction would have been decidedly better I will also be happy to revert to that, so definitely still feel free to chime in.

gwhitney:
OK, see how you like this. :)

katestange:
I've just been playing with it and I am very happy with it. :)

liam-mulhall:
Okay, I also played/tested and I think it can be merged. Thanks, @gwhitney!

PR #132, opened 2022-07-03
Reinstall prettier-eslint to avoid "prettier ignore" directives.

### LINKED ISSUE #124: Make Prettier Comply With ESLint
In https://github.com/numberscope/frontscope/pull/123, I haven't added the package that makes it so that Prettier doesn't clash with ESLint when you have a long line and try to put an operator on the next line. This work was done in https://github.com/numberscope/frontscope/pull/90, but needs to be done again so that we don't have to decorate too much of the code with Prettier directives.

gwhitney:
Switching to prettier-eslint for formatting and linting allows the use of
  prettier for flowing code (and text, for example in Markdown files) to our
  desired line length, while still adhering to desired eslint rules that
  conflict with prettier, such as operators at the beginning of lines.
  (Note that I still had to allow linebreak after '=' to get reasonable
  results with prettier, but the code seems to run fine that way.)

  This commit also takes advantage of the change to remove all prettier
  directives from comments throughout the source.

  Resolves #124


gwhitney:
I am confused; how did your first commit with the long README line occur, @liam-mulhall ? I tried to set things up so that prettier ran on all files (so it should have flowed the README) and so that lint ran before a commit. Is this not working? If so, I should fix before we merge this PR.

liam-mulhall (commented):
@gwhitney, I reviewed the changed files. If (A) is the correct way to format the code snippet, then the changes look good to me.

```js
// (A)
// The equal sign is on the same line as the long variable.
const someReallyVerboseVariableThatMightBeTooLong =
    'a bit of info  '
    + 'another bit of info '
    + 'last bit of info'
```

```js
// (B)
// The equal sign goes on the next line.
const someReallyVerboseVariableThatMightBeTooLong
    = 'a bit of info '
    + 'another bit of info '
    + 'last bit of info'
```

I did some local testing of `lint` on `.ts`, `.vue`, and `.md` files. The script seems to be working. There's one caveat to working with `.md` files that I've discovered: The `lint` script can make your markdown lines longer than the character width limit if you have something like `code text` or a URL.

Please let me know if (A) is the correct way to format the above code snippet. If so, I see no issue with merging.

liam-mulhall:
> I am confused; how did your first commit with the long README line occur, @liam-mulhall ? I tried to set things up so that prettier ran on all files (so it should have flowed the README) and so that lint ran before a commit. Is this not working? If so, I should fix before we merge this PR.

@gwhitney, I was confused too. I had a line that was something like:

```
The following line is 70 (1234567890) + 6 (spaces) + 14 (`npm run lint`) = 90 characters wide.
1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 `npm run lint`
```

And I guess `prettier-eslint` tried to preserve my `npm run lint` code text because I think `code text` only displays correctly if it's on the same line.

liam-mulhall:
Hmm...

`here's some code text
not on the same line`

Actually, I'm not sure why this happened. I'll investigate a bit more.

liam-mulhall:
Okay, I investigated a bit more and https://github.com/numberscope/frontscope/pull/132#issuecomment-1175368468 is accurate.

gwhitney:
OK, this behavior of prettier-eslint (and I think just prettier, in fact) is clearly unacceptable. I will investigate and resolve; marking the PR as draft until that's done.

gwhitney:
Yes, prettier run by itself reflows the README.md line to be much longer than 80 characters.

gwhitney:
Aha, I have discovered that to effectively wrap Markdown files, prettier must be invoked with `--prose-wrap always`; I will figure out how to do this in the context of prettier-eslint and "undraft" this PR when I have updated it to do so.

gwhitney:
OK, prettier seems to be flowing the markdown fine now. @liam-mulhall whenever it's convenient for you to continue your review of this PR, I'd be happy for you to do so.

As for your question about the equals sign on an assignment with a long variable name or a multi-line expression on the right-hand side of the equals, Prettier and eslint still seemed to be fighting about this (specifically the indentation) even when using prettier-eslint and they were getting along on other operator linebreaks. And frankly, I am fine with the = being at the end of a line before a linebreak, because there's not much semantic information there, unlike with actual computational operators where it's convenient to see what operation is being called very clearly and quickly. So I found that Eslint can be configured to leave = at the end of the line while putting all other operators at the beginnings of lines, and then all seemed well with Prettier and Eslint not fighting with each other, and the results read fine to my eyes. If you are OK with the current output of prettier-eslint as well, then I think these settings are good. If it's important to you that = and all other operators are treated strictly the same (so that the = of an assignment will also be formatted to the beginning of a line) let me know and I will find a way to make that work. Thanks!

liam-mulhall (approved):
In https://github.com/numberscope/frontscope/pull/132#issuecomment-1175583297, @gwhitney asked:

> If you are OK with the current output of prettier-eslint as well, then I think these settings are good.

I'm happy with the current output! I just wanted to make sure it was acceptable to you and not an oversight.

The changes in your latest commits look good. The new `"proseWrap": "always"` rule is working when I test locally.

liam-mulhall:
@gwhitney, suppose you have some poorly formatted changes, you `git add` these changes, and then you try to `git commit`. Right now, I believe the pre-commit hooks are `npm run lint` and `npm run test:unit`. I don't think `npm run lint` can easily fail (maybe it never fails?), so it seems like `npm run lint` runs, succeeds, and then your poorly formatted changes are in version control. (I don't think the re-formatted files are staged prior to the commit.)

I'll do a bit more investigation tomorrow morning. If what I wrote above is true, then we might want to revert this PR and make some fixes.

liam-mulhall:
Okay, I did some testing and I'm convinced that what I wrote above is true. We need to make it so that the commit fails if the files aren't formatted correctly. Maybe we could add a pre-push commit?

PR #133, opened 2022-07-06
Revert "Reinstall prettier-eslint to avoid "prettier ignore" directives."

liam-mulhall:
Reverts numberscope/frontscope#132.

In https://github.com/numberscope/frontscope/pull/132, I wrote:

> @gwhitney, suppose you have some poorly formatted changes, you git add these changes, and then you try to git commit. Right now, I believe the pre-commit hooks are npm run lint and npm run test:unit. I don't think npm run lint can easily fail, so it seems like npm run lint runs, succeeds, and then your poorly formatted changes are in version control assuming the tests also succeed. (I don't think the re-formatted files are staged prior to the commit.)

> I'll do a bit more investigation tomorrow morning. If what I wrote above is true, then we might want to revert this PR and make some fixes.

> Okay, I did some testing and I'm convinced that what I wrote above is true. We need to make it so that the commit fails if the files aren't formatted correctly. Maybe we could add a pre-push commit?

<hr/>

TL;DR: Our pre-commit hook isn't failing when we try to commit poorly formatted files. We need to make it fail so that you can't commit poorly formatted files.


liam-mulhall:
I think it would be easier just to open a separate PR.

gwhitney:
Right, I agree, thanks. I will provide one right away.

PR #135, opened 2022-07-06
style: Don't allow files with prettier/eslint errors to be commited

### LINKED ISSUE #134: git hooks allow "non-prettier" code to be committed
@gwhitney, suppose you have some poorly formatted changes, you `git add` these changes, and then you try to `git commit`. Right now, I believe the pre-commit hooks are `npm run lint` and `npm run test:unit`. I don't think `npm run lint` can easily fail, so it seems like `npm run lint` runs, succeeds, and then your poorly formatted changes are in version control assuming the tests also succeed. (I don't think the re-formatted files are staged prior to the commit.)

I'll do a bit more investigation tomorrow morning. If what I wrote above is true, then we might want to revert this PR and make some fixes.

_Originally posted by @liam-mulhall in https://github.com/numberscope/frontscope/issues/132#issuecomment-1175784445_

gwhitney:
It turns out that prettier-eslint suppresses eslint errors that cannot
  be fixed (as it is focused just on fixing the things it can), and that
  files changed in pre-commit hooks are not automatically staged.
  Together, these issues were allowing files with prettier discrepancies
  and/or eslint errors to be committed.
  This PR fixes the situation by (a) installing and using `lint-staged`
  to overcome the second difficulty, and (b) explicitly running a new
  `lint:check` script after the prettier-eslint step to prevent eslint errors
  from getting through.
  For better infomation to the committer, it also adds a `lint:list` script
  that just lists files that are out of prettier-eslint compliance, and
  documents all of the new scripts in the README.
  Finally, it promotes unused variables to an error so that they will be
  disallowed in the repository (formerly they were warnings, which do not
  affect the return code of eslint, and so code with warnings can still be
  committed.

Resolves #134.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -10,8 +10,13 @@ module.exports = {
         'vue/setup-compiler-macros': true,
     },
     rules: {
+        '@typescript-eslint/no-unused-vars': 'error',
         'max-len': ['error', {code: 80, comments: 80}],
-        'operator-linebreak': ['error', 'before', {'overrides': {'=': 'after'}}],
+        'operator-linebreak': [
+            'error',
+            'before',
+            {overrides: {'=': 'after'}},
+        ],
         'vue/multi-word-component-names': 'off',
     },
 }
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,7 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run lint
+npm run lint:list || echo 'Files listed above will be reformatted and staged.'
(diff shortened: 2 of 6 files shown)


liam-mulhall:
Okay, I've tested this locally, and it seems to be working! :)

I wish prettier-eslint-cli didn't output "success formatting 1 file with prettier-eslint" when you supply the `--list-different` option, but I guess we can't have everything. (I don't think the file has actually been formatted at this point.)

Other than that minor gripe, which has nothing to do with any of your work, @gwhitney, I think we can merge this.

liam-mulhall (approved):
Changed files also look good!

PR #136, opened 2022-07-07
chore: Adjust vue components to type correctly

### LINKED ISSUE #131: Improve Vue components to be correctly typed
Currently `npm run typecheck` returns > 200 errors, apparently because the Vue components have never previously been being processed by the TypeScript compiler. These errors need to be remedied so that `npm run typecheck` returns no output, and then the infrastructure needs to be enhanced so that the Vue components are typechecked with tsc on every development build of the system, or if that's not feasible, at least on every commit via `husky`.

gwhitney:
This PR modifies the node `dev` and `build` scripts and the pre-commit
  actions to typecheck all code in Vue components. Correspondingly, it
  adjusts all of the components to pass TypeScript type checking. In particular,
  this means boilerplate of a call to Vue.defineComponent() in every component
  and view (per Vue 3 documentation; otherwise TypeScript cannot infer the
  proper types).

  The most significant resulting change was to the mechanism of generating
  unique canvas ids in BundleCard, as the previous mechanism did not allow
  defineComponent() to infer the type of the `cid` member. This commit
  therefore switches to a later proposed mechanism in the same Vue GitHub issue
  referenced by the comment on the previous mechanism.

  One other change worth mentioning is that in several places data properties
  called "id" or "uid" were being used as `v-bind:key` values, from objects
  that had no such properties. These have all been changed to use "name"
  properties, which do exist. That leads to the question of whether there will
  be a problem if two Sequences (say) have the same "name". It seems unlikely,
  since presumably before this change, the "id" properties were all undefined
  and hence identical.

  Otherwise, the changes were essentially all type annotation. Note that there
  was a great deal of new annotation in SeqVizParamsModal as the Paramable
  interface is currently based on a (probably not particularly
  TypeScript-conformant) scheme of string properties determining what type
  an "unknown" value should be cast to. In particular, some of this typecasting
  unfortunately as far as I could figure out has to happen in `v-model`
  directives, forcing me to turn off the `vue/valid-v-model` eslint rule (as
  `value as boolean`, for example, does not qualify as an l-value in eslint's
  judgment).

  Resolves #131.

### DIFF
--- a/src/views/About.vue
+++ b/src/views/About.vue
@@ -35,11 +35,13 @@
 </template>
 
 <script lang="ts">
+    import {defineComponent} from 'vue'
     import {RouterLink} from 'vue-router'
     // To be honest: this check exists only to avoid "unused variable" warning
     if (!RouterLink) {
         console.log('Failed import of RouterLink')
     }
+    export default defineComponent({})
 </script>
 
 <style>
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -18,5 +18,7 @@ module.exports = {
             {overrides: {'=': 'after'}},
         ],
         'vue/multi-word-component-names': 'off',
+        // For the Paramable interface, v-model directives need type annotation
+        'vue/valid-v-model': 'off',
     },
 }
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -4,4 +4,5 @@
 npm run lint:list || echo 'Files listed above will be reformatted and staged.'
 npm run lint:staged
 npm run lint:check
+npm run typecheck
 npm run test:unit
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
     "name": "frontscope",
     "version": "0.3.0",
     "scripts": {
-        "dev": "vite",
-        "build": "vue-tsc --noEmit && vite build",
+        "dev": "npm run typecheck && vite",
+        "build": "npm run typecheck && vite build",
         "preview": "vite preview --port 5050",
         "test:unit": "vitest run",
         "typecheck": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
--- a/src/components/BundleCard.vue
+++ b/src/components/BundleCard.vue
@@ -25,18 +25,13 @@
 </template>
 
 <script lang="ts">
+    import {defineComponent} from 'vue'
     import p5 from 'p5'
     // we need a unique id for each canvas
     // see https://github.com/vuejs/vue/issues/5886#issuecomment-308647738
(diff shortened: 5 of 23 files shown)


src/views/About.vue
```diff
@@ -35,11 +35,13 @@
 </template>
 
 <script lang="ts">
+    import {defineComponent} from 'vue'
```
liam-mulhall, line 38: @gwhitney, could you explain this? I don't understand what's going on here.

liam-mulhall (changes requested):
* I pulled the changes, ran `npm run typecheck`, and I'm not getting any errors! :)
* No errors from `npm run dev`, as expected.
* No errors from `npm run build`, as expected.

# Manual Testing

When I go to the scope, I get the following warnings in the browser console (right click on browser window > Inspect) on both Safari, Chrome, and Firefox:

```
[Warning] [Vue warn]: Invalid prop: type check failed for prop "activeViz". Expected Object, got Null  (20) (chunk-YKE26KAB.js, line 1262)
"
"
" at <BundleManager"
"activeViz=null"
"activeSeq="
SequenceRandom
"bundles="
Array (3)
" ..."
">"
"
"
" at <ToolMain"
"onVnodeUnmounted=fn<onVnodeUnmounted>"
"ref=Ref<"
Proxy
">"
">"
"
"
" at <RouterView>"
"
"
" at <App>"
```

```
[Warning] [Vue warn]: Invalid prop: type check failed for prop "activeSeq". Expected Object, got Null  (19) (chunk-YKE26KAB.js, line 1262)
"
"
" at <BundleManager"
"activeViz=null"
"activeSeq=null"
"bundles="
Array (3)
" ..."
">"
"
"
" at <ToolMain"
"onVnodeUnmounted=fn<onVnodeUnmounted>"
"ref=Ref<"
Proxy
">"
">"
"
"
" at <RouterView>"
"
"
" at <App>"
```

I think going forward we want to avoid logs, warnings, and errors in the browser console.

src/views/About.vue
liam-mulhall: See https://vuejs.org/guide/typescript/overview.html#definecomponent.

gwhitney:
I agree, there should not be any warnings or errors in the browser console. I will investigate and push a commit to fix this. Marking as draft in the meantime.

gwhitney:
OK, didn't realize I had to tell both Vue and TypeScript that null was a possible value for two props of the BundleManager (sadly, there is definitely redundancy between the Vue spec and the TypeScript typing).

liam-mulhall (approved):
I looked the file over and checked the console in Chrome, Firefox, and Safari. Looks good!

Issue #137, opened 2022-07-11
bug: prettier-eslint-cli doesn't ignore `.gitignore`


liam-mulhall:
If you:

1. add `foo` to `.gitignore`,
2. `git add .gitignore`, and
3. `git commit`

You should see something like:

```
> frontscope@0.3.0 lint:list
> prettier-eslint --list-different "**/*.*"

71 files were unchanged

> frontscope@0.3.0 lint:staged
> lint-staged

✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ package.json — 1 file
    ❯ * — 1 file
      ✖ prettier-eslint --write [FAILED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ prettier-eslint --write:
prettier-eslint [ERROR]: prettier formatting failed due to a prettier error
prettier-eslint-cli [ERROR]: There was an error formatting "/Users/liam/Code/l/numberscope/frontscope/.gitignore":
    Error: No parser could be inferred for file: /Users/liam/Code/l/numberscope/frontscope/.gitignore
        at normalize (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier/index.js:7556:19)
        at formatWithCursor2 (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier/index.js:8822:50)
        at /Users/liam/Code/l/numberscope/frontscope/node_modules/prettier/index.js:37229:12
        at Object.format (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier/index.js:37243:12)
        at prettify (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier-eslint-cli/node_modules/prettier-eslint/dist/index.js:128:31)
        at format (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier-eslint-cli/node_modules/prettier-eslint/dist/index.js:95:12)
        at async MergeMapSubscriber.project (/Users/liam/Code/l/numberscope/frontscope/node_modules/prettier-eslint-cli/dist/format-files.js:250:26)
failure formatting 1 file with prettier-eslint
husky - pre-commit hook exited with code 1 (error)
```

liam-mulhall:
I'm not sure if I should have labeled this blocking. Assuming there won't be any further changes, I can just integrate this commit into my `add_cypress` branch and move forward.

PR #138, opened 2022-07-11
fix: ignore `.gitignore`

liam-mulhall:
Modify the lint-staged script so that it uses the `.prettierignore` file
to ignore `.gitignore`.

Resolves https://github.com/numberscope/frontscope/issues/137.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,6 +5,7 @@
 *.ico
 *~
 .husky
+.gitignore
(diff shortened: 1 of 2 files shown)


gwhitney:
I just ran into this myself but I don't understand the change to package.json; why specify that the ignore file is .prettierignore when that is the default? Please explain or remove that change and I will happily merge.

PR #142, opened 2022-07-18
UI tweaks

liam-mulhall:
This PR:

* makes the navbar links white and adds an underline to them when you hover over them
* removes the visualizer descriptions from the homepage
* adds an underline to the homepage 'scope button when you hover over it
* uses the footer color for homepage "Go to the 'Scope" button
* makes the Numberscope background image a link to the 'scope
* adds GitHub and Twitter links to footer
* eliminates space from the acknowledgements link in footer

### DIFF
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -17,63 +17,12 @@
                 >CC0 1.0</a
             >.
         </p>
-        <div>
-            <div class="row" id="visualizerInfo">
-                <div class="col">
-                    <img class="visualizerImg" :src="gameOfLife" />
-                    <h2 class="mt-4 visualizerTitle"><b>Game of Life</b></h2>
-                    <p class="orangeText">
-                        The Game of Life is a cellular automaton model. The
-                        playing field for Game of Life is a grid of cells.
-                        Each cell can be either alive or dead and has eight
-                        neighbors, those being the cells directly touching it
-                        horizontally, vertically, and diagonally. Certain
-                        rules dictate whether or not a given cell will be
-                        alive in the next generation.
-                    </p>
-                </div>
-                <div class="col">
-                    <img class="visualizerImg" :src="shiftCompare" />
-                    <h2 class="mt-4 visualizerTitle">
-                        <b>Shift Comparison</b>
-                    </h2>
-                    <p class="orangeText">
-                        The shift comparison picture of a<sub>n</sub> is
-                        formed by setting the pixel at position (x,y) to be
-                        white if a<sub>x</sub>=a<sub>x+y</sub>, and black
-                        otherwise. The picture captures how a<sub>n</sub>
-                        repeats. Many interesting sequences do not repeat, but
-                        by taking the sequence mod n, that is, taking the
-                        remainders after dividing by n, we get sequences that
-                        do repeat.
-                    </p>
-                </div>
-                <div class="col">
-                    <img class="visualizerImg" :src="turtle" />
-                    <h2 class="mt-4 visualizerTitle">
-                        <b>Turtle Graphics</b>
-                    </h2>
-                    <p class="orangeText">
-                        Imagine a small turtle sitting on a paper. It can move
-                        forward and rotate, and as it moves it leaves a trail
-                        of ink behind it. The visualization of integer
-                        sequences with turtle graphics can thus be constructed
-                        with parameters from the numbers of given sequences.
-                        Most integer sequences will produce mindless
-                        scribbling. Some produce surprising geometric figures.
-                    </p>
-                </div>
-            </div>
-        </div>
     </div>
 </template>
 
 <script setup lang="ts">
     import {RouterLink} from 'vue-router'
     import background4 from '../assets/img/toolpage/background4.jpg'
-    import gameOfLife from '../assets/img/homepage/gof.png'
-    import shiftCompare from '../assets/img/homepage/sc.png'
-    import turtle from '../assets/img/homepage/turtle.png'
 </script>
 
 <style>
@@ -97,15 +46,6 @@
         margin-bottom: 2em;
         font-weight: bold;
         font-size: large;
-    }
-    #visualizerInfo {
-        background-color: #809fff;
-        color: white;
-        margin-top: 2em;
-        padding: 4em;
-    }
-    .visualizerImg {
-        width: 70%;
     }
     @media only screen and (max-width: 600px) {
         #background4 {
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -89,6 +89,9 @@
     #goToScopeButton {
         background-color: #3333ff;
     }
+    #goToScopeButton:hover {
+        text-decoration: underline;
+    }
     #numberscopeDescription {
         margin-top: 2em;
         margin-bottom: 2em;
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -36,7 +36,7 @@
         max-width: 50%;
     }
     #goToScopeButton {
-        background-color: #3333ff;
+        background-color: var(--ns-dark-blue);
         margin-bottom: 1ex;
     }
     #goToScopeButton:hover {
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -6,7 +6,7 @@
         <p id="numberscopeDescription">
             The Online Tool for Visualizing Integer Sequences
         </p>
-        <RouterLink class="btn btn-primary" to="/scope">
+        <RouterLink id="goToScopeButton" class="btn btn-primary" to="/scope">
             Go to the 'Scope
         </RouterLink>
         <p class="mt-4">
@@ -85,6 +85,9 @@
         border: solid black;
         border-radius: 10%;
         max-width: 50%;
+    }
+    #goToScopeButton {
+        background-color: #3333ff;
     }
     #numberscopeDescription {
         margin-top: 2em;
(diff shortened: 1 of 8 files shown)


gwhitney:
Oops, sorry, didn't notice you had filed this, I guess it got lost in the github updates on Slack. (Feel free to ping me when you need a review.) Looking now.

gwhitney:
OK, I offer an additional commit with two points:
1) Until now, the layout has had the footer, with its nice acknowledgments, positioned just _off_ the bottom of the screen -- apparently deliberately. This modifies the layout so that on the Home page on a PC the footer should make it onto the page, so that people have a chance of seeing the info there (otherwise there doesn't seem to be any motivation to scroll down on the Home view). Note this involved tweaking some dimensions; I put vertical dimensions into the "ex" unit which is intended for vertical use; "em" is primarily intended for horizontal use.
2) Three different IDs for the sake of making all the links the same color niggled at me, so I found the CSS specifier that would override Bootstrap's setting.

Please try/review and if there's anything you'd prefer to roll back, let me know.

src/views/Home.vue
```diff
@@ -84,21 +35,18 @@
         border-radius: 10%;
         max-width: 50%;
     }
+    #goToScopeButton {
+        background-color: #3333ff;
```
gwhitney, line 39: This is now a "magic number": A color setting that just _coincidentally_ happens to match the same value as used in Footer.vue. If the design intent is for them to be the same color (which I have no objection to) then the CSS must be reorganized so that they are the same color by virtue of both referring to the same color setting in our source files. We can and should avoid any instances of being obliged to make parallel changes to identical information.
TL;DR: Make this setting DRY. ;-)

src/views/minor/Footer.vue
```diff
@@ -3,10 +3,21 @@
...
+            <i class="bi bi-github"></i>&nbsp;
+            <a class="footer-link" href="https://github.com/numberscope"
+                >GitHub</a
+            >
+            <span id="linkSpacer"></span>
+            <i class="bi bi-twitter"></i>&nbsp;
+            <a class="footer-link" href="https://twitter.com/numberscoper"
+                >Twitter</a
+            >
+        </p>
```
gwhitney, line 20: If you'd like a more compact footer, it would be OK with me to put the GitHub and Twitter links at the beginning followed by the thanks, just with some horizontal space and no paragraph break, so that the thanks would wrap on narrow screens (like phones).  But it's also fine with me to leave as is.

gwhitney (changes requested):
OK, other than reviewing my suggested slight adjustments, just one point that definitely needs revision (the first "magic number" comment); the other item is just for your contemplation. When this is all set, I will be happy to merge

src/views/minor/Footer.vue
liam-mulhall: I wouldn't say I have a _strong_ preference one way or the other, but I kind of like having the thanks be prominent and having the social links separate from the thanks.

src/views/Home.vue
liam-mulhall: Added CSS variables for colors in https://github.com/numberscope/frontscope/pull/142/commits/34c2515d3fee076e543ff791c405a06fe109aad4.

liam-mulhall:
> Until now, the layout has had the footer, with its nice acknowledgments, positioned just off the bottom of the screen -- apparently deliberately. This modifies the layout so that on the Home page on a PC the footer should make it onto the page, so that people have a chance of seeing the info there (otherwise there doesn't seem to be any motivation to scroll down on the Home view). Note this involved tweaking some dimensions; I put vertical dimensions into the "ex" unit which is intended for vertical use; "em" is primarily intended for horizontal use.

I tried to get the footer to stick to the bottom of the page using my limited front-end skills. Unfortunately, it seems like we'll have to live with either some space underneath the footer or the footer being off screen. I don't think it's worth our time (at this point) to fuss with it too much more.

> Three different IDs for the sake of making all the links the same color niggled at me, so I found the CSS specifier that would override Bootstrap's setting.

Thank you.

liam-mulhall:
Re the footer issue: I tried to follow [this tutorial](https://www.youtube.com/watch?v=yc2olxLgKLk) because I've had success with it in the past, but it didn't work.

src/views/minor/Footer.vue
gwhitney: That's fine. Just wanted to float the point because on my screen anyway it feels like vertical space us at a premium.

src/views/Home.vue
gwhitney: The usual pattern I've seen is to provide such var()s with functional designations, like "header background" or "highlight color". That guides where in the implementation to use them, and what the effects will be if they are changed. Looked at another way, you cant'really make "light blue" be fuchsia, but you could make "keywords" be fuchsia,  for example. Or another way, right now it still looks like coincidences that certain elements are both light blue, rather than the design intent captured in the semantics of the items being used. Should we adopt this pattern?

gwhitney:
> tried to get the footer to stick to the bottom of the page using my limited front-end skills

Aha. I truly did not realize that was the design intent since i basically have never seen the footer on screen until i scrolled. I will give it a whirl to make this a sticky footer whose bottom is exactly the bottom of the screen. That should definitely be doable. Then the bundle cards should scroll leaving the header and footer in place, which all sounds fine.

gwhitney:
Is this latest commit I just added the sort of thing you were looking for? I'll be honest I thought the footer taller than the top navbar looked weird but I managed to pare it down to the same thickness, I think it looks OK that way. But @liam-mulhall @katestange is it now weird that the top bar scrolls off the top but the footer doesn't? To me it sort of makes the nav bar look less important than the footer, which I think is reversed. Should we make them both be pinned? Pin just the top nav bar and put the footer at the bottom with the content between its natural height, just make sure it's visible on the home page? Or go back to neither of them pinned, and let everything scroll, just make sure that everything fits on a typical screen on the homepage?  I am no graphic designer, so thanks for your thoughts.

src/views/Home.vue
liam-mulhall: I apologize for not understanding your suggestion initially.

I think I understand it now. You would like the CSS variables to indicate the design intent of the elements they are applied to, correct?

If the CSS variables indicate the design intent of the elements they are applied to, then it seems like less of a coincidence that they have the same color, for instance. Is that the idea?

If that's the idea, I think it's a good one.

I can't really think of what we would name them other than `light-background-color` and `dark-background-color`. The light blue is used in the navbar and the background color for the buttons on the scope when you hover over them. The dark blue is used in the footer and the "Go to the 'Scope" button.

Did I understand your suggestion correctly? Do you have suggestions for the names?

liam-mulhall:
> Is this latest commit I just added the sort of thing you were looking for?

No. Ideally:

* the footer sits underneath the "main content" and never overlaps with any "main content" (I'm not referring specifically to the `<main>` tag here, I mean the main content as in the featured content of the page)
* if the content is too long for the viewport, the footer is pushed down past the end of the content (and therefore offscreen)
* if the content is too short for the viewport, the footer sits flush to the bottom of the viewport

liam-mulhall:
> I'll be honest I thought the footer taller than the top navbar looked weird but I managed to pare it down to the same thickness, I think it looks OK that way.

It looks fine to me.

> is it now weird that the top bar scrolls off the top but the footer doesn't?

Yes.

> Should we make them both be pinned?

I don't like pinned navbars and footers, but whether they're pinned is not a hill I'm willing to die on.

> Or go back to neither of them pinned, and let everything scroll, just make sure that everything fits on a typical screen on the homepage?

This is the approach I'd prefer for now.

gwhitney:
I wasn't pushing for pinning. I misunderstood the design intent. I will change the CSS so that:
1) Nothing is pinned
2) The footer sits at or below the natural content end.
3) If the natural content ends before the full screen height, extra space is added so the footer sits at the very bottom of the screen.
Will post when I have that working.

src/views/Home.vue
gwhitney: OK, I ended up with --ns-foreground for the default text color of the entire app, and --ns-navigation-{background,foreground} for the navbar and other navigation-like elements, such as button-like things when you are over them, and --ns-information-{background,foreground} for message-y type items, like the footer and the BundleManager reports when a sequence or visualizer is chosen. I also tried to make all button-ish things look and behave the same: dark rounded border, no special colors for the content when not over them, but the nav colors when hovering over them. I wasn't entirely sure that the visualizer and sequence choices should be button-like, but they basically seemed to be behaving them that way, so I made them look that way, too.

I am not pretending that this is now good graphic design, but at least I feel like now it's (a) pretty consistent, and (b) will be pretty easy to change if we improve the visual design while keeping it consistent. Let me know what you think.

gwhitney:
OK, I believe the footer is now working the way you intended, and as far as making the CSS more consistent across the site, see my latest comments in the review above. Please look over when you can, and if it looks OK to you I will be happy to merge. If you think some of the styling changes are unattractive or inappropriate, feel free to suggest or implement alternatives. Otherwise I think this is good to go.

liam-mulhall (commented):
The changes look good!

liam-mulhall:
Unless @katestange would like to look these changes over, I think it's ready to merge, @gwhitney.

gwhitney:
I think they really are just tweaks, and I don't see any other concerns, so I am going ahead with the merge. Thanks for initiating these cleanups!

gwhitney (approved):

Issue #141, opened 2022-07-18
Improve skin of docsite


gwhitney:
Per 2022 Jul 18 meeting, we decided to

- [ ] Use the ReadTheDocs theme tweaked as to color-matching the Numberscope navbar
- [ ] Remove the Numberscope "doc home" button at the top left of the page layout
- [ ] Add a navbar at the top, isomorphic (or nearly so) to the Numberscope navbar to create a feel of a consistent site.

PR #146, opened 2022-08-09
doc/ui: Skin documentation site to match Numberscope

gwhitney:
Now the NavBar appears essentially unchanged when switching between
  doc pages (like About or Documentation) and app pages (like the home
  page and the 'scope itself).

  Resolves https://github.com/numberscope/frontscope/issues/141.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ site
 
 # File for storing your local configuration
 *.local
+
+# Emacs backup files
+*~
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,7 @@
 *.png
 *.ttf
 *.txt
+*.generator
 *~
 .gitignore
 .husky
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ in development.
 
 ## How do I use it?
 
-The project will soon be extablishing a public Numberscope server that you can
+The project will soon be establishing a public Numberscope server that you can
 visit and experiment with. In the meantime, see
 [Running from source](doc/running-from-source.md).
 
--- a/doc/css/theme_override.css
+++ b/doc/css/theme_override.css
@@ -1,0 +1,39 @@
+.wy-grid-for-nav,
+.wy-nav-side {
+    top: 20px;
+}
+
+.wy-nav-side .icon-home {
+    visibility: collapse;
+    max-height: 0px;
+    margin-bottom: 0px;
+}
+
+.wy-side-nav-search > a {
+    margin-bottom: 0px;
+}
+
+.wy-nav-side,
+.wy-side-nav-search {
+    background-color: var(--ns-navigation-background);
+}
+
+.wy-menu-vertical p.caption {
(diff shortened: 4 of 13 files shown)


gwhitney:
Don't forget to npm install when pulling this PR.

liam-mulhall:
Looks great on desktop, but when I tried to check how it would look on mobile, I noticed this:

issue

I think we should try to avoid making the website look bad on mobile. It's true that users almost certainly won't want to use the app on mobile, but someone might discover it on mobile (e.g. via Twitter), and we need to make a good first impression.

liam-mulhall:
@gwhitney, would it be difficult to fix this? To check how a site will look on mobile on Firefox, right click the page > Inspect > click the mobile devices icon.

firefox

gwhitney:
Happy to fix. Not sure exactly when, although perhaps I can squeeze it in before I leave, as it should be quick. But what would really help is if you can provide specific feedback on the changes to the portrait view of the page you would like. Is there anything beyond the color mismatch at the top? I guess we should not repeat the word "Numberscope"; should there be anything in place of the second occurrence of Numberscope?

Having a hamburger menu both at the top right and near-top left is not ideal, but I am really not sure what to do about that: one is from the overall-site menu, and the other is from documentation navigation. So my feeling is that we should not sweat this too much as getting the mobile docsite just perfect is probably not worth our effort -- it just shouldn't look too amateurish/broken. That said, if you have any changes for the  portrait arrangement of the docsite navigation controls, especially ones that can be accomplished just with CSS, I am happy to take care of them. But at the moment very specific action items is really helpful given  the pressures on my time.

liam-mulhall:
Perhaps add a media query for mobile that increases the height of the Numberscope navbar so that the docsite navbar is covered up?

liam-mulhall:
I can try to fiddle with it if you're too pressed for time, @gwhitney.

liam-mulhall:
Oh, wait I think covering up the docsite navbar would make it so that you can't navigate using the links in the sidebar on mobile. Hmm... I'll try to think of another solution.

gwhitney:
If you want to fiddle, feel free, or just provide specific thoughts on what to do besides the color matching. I can likely get to it sometime tomorrow or Thursday. But I am  very reluctant to obscure all controls for navigating the documentation; I think there needs to be access to the doc outline.  Perhaps I can just color-match and remove _all_ of the doc nav bar **except** for the doc's hamburger menu, revealing more space for the doc content underneath most of the site navbar, which we would leave alone to match the tool itself?  How does that sound?

liam-mulhall:
> How does that sound?

That sounds good.

gwhitney:
OK, will try to get to this in the next day or two.

gwhitney:
OK I found a few minutes to monkey around with this. How's the latest commit look?

liam-mulhall:
The latest commit looks much better! Thanks @gwhitney. I'm going to merge. (I meant to comment this before merging.)

PR #147, opened 2022-09-01
chore: Switch to pnpm package manager

### LINKED ISSUE #140: Switch to pnpm
This will allow switching to package.json5 and changing the complicated pseudo-scripts therein to ordinary comments. Should also be faster and for those using pnpm in other projects should use less disk space.

gwhitney:
WARNING: do not attempt to pull this PR or operate with it. It depends on an unreleased development version of pnpm, and is expected to be merged in late 2022/early 2023.

Switches this project to exclusively using the pnpm package manager. Benefits include faster performance, reduced disk space usage if you have multiple projects using pnpm (they share dependency code), and the ability to use the more human-readable JSON5 format for the package manifest file. Note that once this is merged, neither npm or yarn will be able to mange the packages for this project; rather, all developers will be obliged to use pnpm.

Resolves #140.


gwhitney:
The necessary updates to pnpm have been merged and are published in pnpm 7.17.1. This PR is ready for review/ merging whenever it becomes convenient.

gwhitney:
Just to make sure everyone sees it, I will also post down here: if you are on this branch, you must use pnpm version 7.17.1 or later, **not npm**. So make sure you have the necessary version of pnpm before pulling this branch. Thanks.

liammulh:
My first instinct is to wait until the students are done, then merge, but we can discuss in today's meeting.

gwhitney:
With the latest commit to this PR, the duplicated section between the two tests in the ci.yaml specification for github actions has gotten rather long/intricate, and should be textually collapsed for maintenance purposes even if it doesn't actually reduce the server work. Should I add a commit here that does that, or wait until this is merged and file a separate PR? Thanks for your thoughts.

liammulh:
As long as we document the fact that we did some de-duplication of the YAML in the commit message that gets merged into main, I don't mind if you add the commit to this PR.

gwhitney:
Sadly this will have to wait until after the TU Delft project, since it now seems clear the team will clone the repository with an npm-managed version, and we definitely don't want to switch during that project. Or in other words: curses, foiled again.

PR #153, opened 2022-09-22
feat(SequenceInterface): add getFactors() method for factorizations

### LINKED ISSUE #116: add getFactors() to sequences
Now the backend can provide factorizations.  We need visualizers to have access to this information through the sequences class, with something like `.getFactors()`.

gwhitney:
Also adds implementations for SequenceFormula for sequence values up to
  at least one million, and for OEISSequenceTemplate for factorizations
  provided by backscope.

  As an illustration of the facility, adds a simple "ShowFactors" visualizer.

  Resolves #116.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -10,7 +10,10 @@ module.exports = {
         'vue/setup-compiler-macros': true,
     },
     rules: {
-        '@typescript-eslint/no-unused-vars': 'error',
+        '@typescript-eslint/no-unused-vars': [
+            'error',
+            {argsIgnorePattern: '^_'},
+        ],
         'max-len': ['error', {code: 80, comments: 80}],
         'operator-linebreak': [
             'error',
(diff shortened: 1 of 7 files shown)


katestange:
I was able to cause the simple factoring visualizer to crash on A275000, even after it had been cached.  I think the problem is that it has very few terms.  I was using the default `this.end = 20` but the visualizer is not checking that the user's ideal endpoint is realistic for the sequence at hand, and calls `getElement()` for an element that doesn't exist.  I don't know if this matters, if this visualizer is not meant to "last" long-term.  For what it's worth, I think this sequence tends to crash some of the other visualizers too.

gwhitney:
Oh, my mistake. I will add a commit to fix this.

katestange:
Also, the bundle card says "Differences" even though I click on the factoring visualizer (and it is doing factoring).

gwhitney:
A275000 has just three terms ??

gwhitney:
OK, both things you noticed should be fixed now. Thanks!

katestange:
Yes, those issues are fixed  Besides those two comments, I was able to get it to display the correct factors as produced by the backend or the simple trial division for sequences by formula.  I don't see any error in what it is serving to the visualizer.  The visualizer also fails fairly gracefully when there's no factorization, either from formula or cached sequences: it just prints nothing.  So I'd be happy to merge.

gwhitney:
Well, as far as I am concerned please merge whenever you're ready.

PR #159, opened 2022-09-29
refactor: Move frontend factoring into its own file and use in Random

### LINKED ISSUE #154: Add getFactors() implementation for SequenceRandom
It's really the only necessary sequence that doesn't yet have factorizations, even once #153 is merged. Implementing this will require refactoring the factorizer from SequenceFormula into its own shared module and using it in both SequenceRandom and SequenceFormula. While that is happening, the implementation should also be improved to cache the factorizations of numbers it sees.

gwhitney:
Resolves #154.


src/sequences/SimpleFactor.ts
```diff
@@ -0,0 +1,219 @@
+/**
+ * Provides a utility function so that front-end sequences can factor their
+ * smallish values; should go away once essentially all sequences are
+ * computed in the backend.
+ */
+
+// Here's a list of enough primes to make sure we can factor every number
+// up to a million. [Sadly, prettier insists on just one bigint per line :-/ ]
+const smallPrimes = [
```
liammulh: I think an ignore directive is appropriate here.

Adding `// prettier-ignore` above `const smallPrimes = [` allows you to have the array on multiple lines like it was in `SequenceFormula.ts` before this PR.

src/sequences/SimpleFactor.ts
```diff
@@ -0,0 +1,219 @@
...
+
+const sureFactorLimit = 1009n * 1009n // smallest number we can't factor
+
+const factorCache: Record<string, [bigint, bigint][] | null> = {
+    0: [[0n, 1n]],
+    1: [],
+    '-1': [[-1n, 1n]],
+}
+
+export default function simpleFactor(v: bigint): [bigint, bigint][] | null {
```
liammulh: In this function, having multiple returns makes sense. Presumably we want this to be as efficient as possible, and it's not like it's a long function, so you can see all of the returns in one reasonably-sized window.

However, is there a general rule of thumb you like to follow when it comes to multiple returns, @gwhitney? I was reading about this on Stack Overflow, and I agree with the advice in https://stackoverflow.com/a/733858. What are your thoughts?

liammulh (commented):
I've added one suggestion and one question.

src/sequences/SimpleFactor.ts
gwhitney: OK, when we settle the other item, I will add the ignore directive and condense the list.

src/sequences/SimpleFactor.ts
gwhitney: I mostly agree with all the discussion in both top answers to that SO question you reference (there are a couple of comments that I think are off base). The devil is of course in the details. 

I definitely feel in this case the short-circuit clause at the top is more readable than putting almost all of the function in an "if not in cache" conditional with a single "return cached value" at the bottom. (I doubt that there's much difference in execution speed of the compiled code.)

 On the other hand, for the two different returns at the bottom, I could go either way; there is a certain amount of non-DRYness in the fact that in each branch, a value is determined, inserted in the cache, and then returned. So if you prefer, in the case that v has been reduced below the sureFactorLimit I could store the factors in the cache but not return, and move the storing of null into the cache into an else block, and then have a single return of factorCache[vrep] at the bottom. The tradeoff there is one extra access to factorCache (negligible compared to the cost of the factoring we just did) vs. the additional clarity of just one bottom return (still with the one short-circuit return at the top). If you think it would read better/be clearer in this instance with that change, happy to do it. Just let me know.

liammulh:
@gwhitney said he'll adopt my suggestion, and he's answered my question.

gwhitney:
OK, made the one change we agreed on.

PR #164, opened 2022-10-06
Grid visualizer

oliviaLovesTheSun:
This is the grid visualizer with some presets, some binary properties, some built-in formulas, and some built in paths.
With the help of Glen, I modified SeqVizParamsModal to allow for properties to be hidden if a certain property isn't selected or if that specific property is selected.

I have made the code clean to the best of my knowledge.


gwhitney:
Thanks for the early PR!  We haven't had a chance to check it out yet, but we were planning to have a session in our next Thursday meeting on some coding principles/practices that will help with Numberscope maintenance as the project progresses.  So let's return to this PR after that session.

gwhitney:
OK, I am going to put this PR in draft mode, and provide some review items. You can address those and make any other changes they inspire, and then when you're done, let us know that you are ready for further review. The next round may be any one of Kate, Liam or me. And just to set expectations, there will likely be a number of rounds of feedback before the PR is ready for final review and merge. Thanks for bearing with the process.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+        && visualizer.amountOfNumbers > 1000
+    ) {
+        visualizer.amountOfNumbers = 1000
+    }
+}
+
+function getNumberFromFormula(
+    visualizer: VisualizerGrid,
+    currentSequenceIndex: number,
+    augmentForSequence: number
```
gwhitney, line 731: As far as I can tell, the purpose of getNumberFromFormula is to transform the input sequence that the visualizer is being run on by a further function. In other words, if the sequence specified in the UI was A000045, the Virahanka-Fibonacci numbers, but you really wanted the squares of the V-F numbers, you would visualize the OEIS sequence A000045 and set the "formula" parameter to N_SQUARED, correct? There is a planned feature that will allow formulas with OEIS references mixed in that will make it unnecessary to allow for such things, because you will be able to write for example `A000045^2`.

The moral of that story, though, is that Visualizers should be set up just to use the values of a sequence as they come (you can add the row augmentation after you get the sequence value, of course).

In the meantime, given that the mixed OEIS/formula sequences are not available at the moment, if you really want this "transformation" ability right now, it's something that's entirely happening on the Sequence side, so the facility should be moved to a new Sequence class rather than mixing Sequence computation into the internals of a Visualizer. Not only will that simplify the code here, but it will allow _anyone_'s visualizer to use a transformed OEIS sequence immediately, not just yours.

If you want to do that, you should modify `src/sequences/OEISSequenceTemplate.ts` to replace the "mod" parameter with a text box to enter an arbitrary (mathjs) formula to transform the OEIS sequence by, and then use mathjs to parse that formula into a function that you will evaluate on the OEIS sequence values -- see `src/sequences/SequenceFormula.ts` for how to do this. You should accept any formula that has a single free variable. That way, to get the squares of the V-F numbers, you would enter A000045 for the sequence ID and say `x^2` for the transformation. And to recover the previous facility of modding out by 5 (say), you would just use `mod(x, 5)` for the transformation. And all of your examples below are easy to write as mathjs formulas, and then you won't anywhere need a laundry list of different possible transformations that is always doomed not to have the one that the person using the visualizer wants in that instance (which is I presume how you ended up with this particular hardcoded collection of formulas to select from).

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+) {
+    let sequenceElementAsNumber = Number(
+        visualizer.seq.getElement(currentSequenceIndex)
+    )
```
gwhitney, line 735: There is a reason that Sequence instances return bigints -- many of the values of OEIS sequences are much too large to be represented by JavaScript's number type (which is a 64-bit IEEE floating-point value). So you should leave sequence values as bigints, and compute using that type to avoid loss of precision. Otherwise, you run into such weird behavior as "all sufficiently large integers are even" (really!). This will get somewhat easier when you move the transformations out to a Sequence class, so there will be less sequence computation done internally here. There may be some difficulties with checking some properties, especially as mathjs does not yet have bigint support (I am working on that), so if you can't implement them all fully within bigint, and are really forced within some of the checking to convert to number, then you must check for "out of range" conditions (just in those cases) and not visualize entries you can't properly compute on -- otherwise you will just be providing misinformation to the viewer.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+const SHADE3 = 'rgb(060, 060, 060)'
+const SHADE2 = 'rgb(040, 040, 040)'
+const SHADE1 = 'rgb(020, 020, 020)'
+const SHADE0 = 'rgb(000, 000, 000)'
+
+//Path variables
+let x = 0
+let y = 0
+let currentDirection = 'right'
+let numberToTurnAt = 0
```
gwhitney, line 58: Why are directions string constants when most other similar collections of options here are enums? Wouldn't an enum for directions be more consistent?

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+    previousDirection: string,
+    currentDirection: string,
+    scalingFactor: number
+) {
+    if (visualizer.visualInfo == VisualInfo.Show_Path) {
+        visualizer.sketch.fill(visualizer.pathAndNumberColor)
+
+        if (iteration != 0) {
+            if (previousDirection == 'left') {
+                visualizer.sketch.rect(
```
gwhitney, line 909: OK, in this function is one example of something that was discussed in the recent coding practices session: a big conditional with very similar code blocks in each case. Could you use something like any of the options we discussed to align it better with KISS and DRY?

gwhitney (changes requested):
OK, I hope these comments are a good start, mostly focusing on the big-picture structure of this visualizer (and the separation of concerns with the Sequence class hierarchy). I did include one or two very detailed items, though, mostly because there was one variable name that in particular made it harder, not easier, for me to understand this code as I was reading it. Let us all know when you feel you've addressed all these points and made any other changes that may occur to you as you go through these revisions, and one of us will do another round of review. Thanks!

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+
+        return status
+    }
+    setup(): void {
+        this.sketch.createCanvas(400, 400)
+    }
+
+    draw(): void {
+        setPresets(this)
+
```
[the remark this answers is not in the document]
gwhitney: OK, I think all the methods that need to be in the class are now. Thanks.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1441 @@
...
+const SHADE0 = 'rgb(000, 000, 000)'
+
+//Path variables
+let x = 0
+let y = 0
+let currentDirection = 'right'
+let numberToTurnAt = 0
+let incrementForNumberToTurnAt = 1
+let whetherIncrementShouldIncrementForSpiral = true
+
```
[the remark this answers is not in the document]
gwhitney: OK, all moved into the class, thanks.

src/visualizers/VisualizerGrid.ts
gwhitney: Sorry if I was too long-winded/unclear above. Visualizers should not be transforming Sequence values. There is a separation of concerns between generating sequence values and visualizing them between the Sequence and Visualizer class hierarchy. So if you need any/all of the transforms you had, they should be moved into the OEISSequenceTemplate class (and implemented with mathjs formulas, since we already have that). Thanks.

src/visualizers/VisualizerGrid.ts
gwhitney: You said you "heard changing to bigints could be done later"... where did you hear that? We want to stick with bigint manipulation insofar as possible in Numberscope, because of the serious pitfalls of trying to compute with large integers in the number class. Please revert as many instances of sequence values from Number to bigint as possible, before this is merged. Thanks.

src/visualizers/VisualizerGrid.ts
gwhitney: Converted to enum, thanks.

src/visualizers/VisualizerGrid.ts
gwhitney: I guess you "solved" this by completely eliminating the functionality, correct? Do you see any other places in your code where a similar observation might be made, though?

gwhitney:
Thanks for a great start! Looking forward to the next commit.

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: I made this correction.

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: I did not end up making the advised change to SequenceForumla.ts. I feel that squaring the numbers is one of the "core ideas" of the visualizer but that all other formulas are unnecessary. I also feel that it would be quite difficult for me to figure out how to make the advised change. It is okay for me just to square sequence values just for one case?

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: I've tried to eliminate other repetition I notice.

src/visualizers/VisualizerGrid.ts
gwhitney: Well, this needs to be moved to the Sequence side for this PR to go in, so I will do that once this branch is up to date with main.

src/visualizers/VisualizerGrid.ts
gwhitney: Will be revisited in further review, I imagine.

src/visualizers/VisualizerGrid.ts
gwhitney: Will look again once this is up-to-date.

gwhitney (approved):

src/visualizers/VisualizerGrid.ts
gwhitney: All right, bigint use seems to be consistent now, thanks!

gwhitney:
OK, I have done the refactor so that you can now do arbitrary transforms of an OEIS sequence, so that should cover all of your original cases and more. Also, I am sorry if we weren't clear before, but just sequence _values_ (and quantities computed from them) need to be bigints. Sequence indices should be numbers, as they will not get up anywhere near the 9 quadrillion limit ;-) Also coordinates for p5 should be numbers, for similar reasons and because that's what p5 is expecting -- it's about using the best type for each data item, not a blanket "always use bigints instead of numbers". So I have changed those back. Also, there was a bug in bigIntSqrt that led to your code thinking 4 was prime, so I replaced it with the new common floorSqrt and that fixed that.

So someone else on the team can take a look at your code and provide further feedback.

However, in addition, and you can work on this in the meantime, there is a misbehavior on the part of this visualizer (so it's not obvious from just a code review): in the Bundle Card view, it draws outside its Bundle Card. I am not actually sure how that is possible (frankly the framework shouldn't allow it to happen), but no other visualizer does it, so there is something that's going on here that shouldn't be. If you could track that down and make sure your visualizer doesn't draw outside the space provided, that would be great. Thanks, and I am sure Liam or Kate will have their review comments soon.

liammulh:
Hey, @oliviaLovesTheSun, I'm going to do a review now. I'll let you know when I've finished and you can look over my feedback.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+    property10Visualization = PropertyVisualization.None
+    property10MainColor = VIOLET
+    property11 = Property.None
+    property11Visualization = PropertyVisualization.None
+    property11MainColor = MUSTARD
+    property12 = Property.None
+    property12Visualization = PropertyVisualization.None
+    property12MainColor = GRAY
+
+    params = {
```
liammulh, line 193: This object runs from line 193 to line 553 (360 lines). It's too big. Do you think we could create this object dynamically (i.e. in a loop) so that it doesn't take up so many lines?

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+            displayName: 'Property 12 color:',
+            required: false,
+            visibleDependency: 'property12Visualization',
+            visibleValue: PropertyVisualization.Color,
+        },
+    }
+
+    checkParameters() {
+        const status = super.checkParameters()
+        if (typeof this.message !== 'string') {
```
liammulh, line 557: This isn't doing anything since `this.message` is always a string.

liammulh (changes requested):
Okay, @oliviaLovesTheSun, could you take a at my questions and comments so far?

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+    Semi_Prime,
+}
+
+enum PropertyVisualization {
+    None,
+    Color,
+    Circle,
+}
+
+enum Formula {
```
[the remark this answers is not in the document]
gwhitney: Actually, my mistake: the Formula enum is no longer used since the transforms were transferred to OEISSequenceTemplate. So it can just be deleted. Olivia, thanks in advance for cleaning up this detail I missed.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+const PURPLE = '#8814cd'
+const CYAN = '#00ffff'
+const MAGENTA = '#ff00ff'
+const VERDANT = '#9BBF30'
+const VIOLET = '#7f00ff'
+const MUSTARD = '#ffdb58'
+const GRAY = '#808080'
+
+const HIGHLIGHT = RED
+const SHADE13 = 'rgb(255, 255, 255)'
```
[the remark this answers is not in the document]
gwhitney: Or you might even want an _array_ of shades rather than an object, i.e. it might be convenient to have the index into the shades be i from 0 to 13 inclusive -- that might make it easier to use the shades later. Just a thought, your call.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import {VisualizerDefault} from '@/visualizers/VisualizerDefault'
+import {floorSqrt} from '@/shared/math'
+
+// prettier-ignore
+const FIBONACCI_NUMBERS = [
```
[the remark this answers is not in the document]
oliviaLovesTheSun: That definitely looks better.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+    2n, 1n, 3n, 4n, 7n, 11n, 18n, 29n, 47n, 76n, 123n, 199n, 322n, 521n,
+    843n, 1346n, 2207n, 3571n, 5778n, 9349n, 15127n, 24476n, 39603n,
+    64079n, 103682n, 167761n, 271443n, 439204n, 710647n, 1149851n,
+    1860498n, 3010349n, 4870847n, 7881196n, 12752043n, 20633239n,
+    33385282n, 54018521n, 87403803n, 141422324n, 228826127n, 370248451n,
+    599074578n, 969323029n, 1568397607n, 2537720636n, 4106118243n,
+    6643838879n, 10749957122n,
+]
+
+const BLACK = '#000000'
```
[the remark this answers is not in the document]
oliviaLovesTheSun: I did not have any reason. Should I stick with RGB since it's more intuitive?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
oliviaLovesTheSun: The array idea sounds good to me.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+    Left,
+    Up,
+    Down,
+    StartNewRow,
+    None,
+}
+
+enum Property {
+    None,
+    Number_Sequence,
```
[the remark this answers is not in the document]
oliviaLovesTheSun: It is not used. I put this here because I wanted to remember that the end goal in being able to have an input sequence and a property sequence. I will remove this.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+    Abundant,
+    Perfect,
+    Deficient,
+    Semi_Prime,
+}
+
+enum PropertyVisualization {
+    None,
+    Color,
+    Circle,
```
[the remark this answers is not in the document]
oliviaLovesTheSun: Circle is not used. I will remove this and then hopefully implement it at some later point.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+const SHADE7 = 'rgb(140, 140, 140)'
+const SHADE6 = 'rgb(120, 120, 120)'
+const SHADE5 = 'rgb(100, 100, 100)'
+const SHADE4 = 'rgb(080, 080, 080)'
+const SHADE3 = 'rgb(060, 060, 060)'
+const SHADE2 = 'rgb(040, 040, 040)'
+const SHADE1 = 'rgb(020, 020, 020)'
+const SHADE0 = 'rgb(000, 000, 000)'
+
+enum Preset {
```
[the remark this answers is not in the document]
oliviaLovesTheSun: I'm using enums because they seem to be a simple solution to get rid of holding important data when it can only have a certain set of values, each of which have specific meanings. This is also valuable because checking .equals("primes") may have a type while === Preset.Primes can't have a typo. Does this answer your question? How else would I do this?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
oliviaLovesTheSun: I will fix this.

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: That sounds like a good idea. I do not know how to do this, but I will give it a shot or reach out for help.

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: I think I need to remove this.message since it is not in the UI.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+}
+
+enum Formula {
+    N,
+    N_SQUARED,
+}
+
+class VisualizerGrid extends VisualizerDefault {
+    name = 'Grid'
+    description = 'This puts numbers in a grid!'
```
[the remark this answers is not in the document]
oliviaLovesTheSun: Absolutely. How long is the description supposed to be. Would this text work?
"This visualizer allows numbers to be put in spirals or in rows and to have certain numbers highlighted depending on certain properties. The most famous use of this can be seen in Ulam's spiral where the natural numbers are put a spiral starting with 41 and then each of the primes are highlighted. Ulam's spiral can be made using this tool, but other interesting visualizations can also be made such as highlighting abundant numbers instead of primes or putting the natural numbers in rows and then highlighting the prime."

Or would this shorter text work?
"This visualizer allows numbers to be put in spirals or in rows and to have certain numbers highlighted depending on certain properties."

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: I would just pick one and stick with it for the sake of consistency. I don't think it matters too much which one you pick.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: I would prefer the longer description. If it looks bad in the UI, that's okay. We can fix the UI. (It needs to be overhauled anyway.)

src/visualizers/VisualizerGrid.ts
liammulh: Sounds good. I will resolve this once you remove it.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: Sounds good!

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: Agreed, array is best.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1214 @@
...
+            if (iteration === this.numberToTurnAt) {
+                this.numberToTurnAt += this.incrementForNumberToTurnAt
+                if (this.whetherIncrementShouldIncrementForSpiral) {
+                    this.incrementForNumberToTurnAt += 1
+                    this.whetherIncrementShouldIncrementForSpiral = false
+                } else {
+                    this.whetherIncrementShouldIncrementForSpiral = true
+                }
+
+                if (this.currentDirection === Direction.Right) {
```
liammulh, line 758: @oliviaLovesTheSun, I think my question in https://github.com/numberscope/frontscope/pull/164#discussion_r1010866782 was too broad.

In this specific code block, wouldn't it be easier to do `if (this.currentDirection === "right")` etc.? Is there a reason you're using a numeric `enum` for your directions instead of just using strings?

gwhitney: I will chime in here that I like enums because they make it clear that there are only/exactly n possibilities (and provide the typo checking that Olivia already mentioned), whereas strings could potentially be very open-ended and leave the reader wondering "hmm, is direction 'southwest' allowed?" -- that sort of thing. There's no ambiguity with what's allowed or not with enums.

liammulh: Ah, the typo checking makes sense now. I was confused because there's a typo in the explanation: "This is also valuable because checking .equals("primes") may have a type while === Preset.Primes can't have a typo."

So we're using an `enum` because:
- it limits the possibilities for directions
- it prevents typos

Is there a clear advantage to using a numeric `enum` to an `enum` with string constants? I'm wondering if perhaps an `enum` like

```
enum Direction {
  Up = "UP",
  Down = "DOWN",
  Left = "LEFT",
  Right = "RIGHT",
}
```

would make debugging easier?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: Looks like this was resolved in https://github.com/numberscope/frontscope/pull/164/commits/56cffaa137d11b8316c983ea4d3b51a3778af252.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: Sorry, @oliviaLovesTheSun, Kate told me there's an even better way to do this. You could use https://p5js.org/reference/#/p5/lerpColor. I would recommend trying to use it.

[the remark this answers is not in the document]
katestange: This would be in line with our goal to reduce repetition (here you have SHADE1, SHADE2, etc.; we can avoid naming variables like that).  You could store just the first and last colour in variables (only two variables, good), and then whenever you want to call a shade, you can use lerpColor() to get the exact shade you need.  This avoids any hardcoding of the "lerping" (linear interpolation) itself (so the human programmer never has to compute +20, +20 etc. and type the series of results).  (A total aside, but this makes me realize one way to encapsulate some of our coding principles is that the computer should do the work, never the programmer.  So the computer should compute an arithmetic progression of colours for you, so you never have to type it in at all. 
 I wonder if Liam and Glen think this makes sense.)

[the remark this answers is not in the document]
gwhitney: I do think it makes sense.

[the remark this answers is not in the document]
liammulh: Olivia used an array in https://github.com/numberscope/frontscope/pull/164/commits/334b3d2d7c717711e920d57876a7ac4dc08a5ef9. I think her plan is to just use the array. Is that correct, @oliviaLovesTheSun?

src/visualizers/VisualizerGrid.ts
liammulh: Gotcha. I think Olivia wants to stick with the way the code is currently written based on my last conversation with her. Is that correct, @oliviaLovesTheSun?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: @oliviaLovesTheSun, it looks like you ended up using this in https://github.com/numberscope/frontscope/pull/164/commits/56cffaa137d11b8316c983ea4d3b51a3778af252, so I will mark this as resolved.

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: Yes.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
oliviaLovesTheSun: Yes.

oliviaLovesTheSun:
The visualizer currently has some issues:

1. It doesn't use getFactors from Numberscope.
2. I had to revert some value from hex to RGB, but I don't know how to do the shades for loop in hex.
3. It still does not change the size of the window it draws in.
4. It does not square values for resetAndAugmentRowAndValueAreSquared.
5. The properties could possibly be condensed more.
6. It is extremely slow when highlighting whether a number is in another sequence.
7. I am curious about feedback on whether the UI is structured in a good way.

The visualizer also has some more aspects I would like to implement:

1. I want to be able to repeat each number the number of times of that number in the grid.
2. I would like to test a bunch of sequences from the OEIS and add a lot more presets.

src/visualizers/VisualizerGrid.ts
gwhitney: Well, I don't think you actually need the properties in an array. You just need to add them to your visualizer and to the params object in a loop. That loop should go in the constructor of your visualizer. The visualizer doesn't currently have a constructor, but you can add one -- be sure to call your super() constructor in it, first thing. Then the loop to add properties will have stuff like
```
for (let p = 1; p <= NumberOfProperties; ++p) {
  propertyName = 'Property' + p;
  propertyVisualizationName = <blah blah>;
  ... etc...
  this[propertyName] = <whatever the initial value of the pth Property should be>
  this[propertyVisualizationName] = <whatever the initial value of the pth Property Visualization should be>
  ... etc...
  this.params[propertyName] = getProperty(...correct arguments go here...)
  this.params[propertyVisualizationName] = getPropertyVisualization(...again, fill in the right arguments...)
  ... etc...
}
```

This has the advantage that you could for now set the NumberOfProperties to something that seems like enough (a half dozen, maybe? Have you ever wanted to visualize more than that at once?) and then it can be bumped up later if anyone ever actually needs more. Or if you already have a use case for 10 or 11, you can just set it to that, and then later it could still be bumped even higher with almost no work.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1067 @@
...
+            this.property3 = Property.None
+            this.property4 = Property.None
+            this.property5 = Property.None
+            this.property6 = Property.None
+            this.property7 = Property.None
+            this.property8 = Property.None
+            this.property9 = Property.None
+            this.property10 = Property.None
+            this.property11 = Property.None
+        }
```
gwhitney, line 432: Similarly, you should be doing setting a whole list of similar properties like this in a loop, not written out.

gwhitney:
> 2\. I had to revert some value from hex to RGB, but I don't know how to do the shades for loop in hex.

If `value` is a number between 16 and 255, then `'#' + value.toString(16).repeat(3)` is the hex color representation for that level of gray. (The 16 argument to .toString is the base.)

>     3\. It still does not change the size of the window it draws in.

It shouldn't _change_ the size of the window, it should _adapt_ to the size of the window it is given...

>     4\. It does not square values for resetAndAugmentRowAndValueAreSquared.

Sorry I don't think I understand this point, not (yet?) familiar enough with your code.

>     5\. The properties could possibly be condensed more.

That would be great. For example, in particular, rather than separate properties for each last digit, couldn't you just have a property for congruent to some constant mod some other constant, and allow the constants to be specified as well? Anything you can do to reduce the number and repetitiveness of your properties will improve your code (and the usability of the visualizer).

> 
>     6\. I am curious about feedback on whether the UI is structured in a good way.

Let's worry about that later, when it's all working and the code is solid.

> 
>     7\. I am curious what properties/sequences others would recommend adding as presets and properties.

I would not add anything else at this stage. There's already _lots_ there and it seems more streamlining is necessary, rather than adding more. The fact that you want to add more perhaps means that many of the properties should be removed and replaced by something more open-ended, like a (mathjs) formula which evaluates to true when the property holds, so like "even" would be `n % 2 == 0` and "odd" would be `n % 2 == 1` and "is a perfect square" would be `sqrt(n) == floor(sqrt(n)`, etc. Then the user could write other properties you haven't even thought of on the fly. Just a thought.

> 
> The visualizer also has some more aspects I would like to implement:

Please do that in a subsequent PR, not this one. Please let's focus on getting this PR merged. There are still some open items above, and then I think Kate may likely want to do a review. Oh and as I mentioned whenever you'd like to, the documentation comments that go to the user guide should be added.

>     1. I want to be able to repeat each number the number of times of that number in the grid.
> 
>     2. I would like to test a bunch of sequences from the OEIS and add a lot more presets.

oliviaLovesTheSun:
The code currently has two issues that I am aware of:
1. The grid doesn't change its size to the grid it's drawing in.
2. The properties such as whether a number is divisible should mostly be removed and instead replaced by allowing the user to type in a formula.

gwhitney:
There seems to still be one open conversation from Liam's review -- I am sure Liam will look into that early this upcoming week. As to the two items you mention:
> 1. The grid doesn't change its size to the grid it's drawing in.

Take a look at Devlin's PR. It also started out not responding to the canvas size, and now it tailors itself nicely. So you should be able to see what's going on there and emulate it in the Grid visualizer.

> 2. The properties such as whether a number is divisible should mostly be removed and instead replaced by allowing the user to type in a formula.

If you decide to go this route for streamlining the "Properties" aspect, the model is the changes to OEISSequenceTemplate that I added to this PR, so look there.

src/visualizers/VisualizerGrid.ts
```diff
@@ -319,7 +319,7 @@ class VisualizerGrid extends VisualizerDefault {
     }
 
     setup(): void {
-        this.sketch.createCanvas(400, 400)
+        this.sketch.createCanvas(this.sketch.width, this.sketch.height)
     }
```
gwhitney, line 323: Individual visualizers do not need to create a canvas, as far as I know. None of the others do. If that's the only thing you have to do in the `setup()` function, you can just eliminate it. But make sure your visualizer re-draws properly in the main display after being shown in the bundle card, and re-draws properly when you go back and forth from the draw view to the bundle card view. My recollection is that if some things are not reset in setup, then the drawing doesn't come out right when you do those things. If you have any problem like that, take a look at what other visualizers do in setup and see if you need to emulate it.

gwhitney:
Other than that, looks like another improvement -- hopefully the visualizer uses the space available better, _and_ you've gotten rid of a "magic number" (i.e. unexplained constant) appearing in the code. So I think just let us know when you've decided how you want to approach the repetitiveness of some of the properties, and have had a chance to do that and take care of the "setup()" issues I just highlighted, and then this should be ready for Kate to review. Excellent.

liammulh (approved):
I don't think there's anything unresolved from my initial review.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+                let a = this.seq.first;
+                a < Math.min(this.seq.last - this.seq.first + 1, 200);
+                a++
+            ) {
+                if (this.seq.getElement(a) == num) {
+                    return true
+                }
+            }
+            return false
+        } else if (property === Property.Prime) {
```
katestange, line 705: I'm trying to understand the big picture on Property.Number_Sequence.  Is this only for presets comparing two sequences?  It looks as if it detects whether num is in this.seq.  But unless we are in such a preset, isn't this always true, because we are only calling hasProperty() on nums from this.seq to begin with?  At least this is true I think if the user chose to use Number_Sequence as a property explicitly.  (Let me know if I'm misunderstanding of course.) Perhaps this capability is only here for use with the presets that compare two sequences, in which case the user shouldn't be able to choose Number_Sequence in the dialog at all, because it doesn't make sense outside the presets.  I wasn't able to use that setting in the dialog and produce any output (seemed to get stuck in an infinite loop).

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            this.scalingFactor / 2
+        )
+    }
+
+    hasProperty(num: bigint, property: Property) {
+        if (property === Property.Number_Sequence) {
+            for (
+                let a = this.seq.first;
+                a < Math.min(this.seq.last - this.seq.first + 1, 200);
+                a++
```
katestange, line 698: If I choose Random Integers in Range, and then I chose Grid Visualizer and alter the "Property" option for the first property to be "Number_Sequence" then I seem to fall into an infinite loop, even if I decrease the Amount of Numbers to a single digit.  It doesn't look as if this loop here can be the cause because you limit it to 200 iterations, but I'm not sure where else to look for why it hangs if I do that.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            this.x + this.scalingFactor / 4,
+            this.y + this.scalingFactor / 4,
+            this.scalingFactor / 2,
+            this.scalingFactor / 2
+        )
+    }
+
+    hasProperty(num: bigint, property: Property) {
+        if (property === Property.Number_Sequence) {
+            for (
```
katestange, line 695: Assuming this for loop is what you are using to compare two sequences in some of your presets, I think it may be somewhat inefficient. There's a javascript method includes() for an array. If you can pre-load the first 200 terms of this.seq into an array just once, then use the includes() method, I'm betting it would be faster? (I don't know for sure.)

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            return isPolygonal(num, 8n)
+        } else if (property === Property.Abundant) {
+            return isAbundant(num)
+        } else if (property === Property.Perfect) {
+            return isPerfect(num)
+        } else if (property === Property.Deficient) {
+            return !isPerfect(num) && !isAbundant(num)
+        } else if (property === Property.Semi_Prime) {
+            return isSemiPrime(num)
+        }
```
katestange, line 765: Does Javascript have the capability to create a dictionary or something like that to call a function from a list of functions based on a key (the value of variable ```property```)?  It would maybe save on this repetition?  This may be a question for @gwhitney or @liammulh too.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+)
+
+/*
+ *   FUNCTIONS TO CHECK FOR PROPERTIES
+ */
+//Taken from Stack Overflow :
+//https://stackoverflow.com/questions/40200089/number-prime-test-in-javascript
+//TODO This should be replaced with the getFactors from Numberscope.
+function isPrime(num: bigint) {
+    if (num === 0n || num === 1n) {
```
katestange, line 863: What's the status of this TODO?  Should it happen before merge?  Another possibility:  maybe ```shared/math``` should have a primality test.  It's pretty simple to implement.  If this is getting left as-is, that's fine for now, I'm just curious of the status for this PR.

katestange: I think the speed of the grid visualizer would increase in many cases if we use this.seq.getFactors (for abundant, perfect, etc. below too).

katestange (changes requested):
Looks good!  Fun to play with.  I may have more comments as I play with it more, but I figured I would put these in for now.

src/visualizers/VisualizerGrid.ts
katestange: Just further experience here to share:  I'm trying A001521 with property Prime.  With 10 or 40 or 100 or 130 terms, and other presets normal, it works nicely.  But with 150+ terms it seems to bog down and not return.  I'm guessing this is the primality test being slow.  It works pretty much instantaneously with property Even for example.

katestange:
Just a question:  I enjoy the colour by last digit aspect, but I that's not accessible via the main parameters, only as a preset?  Since I might want to use a different modulus/base.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            required: false,
+            description:
+                'Note: There is an increased lag time over 10,000 numbers',
+        },
+        startingIndex: {
+            value: this.startingIndex,
+            displayName: 'Starting Index',
+            required: false,
+            description: 'Note: This does not accept negative numbers.',
+        },
```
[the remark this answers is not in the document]
gwhitney: On the other hand, some sequences (including ones in the OEIS) do have perfectly valid negative indices. So why should they be disallowed here?

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+        + ' One use of this tool is to put the natural numbers in a spiral'
+        + " and to highlight the prime ones, which is known as Ulam's "
+        + 'spiral. '
+        + 'Another use of this tool is highlighting properties such as'
+        + ' whether a number is abundant, polygonal, '
+        + 'or in some other sequence.'
+
+    //Grid variables
+    amountOfNumbers = 40000
+    currentSequenceIndex = 0
```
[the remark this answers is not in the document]
gwhitney: Presumably the default should be a square, so maybe 3600 = 60x60? 4096 = 64x64?

src/visualizers/VisualizerGrid.ts
gwhitney: Yes, I completely agree that if an enum for the different Property values is kept as the implementation here, then they should be implemented by an array of functions, so that you can just call `propertyIndicatorFunction[property](num)` rather than needing a very long sequence of if-else-ifs.

src/visualizers/VisualizerGrid.ts
gwhitney: I definitely think that this code should be changed to rely on getFactors() before merge, for all of the reasons you list and just for general DRYness. I do not think shared/math should get an isPrime() test, because it's redundant both with getFactors() and `mathjs.isPrime()` (although the mathjs one only works on numbers at the moment, not bigints). I believe that switching to use getFactors() for primality testing should for A001521 should be rapid.

katestange: @gwhitney you may be right about the isPrime() test in shared/math, but I will make a couple points in its favour and see what you think:  (1) Primality testing is much much faster than factoring and can go much much further than factoring.  So even with full mathjs.isPrime() and getFactors() functionality, we will be missing out on some serious computational capacity.  I can imagine a use case for primality testing on sequences that go beyond our factoring capabilities (e.g. this Grid visualizer, where primes are one of the main inspiration/focus) -- there are plenty of huge sequences we can't factor but could primality test.  Also the algorithm is very simple.  Did I sway you at all?  Another approach would be to add primality testing via pari to the backend (alongside factoring in some way).

gwhitney: It can't be right for us to re-implement high-power primality testing. In the long run, either mathjs will have it, or we will do it in pari in the backend. We already have getFactors(), we could easily add an analogous getPrimality() that returns a boolean or a 0/1 value that it reads from the backend where it's computed by Pari. If you want to file an issue for this latter plan, feel free. I think the current limit on factoring is 2^200. Below that limit the backend should just figure out the primality from the factorization. Above that limit it shouldn't factor, but should just prime test up to some other limit -- and I would have to defer to you Kate as to what a reasonable limit there would be. And we have to decide whether -3 is prime (I assume so.)

katestange: Sounds good.  Issue filed #208

src/visualizers/VisualizerGrid.ts
oliviaLovesTheSun: Number_Sequence puts the natural numbers in a spiral and then highlights the current number sequence.
Any other preset puts the current number sequence in a spiral and then highlights the properties the user selects.
This is a temporary feature because eventually the use should be able to select multiple number sequences with one as the base sequence that goes in a spiral and the rest of the sequences being highlighted if they are contained in the current sequence.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+    constructor() {
+        super()
+        for (let i = 1; i < MAXIMUM_ALLOWED_PROPERTIES; i++) {
+            this.propertyObjects.push({
+                property: Property.None,
+                visualization: PropertyVisualization.None,
+                color: DEFAULT_COLORS[i],
+            })
+        }
+
```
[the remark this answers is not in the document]
oliviaLovesTheSun: I agree. None is a nonsensical property.

gwhitney:
Per today's conversation in ExpMathLab, this should now be considered a n open PR that anyone in the development group can commit to in order to finish up the remaining points in the review and get this into mergeable state. Thanks.

gwhitney:
@oliviaLovesTheSun, before you "sign off" from this project for the semester, please do leave a comment here as to whether you would like your name on the doc page for this visualizer as its original submitter. Thanks!

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+    params: {[key: string]: ParamInterface} = {
+        preset: {
+            value: this.preset,
+            from: Preset,
+            displayName: 'Presets',
+            required: false,
+            description:
+                'Note: If a preset is selected, properties no longer function.',
+        },
```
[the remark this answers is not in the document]
liammulh: @katestange, could you suggest a wording that makes more sense to you?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: Action item: remove the `None` property.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: We should change `amountOfNumbers = 4096`.

src/visualizers/VisualizerGrid.ts
liammulh: For this PR, we are going to temporarily disable anything having to do with multiple sequences.

src/visualizers/VisualizerGrid.ts
liammulh: See https://github.com/numberscope/frontscope/pull/164#discussion_r1040058421.

src/visualizers/VisualizerGrid.ts
liammulh: Create an object like @gwhitney described and in doing so condense the code.

oliviaLovesTheSun:
@katestange 
Regarding the modulus/base idea, I think that's a good idea. If there were a modulus base property, the user could implement coloring by any base.

oliviaLovesTheSun:
@gwhitney Yes! I would like for my name on the doc page for this visualizer to include myself as the original submitter. Thank you!

src/visualizers/VisualizerGrid.ts
liammulh: @gwhitney, this should be resolved in https://github.com/oliviaLovesTheSun/frontscope/commit/7ad0ad2a0c9d7d5f75701a97de50e3d18f681735. I tested this change manually by:

1. Creating a bundle using default params and random integers in range 0-9.
2. Ensuring visualizer was drawn on the initial "bundle card" screen.
3. Ensuring the visualizer was drawn on the "big" screen.
4. Clicking back to the "bundle card" screen and ensuring the visualizer was drawn.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            displayName: 'How path moves about grid',
+            required: false,
+        },
+        showNumbers: {
+            value: this.showNumbers,
+            forceType: 'boolean',
+            displayName: 'Show numbers',
+            required: false,
+            description:
+                'Note: Show_Path and Show_Numbers default to 400 numbers',
```
katestange, line 253: In the dialog box, the user doesn't know what "Show_Path" and "Show_Numbers" refer to.   Alternative text: "If user elects to show numbers, the grid will display only 400 numbers."   (I believe the path functionality is not currently operational.)

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: I think a nice wording is "Presets override Property selections below."  But the better solution (make this an issue for later) is to make properties disappear from the dialog box when a preset is selected.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            required: false,
+            description:
+                'Note: If a preset is selected, properties no longer function.',
+        },
+        amountOfNumbers: {
+            value: this.amountOfNumbers,
+            displayName: 'Amount of Numbers',
+            required: false,
+            description:
+                'Note: There is an increased lag time over 10,000 numbers',
```
katestange, line 233: I would take the word "Note: " off of each of these.  Their method of display in the dialog makes it clear that they are notes, and this will save space, possibly preventing wraps.  This one in particular wraps the text and could be shortened to "Warning:  lag time increases beyond 10,000" which should be shorter.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: One common ui here is that selecting a preset actually just fills in all of the properties as is appropriate for that preset, so they are all visible and could be further edited (and if they are, the value of the "preset" field would clear itself to indicate that the current settings no longer correspond to the preset).

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+            forceType: 'boolean',
+            displayName: 'Show numbers',
+            required: false,
+            description:
+                'Note: Show_Path and Show_Numbers default to 400 numbers',
+        },
+        numberColor: {
+            value: this.numberColor,
+            forceType: 'color',
+            displayName: 'Number color:',
```
katestange, line 258: Since the colour dialogs pop up to the left of the displayName, remove the colon here and in 'Background color:'

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+        startingIndex: {
+            value: this.startingIndex,
+            displayName: 'Starting Index',
+            required: false,
+            description: 'Note: This does not accept negative numbers.',
+        },
+        pathType: {
+            value: this.pathType,
+            from: PathType,
+            displayName: 'How path moves about grid',
```
katestange, line 244: I suggest displayName 'Grid path' (to avoid text wrapping)

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+        super()
+        for (let i = 1; i < MAXIMUM_ALLOWED_PROPERTIES; i++) {
+            this.propertyObjects.push({
+                property: Property.None,
+                visualization: PropertyVisualization.None,
+                color: DEFAULT_COLORS[i],
+            })
+        }
+
+        for (let i = 0; i < MAXIMUM_ALLOWED_PROPERTIES; i++) {
```
katestange, line 282: The later properties (beyond 1) are inside an indented gray box.  This visual doesn't match the heirarchy; Properties should be at the top level, and their colour choices/Visualizations individually indented.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,965 @@
...
+    }
+
+    return propertyVisualization
+}
+
+function getPropertyColor(value: string, index: number) {
+    const propertyColor = {
+        value: value,
+        forceType: 'color',
+        displayName: `Property ${index + 1} color:`,
```
katestange, line 164: remove colon

src/visualizers/VisualizerGrid.ts
gwhitney: Hmm, the trouble here is that making a property optional automatically indents it, and property 2 is optional because it only shows when property one is set, etc. So rearranging this layout would require changes to the general setup of the parameter dialog, which is likely beyond the scope of this PR. I'd suggest we file an issue once this goes in, rather than trying to fix it here.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: OK, I am now working on removing None.

gwhitney:
(That was me rebasing on main before I do anything else.)

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Ah, it took me a while to figure out what was meant here. It's not the "color" of the property that's being set to "None", it's the "Visualization" of the property (which otherwise can either be the primary color, which means the whole grid box, or the secondary color, which means a small centered subbox. I will remove the "None" visualization, which is indeed useless.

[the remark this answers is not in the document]
gwhitney: OK, I believe I have resolved this concern. Can someone pull the latest version of this PR and verify things seem to be working as expected without the Visualization option of "None" and then resolve this conversation if so? Thanks.

gwhitney:
Is it just me, or is this visualizer really slow? I used the natural numbers, and just one property "divisible by seven", and didn't touch anything else and the picture took quite a number of seconds to show up -- a really noticeable delay, which is bad for the user experience. But it seems to me that testing 40000 numbers for divisibility by 7 should not take anywhere near that long... How performant does it seem for the rest of you with those settings?

liammulh:
It is also slow for me. One simple way to test wall clock time would be to put `console.time` and `console.timeEnd` statements in the code and see what the bottleneck(s) is/are.

gwhitney:
Right well once we work through the outstanding items, we can also just take a look for any obvious inefficiencies; and of course, we can also make the Visualizer not draw the entire canvas at once, but instead do a certain number of points per frame. The latter won't speed anything up, but it will vastly improve responsiveness, which is arguably more important than raw speed.

gwhitney:
Let's wait on changing the default number of terms to 4096 until someone has a chance to look at the speed... if it turns out to be easy to speed up, we can leave the default as it is, since as far as I can tell the only motivation for changing the default was the performance.

liammulh:
We're going to close this PR and open a new one based on a branch in this repository.

PR #176, opened 2022-10-20
doc(Pull Requests): Add the snake_case requirement

gwhitney:
Since it is easy to miss in the "working with Git" document,
  which experts may not read anyway, but we would expect
  everyone making a PR to read the PR checklist.


katestange (approved):
This looks fine to me!

Issue #172, opened 2022-10-18
Implementation plan for powmod


gwhitney:
- [x] Gather any existing math utilities into their own module, and add a `powmod(x, y, m)` function that computes the mod-m residue of x^y reasonably efficiently. Presumably this would be implemented for bigints x,y,m -- is there a need for any of them to be any other type?
- [ ] Modify SequenceFormula to evaluate the sequence formula in a context in which it has access to powmod (and document this in the on-line documentation)

Does that sound like the right approach for now? Let me know and let me know about the types, and if so, I will put this on my to-do list after Zulip.

katestange:
Powmod using square-and-multiply should have no problem handling bigints for all the inputs; it's a pretty straightforward polynomial-time thing in the number of digits.  I think bigints is appropriate.

Yay for getting SequenceFormula to do powmod!  Go for it!

gwhitney:
This is bogged down in the possibly eternal wait for mathjs to properly support bigints...

Issue #41, opened 2021-11-12
Abstract visualizers away from being tied specifically to p5


gwhitney:
We envision that there may be network visualizers and/or table visualizers and/or ... that will want to use other rendering libraries besides p5. Right now, the VisualizerInterface explicitly supposes that there will be a p5 sketch associated with the visualizer. It should be divided into a generic interface that is agnostic about what is doing the rendering, and a specialized P5VisualizerInterface that extends the generic one and assumes there's a p5 sketch, so that other sorts of visualizers can be introduced into the hierarchy.

gwhitney:
Example libraries that one might want to use for visualiazation: three.js to produce 3D graphics; vis.js for networks and other data structures, including possibly "ordinary" graphs of sequence values; and/or bootstrap-table for tabular data (and quite possibly VisualizerDifference should really be done with a table library rather than p5).

liam-mulhall:
See https://github.com/numberscope/frontscope/issues/120#issuecomment-1194525172 for how we will proceed.

PR #177, opened 2022-10-20
feat: math utilities for the frontscope

### LINKED ISSUE #163: Make "Intro to working with BigInts" document
Transitioning from JavaScript's Number type to BigInts is a bit painful, and it would help to make a document that provides an introduction to working with BigInts.

- Related: We should also note that JavaScript's mod is different from the actual mathematician's mod. See also https://github.com/numberscope/frontscope/issues/48.

gwhitney:
Adds a new module `shared/math` that has some convenience
  functions for math in frontscope, especially for working
  with bigints.
  Specifically this first revision has safeNumber, for
  converting to a number with a check for loss of accuracy;
  floorSqrt for finding a bigint square root; modulo for the
  mathematician's modulo of bigints; and powmod for calculating
  a power of a number to some modulus (much more efficiently than
  exponentiating and then modding).

  Also provides documentation and tests for the new features and
  a guide to using bigints in frontscope.

  Resolves #163.
  Partially addresses #41.
  Does the first half of #172.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,10 +20,14 @@ applies to other repos as well.)
 1. [Create a fork of the numberscope/frontscope repo](./doc/working-with-git-and-github.md#create-a-fork).
 2. [Clone your fork of numberscope/frontscope](./doc/working-with-git-and-github.md#clone-a-repo).
 3. [Create a branch for your contribution](./doc/working-with-git-and-github.md#create-a-branch).
-4. [Read about basic Git operations](./doc/working-with-git-and-github.md#basic-git-operations).
+4. If you are unfamilar with them or would like a refresher,
+   [read about basic Git operations](./doc/working-with-git-and-github.md#basic-git-operations).
 5. [Push your branch to GitHub](./doc/working-with-git-and-github.md#push-a-branch).
 6. [Read Numberscope's coding principles guide](./doc/code-principles.md).
-7. [Read Numberscope's style guide](./doc/code-style.md).
+7. [Read Numberscope's style guide](./doc/code-style.md). You should also
+   familiarize yourself with frontscope's internal code structure and APIs;
+   the documentation available on these topics is listed in the "Internal code
+   and APIs" section of the navigation bar to the left.
 8. [Work through Numberscope's pull request checklist](./doc/pull-request-checklist.md).
 9. [Submit a pull request](./doc/working-with-git-and-github.md#submit-a-pull-request).
 
--- a/src/shared/__tests__/math.spec.ts
+++ b/src/shared/__tests__/math.spec.ts
@@ -1,0 +1,72 @@
+import * as math from '../math'
+import {describe, it, expect} from 'vitest'
+
+const large = 9007199254740993n
+
+describe('safeNumber', () => {
+    it('leaves small numbers alone', () => {
+        expect(math.safeNumber(7)).toBe(7)
+    })
+    it('makes small bigints into numbers', () => {
+        expect(math.safeNumber(-11n)).toBe(-11)
+    })
+    it('throws on numbers of great magnitude', () => {
+        expect(() => math.safeNumber(10e40)).toThrowError('Attempt')
+    })
+    it('throws on bigints of great magnitude', () => {
+        expect(() => math.safeNumber(-large)).toThrowError('Attempt')
+    })
+})
+
+describe('floorSqrt', () => {
+    it('does nothing on negative bigints', () => {
+        expect(math.floorSqrt(-13n)).toBe(-13n)
+    })
+    it('makes negative numbers into bigints', () => {
+        expect(math.floorSqrt(-15)).toBe(-15n)
+    })
+    it('finds the exact square root of a square number or bignumber', () => {
+        expect(math.floorSqrt(large * large)).toBe(large)
+        expect(math.floorSqrt(16)).toBe(4n)
+    })
+    it('finds the floor of the square root of a number or bignumber', () => {
+        expect(math.floorSqrt(large * large + 10000n)).toBe(large)
+        expect(math.floorSqrt(35)).toBe(5n)
+    })
+})
+
+describe('modulo', () => {
+    it('gives the bigint remainder upon division', () => {
+        expect(math.modulo(7, 5)).toBe(2n)
+        expect(math.modulo(large, 10)).toBe(3n)
+        expect(math.modulo(12, 5n)).toBe(2n)
+        expect(math.modulo(99999999999999999999999999n, 100n)).toBe(99n)
+    })
+    it('always provides a nonnegative residue', () => {
+        expect(math.modulo(-7, 5)).toBe(3n)
+        expect(math.modulo(-large, 100n)).toBe(7n)
+        expect(math.modulo(25, 5n)).toBe(0n)
+    })
+    it('requires a positive modulus', () => {
+        expect(() => math.modulo(77, -7)).toThrowError('Attempt')
+        expect(() => math.modulo(-109n, -6n)).toThrowError('Attempt')
+    })
+})
+
+describe('powmod', () => {
+    it('computes n**exponent modulo the modulus', () => {
+        expect(math.powmod(2, 3, 7)).toBe(1n)
+        const largeCube = large * large * large
+        expect(math.powmod(large, 3, 7n)).toBe(math.modulo(largeCube, 7))
+        expect(math.powmod(6n, 2401n, 7)).toBe(6n)
+        expect(math.powmod(-3n, 343n, 7n)).toBe(4n)
+    })
+    it('requires a positive modulus', () => {
+        expect(() => math.powmod(77, 2n, -7)).toThrowError('must be')
+        expect(() => math.powmod(-109n, 1000, -6n)).toThrowError('must be')
+    })
+    it('handles negative exponents', () => {
+        expect(math.powmod(23, -2n, 18)).toBe(13n)
+        expect(() => math.powmod(12n, -3, 18n)).toThrowError('MUST be')
+    })
+})
--- a/src/shared/math.ts
+++ b/src/shared/math.ts
@@ -1,0 +1,110 @@
+/** md
+## Math utilities for numberscope
+
+The primary resource for doing math inside the frontscope code is the
+[mathjs](http://mathjs.org) package, which this repository depends on.
+The `shared/math` module of Numberscope provides some additional utilities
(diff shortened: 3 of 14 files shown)


gwhitney:
Note that the update to vite and the addition to vite.config.ts were necessitated because the previous setup caused `npm run dev` to barf on bigint literals, cf. https://github.com/jonathanpmartins/v-money3/issues/70#issuecomment-1284503693

src/shared/__tests__/math.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+        expect(() => math.safeNumber(10e40)).toThrowError('Attempt')
+    })
+    it('throws on bigints of great magnitude', () => {
+        expect(() => math.safeNumber(-large)).toThrowError('Attempt')
+    })
+})
+
+describe('floorSqrt', () => {
+    it('does nothing on negative bigints', () => {
+        expect(math.floorSqrt(-13n)).toBe(-13n)
```
liammulh, line 23: What is the rationale for returning the same thing here?

`Math.sqrt(-13)` returns `NaN`.

liammulh: I read the implementation. I see this is just how bigint-isqrt does things.

Do you understand why it doesn't do anything to negative numbers instead of returning `NaN` or something along those lines?

CONTRIBUTING.md
```diff
@@ -20,10 +20,14 @@ applies to other repos as well.)
 1. [Create a fork of the numberscope/frontscope repo](./doc/working-with-git-and-github.md#create-a-fork).
 2. [Clone your fork of numberscope/frontscope](./doc/working-with-git-and-github.md#clone-a-repo).
 3. [Create a branch for your contribution](./doc/working-with-git-and-github.md#create-a-branch).
-4. [Read about basic Git operations](./doc/working-with-git-and-github.md#basic-git-operations).
+4. If you are unfamilar with them or would like a refresher,
+   [read about basic Git operations](./doc/working-with-git-and-github.md#basic-git-operations).
 5. [Push your branch to GitHub](./doc/working-with-git-and-github.md#push-a-branch).
 6. [Read Numberscope's coding principles guide](./doc/code-principles.md).
-7. [Read Numberscope's style guide](./doc/code-style.md).
+7. [Read Numberscope's style guide](./doc/code-style.md). You should also
```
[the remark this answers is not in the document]
gwhitney: Good point. Suggestions for improvement? There is no one file to link to, at least not at the moment :-/

src/shared/math.ts
```diff
@@ -0,0 +1,110 @@
...
+const maxSafeNumber = BigInt(Number.MAX_SAFE_INTEGER)
+const minSafeNumber = BigInt(Number.MIN_SAFE_INTEGER)
+
+/** md
+#### safeNumber(n: number | bigint): number
+
+Returns the `number` mathematically equal to _n_ if there is one, or
+throws an error otherwise.
+**/
+export function safeNumber(n: number | bigint): number {
```
liammulh, line 54: Soft suggestion: I like to list exports at the bottom of the module so that you can see all the exports in one place.

CONTRIBUTING.md
```diff
@@ -20,10 +20,14 @@ applies to other repos as well.)
...
+   familiarize yourself with frontscope's internal code structure and APIs;
+   the documentation available on these topics is listed in the "Internal code
+   and APIs" section of the navigation bar to the left.
```
liammulh, line 30: At the end of line 30, I suggest adding:

"If you're reading this on GitHub, you can view this documentation on our website here: \<link\>. Alternatively, if you'd prefer to stay on GitHub, you can view the documentation in the source code here: \<link\>."

CONTRIBUTING.md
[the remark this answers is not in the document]
liammulh: I added a suggestion [here](https://github.com/numberscope/frontscope/pull/177#discussion_r1001312539).

src/shared/math.ts
gwhitney: Hmm, and I like the self-documenting aspect of in-line exports that tells you as you are looking at the function or class that this is one of the deliverables of the module (as opposed to an internal function). I see the benefits of each approach, and unfortunately you can't do both (I tried, and redundant exports are not allowed). How to resolve the disparity in our respective judgment of the pros and cons of these? Most of the code base uses in-line exports, so we could just keep that for consistency, or on Monday we could have a three-way discussion/decision which way to go and if it ends up with bottom exports put an issue in to convert the rest of the codebase...

I will implement the rest of the needed changes a bit later today, but I wanted to give you some time to contemplate this.

gwhitney: I guess I can crystallize my discomfort with bottom exports by saying that in my eyes, it increases non-locality of information.

gwhitney: Oh I and I think if we decide to prefer in-line exports it should go in the style guide, and if we decide to do bottom-exports we should install and turn on: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/exports-last.md

CONTRIBUTING.md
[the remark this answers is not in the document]
gwhitney: OK, based on your suggestion (thanks!) I have reworked this. Let me know what you think, and then as soon as we resolve the export style thing I hope this can be merged so folks can start to use the bigint utilities. Thanks!

CONTRIBUTING.md
gwhitney: (Good ideas, you can see how I worked this in and let me know what you think.)

gwhitney:
Addition to the PR at Kate's request on Slack.

src/shared/math.ts
liammulh: We're going with in-line exports.

liammulh:
I'm done with my review. Thanks for the changes, @gwhitney.

src/shared/math.ts
```diff
@@ -0,0 +1,135 @@
...
+**/
+export function natlog(n: number | bigint): number {
+    if (typeof n === 'number') return Math.log(n)
+    if (n < 0) return NaN
+
+    const s = n.toString(16)
+    const s15 = s.substring(0, 15)
+
+    return nlg16 * (s.length - s15.length) + Math.log(Number('0x' + s15))
+}
```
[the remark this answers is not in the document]
katestange: Also, I was worrying about the algorithm I recommended and you are using here (modulo this error above).  If Math.log gives only a certain precision, what's the precision guarantee on something like this?  I'm worried it will actually have mostly incorrect digits.  In particular, if any one return from Math.log has only 16 digits of precision, how can some hyper-super-gigantic log that has 100 digits before the decimal have all 100 digits being correct?  It's a result of adding a bunch of limited-precision numbers that all start at potentially different tens-places...  (Sorry if this doesn't make any sense, we can chat in person if needed.)  Unless there's a better algorithm, I think we need to return only the 16 most significant digits -- or possibly less, because rounding errors add up and affect more significant digits eventually -- in any case, return only as many digits as we guarantee accuracy for.  Or if javascript Number is not amenable to that (e.g. how many digits of accuracy does the answer "100" imply, 3 or 1?) we should maybe also return explicitly the number of digits of accuracy the particular answer comes with in each case.

src/shared/math.ts
```diff
@@ -0,0 +1,135 @@
...
+try {
+  const myNumber = safeNumber(9007199254740992n)
+} catch (err: unknown) {
+  console.log('This will always be logged')
+}
+try {
+  const myNumber = safeNumber(9007n)
+} catch (err: unknown) {
+  console.log('This never will be and myNumber will be 9007')
+}
```
katestange, line 32: This doesn't seem grammatical?  Will be what? Logged?

src/shared/__tests__/math.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    })
+    it('requires a positive modulus', () => {
+        expect(() => math.powmod(77, 2n, -7)).toThrowError('must be')
+        expect(() => math.powmod(-109n, 1000, -6n)).toThrowError('must be')
+    })
+    it('handles negative exponents', () => {
+        expect(math.powmod(23, -2n, 18)).toBe(13n)
+        expect(() => math.powmod(12n, -3, 18n)).toThrowError('MUST be')
+    })
+})
```
katestange, line 72: why "must" and "MUST"?

katestange (commented):
Looks great, just some comments on the log functionality.

src/shared/__tests__/math.spec.ts
gwhitney: Dunno, I didn't write those errors, the author of the package I used did. Shrug.

src/shared/math.ts
gwhitney: Correct, but I will insert the word.

src/shared/math.ts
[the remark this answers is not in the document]
gwhitney: The approximation here is that a bigint is its first 15 digits times 16^(num of digits - 15). Since Javascript number has 53 bits of precision and 53 < 15*4, this provides a full "number's worth" of precision in the answer. There is no iterating over blocks, that would be a waste of time. I am confused about your examples, I will add them to the tests and fix them if need be.

src/shared/math.ts
```diff
@@ -0,0 +1,135 @@
...
+**/
+export const powmod = modPow // just need to fix the name
+
+const nlg16 = Math.log(16)
+
+/** md
+#### natlog(n: number | bigint): number
+
+Returns the natural log of the input.
+**/
```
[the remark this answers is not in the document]
gwhitney: It provides the full precision of a JavaScript number, which is the return type; so is that worth saying?

src/shared/math.ts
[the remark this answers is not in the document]
gwhitney: Put all of your examples in the tests and they are all fine and full precision (as verified by Python, which also has arbitrary-precision logs).

src/shared/math.ts
gwhitney: Done.

gwhitney:
OK, have responded to all of your review comments. Let me know if anything else is needed for merge.

src/shared/math.ts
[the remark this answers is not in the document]
katestange: Now that I understand what happened, this is fine as is.

src/shared/math.ts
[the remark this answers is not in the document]
katestange: Yeah it was user error, I went and checked and it was my fault.  Sorry!

Issue #182, opened 2022-11-01
pip command in postinstall script can cause error


liammulh:
When working with @Decos14 today, we encountered the following error:

```
> frontscope@0.3.0 postinstall
> python3 -m venv .venv && cd tools && node pyrun.mjs pip install -U pip && node pyrun.mjs pip install -r requirements.txt
Requirement already satisfied: pip in c:\users\devli\desktop\numberscope\frontscope\.venv\lib\site-packages (22.2.2)
Collecting pip
  Using cached pip-22.3-py3-none-any.whl (2.1 MB)
ERROR: To modify pip, please run the following command:
C:\Users\devli\Desktop\Numberscope\frontscope\.venv\Scripts\python.exe -m pip install -U pip
[notice] A new release of pip available: 22.2.2 -> 22.3
[notice] To update, run: C:\Users\devli\Desktop\Numberscope\frontscope\.venv\Scripts\python.exe -m pip install --upgrade pip
npm ERR! code 1
npm ERR! path C:\Users\devli\Desktop\Numberscope\frontscope
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c C:\Users\devli\AppData\Local\Temp\postinstall-a368a58a.cmd
npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\devli\AppData\Local\npm-cache\_logs\2022-11-01T21_21_51_006Z-debug-0.log
```

The culprit seems to be `node pyrun.mjs pip install -U pip`. We removed that clause from the script and the postinstall script was able to finish without error.

@Decos14's computer runs Windows 11, I believe.

liammulh:
The obvious fix seems to be to change `node pyrun.mjs pip install -U pip` to `node pyrun.mjs python -m pip install -U pip` and see if the problem persists.

liammulh:
But I wonder if creating the venv gives you pip automatically if you have a newish (>=3.5) version of Python...

liammulh:
I suppose it has to.

gwhitney:
The entire point of that clause was to be sure we are always running with the latest version of pip, which venv may not install (as encountering this error amply illustrates). I 100% agree with [the obvious solution](https://github.com/numberscope/frontscope/issues/182#issuecomment-1299369438) and recommend that a PR implementing it be filed.

PR #184, opened 2022-11-03
fix: use `python -m` in `postinstall` script

liammulh:
Fixes https://github.com/numberscope/frontscope/issues/182.


gwhitney:
Works for me. Merging.

PR #191, opened 2022-11-13
Jenny branch

LeongHouHeng:

<hr/>

In this visualizer, the brightness of each circle represents how fast 1/a^n, where a is a number in the sequence, approaches zero as n increases. 
Each prime number is assigned a color, and the color of each circle represents the color combination of its prime factors.

Press the "Left arrow key" to show/hide the color of each prime factor.
Press the "Right arrow key" to refresh.


src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+                    }
+                }
+            }
+        }
+
+        return colorAll
+    }
+
+    //Make sure every value is unique in an array
+    valueDoesNotRepeat(n: number, arr: number[]) {
```
liammulh, line 243: You could use [the JavaScript `includes()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) for this.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+            this.subG.rect(0, 0, 800, 90)
+            this.sketch.image(this.subG, 0, 700)
+        }
+    }
+
+    //change the brightness of the circle
+    changeColor(difference: number, now: number) {
+        if ((now + difference) % 100 < now) {
+            return 100
+        } else {
```
liammulh, line 185: You don't need the else block here since you return in the if block.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+        this.subG.colorMode(this.sketch.HSB)
+        this.subG.noStroke()
+        this.subG.fill(0, 0, 100)
+        this.subG.rect(0, 0, 800, 90)
+        this.subG.fill('black')
+        let tmpX = 0
+        let tmpY = 0
+
+        //Show the color of every prime number
+        for (let i = 0; i < this.primeNum.length; i += 1) {
```
liammulh, line 152: Why not use `i++` here?

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        this.sketch.image(this.subG, 0, 700)
+
+        //this.sketch.noLoop()
+    }
+
+    undrawBox() {
+        if (this.subG) {
+            console.log('undraw')
```
liammulh, line 172: `console.log()` is great for development, but when we deploy our front-end code, it's generally considered unprofessional to have logs in the console.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
+import type p5 from 'p5'
+import {VisualizerDefault} from './VisualizerDefault'
+import type {VisualizerInterface} from '@/visualizers/VisualizerInterface'
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+const colorMap = new Map()
+let countClick = 0
+
+//Show description box by pressing left arrow key
```
liammulh, line 10: At some point, it would be nice if you could include some documentation for your visualizer. Look at how Glen does it in [the Show Factors Visualizer](https://github.com/numberscope/frontscope/blob/main/src/visualizers/VisualizerShowFactors.ts). Then look at how the documentation shows up on the website: https://numberscope.colorado.edu/doc/src/visualizers/VisualizerShowFactors/.

Add a comment like:

```
/** md
# Sequence Color Visualizer

Add instructions and notes for your visualizer here...
**/
```

You can add those `/** md **/` comments anywhere in your code to create your documentation. Notice how Glen splits his documentation up in Show Factors.

If you're not familiar with markdown, here's [a guide](https://www.markdownguide.org/basic-syntax/).

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+        },
+    }
+
+    private current_num = 0
+    private X = 0
+    private Y = 0
+    private subG = this.sketch.createGraphics(800, 90)
+    private boxIsShow = false
+    private primeNum: number[] = []
+    private count_prime = 0
```
liammulh, line 30: Why is this variable `snake_case` whereas most of your other variables are `camelCase`? For more info on variable naming conventions, see [this Wikipedia article](https://en.wikipedia.org/wiki/Naming_convention_(programming)#Examples_of_multiple-word_identifier_formats).

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+            forceType: 'integer',
+            displayName: 'number of terms',
+            required: true,
+        },
+    }
+
+    private current_num = 0
+    private X = 0
+    private Y = 0
+    private subG = this.sketch.createGraphics(800, 90)
```
liammulh, line 27: You should probably store these values (800 and 90) in a variable so you can re-use them. For example, when you set the position of the circle, you type 800 again. 800 is considered a ["magic number"](https://en.wikipedia.org/wiki/Magic_number_(programming)) because it's not immediately obvious what 800 is for, and why it pops up again later in the code.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+        this.sketch.frameRate(30)
+
+        //Set position of the circle
+        this.X = 800 / this.n + 30
+        this.Y = 800 / this.n + 50
+
+        //Obtain all prime numbers from the sequence
+        for (let i = this.seq.first; i < this.n; i++) {
+            if (
+                this.isPrime(Number(this.seq.getElement(i))) == true
```
liammulh, line 55: You could store the result of `Number(this.seq.getElement(i))` in a variable and then re-use it on line 55 and line 57.

src/visualizers/SeqColor.ts
liammulh: There are other instances of magic numbers in your code. Try to avoid using them.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,258 @@
...
+                    this.primeNum
+                ) == true
+            ) {
+                this.primeNum.push(Number(this.seq.getElement(i)))
+                this.count_prime += 1
+            }
+        }
+    }
+
+    draw() {
```
liammulh, line 67: The `draw()` method is long. Generally, long methods are considered bad because they are harder to read for people who aren't intimately familiar with the code. Do you think you could take some of the code in `draw()` and put it in separate functions and then call those functions in `draw()`?

liammulh (changes requested):
@LeongHouHeng, I've added my initial review to your PR. If you have any questions, please reach out.

liammulh (approved):

liammulh:
Looking better! Thanks @LeongHouHeng!

I'll let either @katestange or @gwhitney take a lot if they want to.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            // Monitor keyboard events after finishing drawing
+            // Bug: multiple clicks detected when only one click happened
+            this.keyboardEvents()
+        }
+    }
+
+    keyboardEvents() {
+        // Show description box when the lft arrow key is pressed
+        if (this.sketch.keyIsDown(this.sketch.LEFT_ARROW)) {
```
katestange, line 100: left not lft

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+            if (this.boxIsShow == true && this.subG != null) {
+                this.boxIsShow = false
+                this.undrawBox()
+            } else {
+                this.boxIsShow = true
+                this.drawBox()
+            }
+        }
+        // Refresh
+        if (
```
katestange, line 110: I'm finding the following behaviour:  When I initially create a bundle, it draws the small version with coloured balls.  Then when I click "Draw" and go into the main canvas, it is black/blank.  The left key works to show/hide the legend.  But no circles appear until I hit the right key for "refresh".  So maybe it needs to have an automatic refresh somehow when moving from bundle card to main canvas and back?  Would a change of ```this.firstDraw``` bool go into ```setup()```?  I'm not exactly sure.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+import {VisualizerDefault} from './VisualizerDefault'
+import type {VisualizerInterface} from '@/visualizers/VisualizerInterface'
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+/** md
+# Sequence Color Visualizer
+In this visualizer, the brightness of each circle represents how fast $1/a^n$, where a is a number in the sequence, approaches zero as n increases.
+Each prime number is assigned a color, and the color of each circle represents the color combination of its prime factors. 
+
```
katestange, line 11: Maybe you could explain that the first circle is associated to the first term of the sequence, and the second circle to the second number in the sequence, etc.  You could also say that the fuzziness of the ball is a visual indicator:  a brighter, less fuzzy ball means that term is bigger.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+        for (let x = this.n; x >= 2; x--) {
+            // Calculate the difference
+            //  between 1/(number)^n and 1/(number)^(n-1)
+            const diff = Math.abs(
+                Math.log(
+                    1 / Math.pow(Number(numberNow), x - 1)
+                        - Math.log(1 / Math.pow(Number(numberNow), x))
+                )
+            )
+
```
katestange, line 132: What about negative terms?  Right now it draws black.  I suggest you just take the absolute values of terms, so that the sequences will converge.  If you do this, you should also document this behaviour in the documentation above.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+        for (let x = this.n; x >= 2; x--) {
+            // Calculate the difference
+            //  between 1/(number)^n and 1/(number)^(n-1)
+            const diff = Math.abs(
+                Math.log(
+                    1 / Math.pow(Number(numberNow), x - 1)
+                        - Math.log(1 / Math.pow(Number(numberNow), x))
+                )
+            )
```
katestange, line 131: The visualizer will hang pretty quickly on very large numbers (of which there are plenty in the OEIS).  You might want to have a cut-off for a, where if you see a number bigger than that cutoff, you just "fake it" by drawing a pretty non-fuzzy bright circle (I'm assuming that as the numbers get very big, the eye won't detect the "fake" here; maybe I'm wrong?).

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+
+/** md
+# Sequence Color Visualizer
+In this visualizer, the brightness of each circle represents how fast $1/a^n$, where a is a number in the sequence, approaches zero as n increases.
+Each prime number is assigned a color, and the color of each circle represents the color combination of its prime factors. 
+
+Press the "Left arrow key" to show/hide the color of each prime factor.
+Press the "Right arrow key" to refresh.
+ 
+ **/
```
katestange, line 15: Another thought:  since a lot of sequences (including just the natural numbers) get to a bright full ball fairly quickly (in other words, all balls look pretty bright), maybe we could somehow "tune down" the brightness (e.g. run the given algorithm on a/10 or on log(a)) until we can see more a visual difference between, say natural numbers and squares?

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    keyboardEvents() {
+        // Show description box when the lft arrow key is pressed
+        if (this.sketch.keyIsDown(this.sketch.LEFT_ARROW)) {
+            if (this.boxIsShow == true && this.subG != null) {
+                this.boxIsShow = false
+                this.undrawBox()
+            } else {
```
katestange, line 104: Another thing that might be nice is to be able to toggle the labels (the numbers 1/a^n) on and off with a key.  That way, when we turn them off, we can have a nice pure visual.  You could even have them toggle on/off along with the legend (left key).  So when legend is on, numbers show.  When legend is off, numbers are gone.

katestange (commented):
Just some ideas from playing with this.  I like this visualizer for its distinctive fuzzy visual effect.  To pick up more about the growth of the sequence, I suggested in a comment that you could "tune down" the brightness so there's more variation between small and large terms.  Another thing that might be nice is to be able to toggle the labels (the numbers 1/a^n) on and off with a key.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    drawCircle(numberNow: number) {
+        this.sketch.ellipseMode(this.sketch.RADIUS)
+        let radius = 50
+        let bright = 0
+
+        for (let x = this.n; x >= 2; x--) {
+            // Calculate the difference
```
katestange, line 124: I've noticed what I think is a bug.  Try the visualizer with the sequence from formula n, n+1, n+2, n+3 etc.  The circles should just be "shifting left" between each of these.  But the first circle is not fuzzy, it looks like a single disk, e.g. n+2.  So something funny is happening with the first circle.

katestange (commented):
Just noticed a bug.

gwhitney:
As happened to Clyde, you've managed to check in a commit that contains lint errors. So clearly the Numberscope maintainer group (Liam and Kate and me) have not managed to convey how to create/arrange for a system in which commits must pass lint. This seems to mean that the Husky hooks are not running when you commit in  your environment. This is not your fault, there must be something in our setup instructions and code that is creating an environment it which it is possible to commit without running the hooks. Likely it has to do with something being different on Windows, since Clyde who also had this happen, is running on Windows as well. Any thoughts what it might be, @liammulh ?

gwhitney:
Thanks for the new commit that fixes the lint errors! @LeongHouHeng please let @katestange know when this is ready for her to check on her review comments and resolve them if they have been addressed or provide further feedback if not.

liammulh:
We are going to continue this PR in a branch based in this repository.

PR #195, opened 2022-11-14
docs(PR checklist): reviewer updates contributors

liammulh:

<hr/>

Adds notes about optionally updating the list of contributors at the end of your PR review.


gwhitney:
Hmmm, personally I don't think this should be on the submitter's checklist at all. I think it is just on the reviewer's checklist: just before merge, if it is a new contributor, add a commit adding them to the contributors. I don't think that's optional, we do have a need to track all contributors. The part I was proposing as optional was whether the person wants to mention themselves on the doc page for a new visualizer as the original submitter of that visualizer. So if you agree with these points then some reworking of this PR is needed. If you disagree, let's discuss.

Issue #155, opened 2022-09-22
Documentation error in frontscope run from source instructions


katestange:
In [doc/running-from-source.md](https://github.com/numberscope/frontscope/blob/main/doc/running-from-source.md), in the first section, item 4, it should say to run `npm run prepare` not `npm prepare`.

katestange:
Maybe while we're in there, we should add a note in item 3 that the file `.env.local` ought to be in your root frontscope directory, not some other weird place (as I once did and had to have Liam rescue me...).  I managed to make this error because my other git and npm commands can run regardless of where in the tree I am, so I forgot where I was.

gwhitney:
I am a little confused. `npm install` should automatically run the `prepare` script. Is that not happening for you, Kate? But you are right, if you need to do `prepare` manually, it should be `npm run prepare`, so we might as well fix the docs and your other point too.

katestange:
Nope, I didn't need that command, but I noticed it was wrong.  So it's a separate question whether it should be in the docs at all, and I don't know the answer to that.

liammulh:
https://github.com/numberscope/frontscope/pull/148/commits/59fd25623bc1f78bae3e0e097eb8fbecb53e6926 changes `npm prepare` to `npm run prepare`.

katestange:
I think this is no longer relevant because of #161 ?

gwhitney:
Well, I see a couple remaining points
1) The onboarding doc suggests you edit .env. You shouldn't, you should create a .env.local to override it so that your personal env won't affect the repository when you commit changes.
2) It doesn't say where .env is, as you requested.
3) It doesn't say how to check whether husky ran, as you requested. That's a good idea. But I am wondering why we are bothering to bring up such an unlikely event in a basic onboarding document? I don't think anyone has ever had a problem with husky running, and there is no reason to think there will be a problem. So I propose removing all mention of husky.

If there's agreement on (1-3) I am happy to submit a tiny PR for them.

katestange:
I suggested we simply take out any reference to husky when I reviewed the PR and Liam agreed.

katestange:
And I think the onboarding doc doesn't suggest you edit .env anymore.

gwhitney:
No, #161 is already merged, but all three comments I just made were based on looking at current doc/running_from_source.md in main.  But I did err when I spoke of it as the onboarding doc. I was confusing it with onboarding.md. But I still think about doc/running_from_source.md that

[...]

PR #202, opened 2022-11-15
doc: Small fixes to running-from-source

gwhitney:

<hr/>

Resolves https://github.com/numberscope/frontscope/issues/155.

### DIFF
--- a/doc/husky-pre-commit.md
+++ b/doc/husky-pre-commit.md
@@ -1,0 +1,10 @@
+# Husky pre-commit actions
+
+Assuming that the Git hooks specified by this repository have been
+[installed](running-from-source.md), the
+[Husky](https://github.com/typicode/husky) package will execute the following
+actions prior to every call to `git commit`:
+
+```sh
+{! ../.husky/pre-commit extract: { start: list } !}
+```
--- a/doc/husky-pre-commit.md
+++ b/doc/husky-pre-commit.md
@@ -8,3 +8,7 @@ actions prior to every call to `git commit`:
 ```sh
 {! ../.husky/pre-commit extract: { start: list } !}
 ```
+
+<!-- CUT HERE -->
+
+([Direct link](../.husky/pre-commit) to the Husky pre-commit script.)
--- a/doc/running-from-source.md
+++ b/doc/running-from-source.md
@@ -32,10 +32,10 @@
     (This command also installs Git hooks using
     [Husky](https://github.com/typicode/husky). For a comprehensive list of
     what commands are run when you `git commit` -- typically linting and
-    testing -- see [this file](../husky/pre-commit.txt).)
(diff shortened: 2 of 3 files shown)


gwhitney:
Note the link to the husky file is obliged to have a different form in the repo and on the docsite, since MkDocs does not serve files in directories whose names begin with `.` -- therefore, this PR adds a facility so that Markdown files in our docs can provide alternatives to be visible in the repository and in the docsite via a block of the form:
```
<!-- repository -->
Note repository markdown goes here. It will be visible in the
github markdown display because it is uncommented. Immediately
after this block must be a docsite comment like:
<!-- docsite: This will be extracted into docsite, but is invisible in repo because commented. It must be on one line. -->
```

liammulh:
When I `npm run build`:

```
WARNING  -  Documentation file 'doc/running-from-source.md' contains a link
            to 'husky/pre-commit.txt' which is not found in the
            documentation files.
```

Do we want to ignore this? If so, we need to add a line to the PR checklist saying it's OK to ignore. If not, it needs to be fixed.

The information contained in https://github.com/numberscope/frontscope/pull/202#issuecomment-1316032494 would be useful to have in the docs themselves. It seems a bit weird to have docs about docs, but I don't expect people who write docs to either find this PR comment or read `mkdocs.yml` and grok what is going on in there.

gwhitney:
OK, I will use a different strategy I came up with afterwards for providing this information, that should avoid both this warning and the need for tricky html comments in the docs. Will add a commit for this soon.

gwhitney:
Alright, that should be better. Now there is just a

`<!-- CUT HERE -->`

directive. Everything below that will not appear in the doc site. I've used this and ordinary content extraction to create a little appendix on the doc site that lists the Husky pre-commit actions, whereas on github you just see the extraction marker, but there is an additional direct link to the Husky actions for convenience when viewing it that way. Let me know if this all strikes you as OK. I think it's simpler, and it generates no warnings.

liammulh:
This does seem simpler. Aside from it looking kinda weird when you view the raw markdown, I'm OK with this.

liammulh:
Looks good when I test it locally. I'm OK with merging, but maybe @katestange will have ideas. I'll wait for meeting.

gwhitney:
I agree it's a bit weird. But the raw markdown is maintained by authors; the key for usability is that it should look reasonable in the way it appears when rendered (a) on the docsite, and (b) via Github. Do both look OK to you?

liammulh:
It seems a bit strange on GitHub as well. But I understand that we're running up against the limitations of Mkdocs. As long as the docsite looks reasonable and the important docs look good on GitHub, I think it's fine.

liammulh:
We discussed this PR in last week's meeting, and from what I recall, we're all OK with the tradeoffs involved in this approach. Merging.

PR #212, opened 2022-11-28
Continuous integration

liammulh:

<hr/>

Adds a basic GitHub Actions workflow for continuous integration. This workflow runs lint, and then runs unit tests.


liammulh:
For meeting today:

1. Should this just be on pull_request?
2. Should we see if we can prevent pushes to the main branch? I think GitHub has a way of doing this.

liammulh:
I think this is the rule that we could enable for 2:

rule

gwhitney:
Liam will: change to run lint check (not lint), run only on pull request.

liammulh:
@gwhitney, it now runs `lint:check` and only runs on PR.

gwhitney:
Hmm, most projects I have seen the lint and the unit tests show up as distinct checks, so in our case we would have "2 successful checks". That let's you see right away what bit has gone wrong. Any chance you could set this up to work that way?

liammulh:
Sure, I'll try to get that working.

liammulh:
@gwhitney, do either of https://github.com/dorny/test-reporter or https://github.com/test-summary/action appeal to you?

I don't think GitHub has a built-in feature for summarizing test results, which is annoying. If we want the results displayed in the PR, I think we'll have to generate a personal access token and then use it to create a comment in the PR, which sounds like a lot of work to have some simple results displayed in the PR when you could just go check the results of the test run.

liammulh:
Oh, I see I completely misread your comment!

gwhitney:
Right, just looking for there to be two separate test items, one for the lint test and one for the unit tests, rather than a single item encompassing both. Sorry if I was unclear.

liammulh:
No, you were clear, I'm just a bit tired today.

liammulh:
@gwhitney, I believe this is ready for review again. The YAML is not DRY for reasons outlined below.

If we want two checks (i.e. two jobs) I think there will necessarily be some redundancy. Even if we were to use one of the "code reuse" strategies outlined in [this SO answer](https://stackoverflow.com/a/71570780/15027348), we'd still have the issue where "each job may run on a different machine."[1]

[1] See https://stackoverflow.com/a/65243912/15027348.

At least, that's my understanding based on reading that SO thread, and reading through the GitHub Actions docs. The code reuse strategies allow you to avoid YAML duplication, but they don't avoid duplicating work done for the setup of jobs.

liammulh:
Also, my editor removed some whitespace in files that shouldn't have been touched. Let me fix that...

gwhitney:
OK, I am not too fussed. Best to just get the CI working. In the pnpm branch, I will also have to set up pnpm for each test, at which point it will likely be worth it to avoid the YAML duplication even if it doesn't save any action minutes. But this is fine for now. Merging.

PR #228, opened 2022-12-18
Adding a mouseover feature for scaling on the Y-axis

### LINKED ISSUE #220: Factor Histogram Visualizer: add scale to y-axis
The histogram can be a bit tricky to interpret as no scale is shown on the y-axis. Such a scale should be given and/or there should be a way to see the count of values in each bin.

Decos14:

Resolves #220.

### DIFF
--- a/src/visualizers/VisualizerFactorHistogram.ts
+++ b/src/visualizers/VisualizerFactorHistogram.ts
@@ -23,6 +23,7 @@ class FactorHistogramVisualizer extends VisualizerDefault {
     binSize = 1
     terms = 100
     firstIndex = NaN
+    mouseOver = true
 
     params = {
         /** md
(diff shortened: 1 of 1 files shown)


gwhitney:
The mouseover is nice! However, the height readout seems to go outside the little white box when  it has three or more digits; could you fix the sizing of the box so that doesn't happen? And you could you just add some static labeled tick marks to provide scale on the y-axis (once and for all, not changing with mouse motion or anything)?

Thanks, this is a great start!

gwhitney:
Wow, that was great! The tickmarks on the y-axis are a little unorthodox, but they are fine with me at least for the time being. But there is something very weird going on with the justification of the bin number and the bin height in the mouseover box. When I do 1000 terms of the sequence from formula `n+2`, I get that the bin number and the height align nicely in the rightmost columns, but they become progressively askew as I mouseover columns to the left, until by the leftmost column, the bin number is at least a full numeric character width to the right of the bin height in the popup box. Could you check and see if somehow the bin number is leaking into your positioning formulas? Or maybe there is some other issue going on in the positioning? Thanks!

katestange:
Closed in favour of #271 (which is based on this PR, with improvements).

PR #230, opened 2022-12-20
feat: add grid visualizer

liammulh:

<hr/>

This is @oliviaLovesTheSun's grid visualizer PR. We are continuing it based on a branch in this repository. The original PR, which was submitted from a fork is https://github.com/numberscope/frontscope/pull/164.


src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+    VisualizerGrid,
+    'Puts numbers in a grid.'
+)
+
+/*
+ *   FUNCTIONS TO CHECK FOR PROPERTIES
+ */
+//Taken from Stack Overflow :
+//https://stackoverflow.com/questions/40200089/number-prime-test-in-javascript
+//TODO This should be replaced with the getFactors from Numberscope.
```
liammulh, line 855: Kate: What's the status of this TODO? Should it happen before merge? Another possibility: maybe shared/math should have a primality test. It's pretty simple to implement. If this is getting left as-is, that's fine for now, I'm just curious of the status for this PR.

[...]

Glen: 

I definitely think that this code should be changed to rely on getFactors() before merge, for all of the reasons you list and just for general DRYness. I do not think shared/math should get an isPrime() test, because it's redundant both with getFactors() and mathjs.isPrime() (although the mathjs one only works on numbers at the moment, not bigints). I believe that switching to use getFactors() for primality testing should for A001521 should be rapid.

[...]

Liam: For this PR, we are just going to use `getFactors`.

See https://github.com/numberscope/frontscope/pull/164#discussion_r1026800591 for the original discussion. (It includes discussion of primality testing on backscope.)

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            displayName: 'How path moves about grid',
+            required: false,
+        },
+        showNumbers: {
+            value: this.showNumbers,
+            forceType: 'boolean',
+            displayName: 'Show numbers',
+            required: false,
+            description:
+                'Note: Show_Path and Show_Numbers default to 400 numbers',
```
liammulh, line 251: Kate: In the dialog box, the user doesn't know what "Show_Path" and "Show_Numbers" refer to. Alternative text: "Showing numbers limits the grid to 400 numbers." (I believe the path functionality is not currently operational.)

See https://github.com/numberscope/frontscope/pull/164#discussion_r1046394309 for the original comment.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            required: false,
+            description:
+                'Note: If a preset is selected, properties no longer function.',
+        },
+        amountOfNumbers: {
+            value: this.amountOfNumbers,
+            displayName: 'Amount of Numbers',
+            required: false,
+            description:
+                'Note: There is an increased lag time over 10,000 numbers',
```
liammulh, line 231: Kate: I would take the word "Note: " off of each of these. Their method of display in the dialog makes it clear that they are notes, and this will save space, possibly preventing wraps. This one in particular wraps the text and could be shortened to "Warning: lag time increases beyond 10,000" which should be shorter.

See https://github.com/numberscope/frontscope/pull/164#discussion_r1046397888 for the original comment.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            forceType: 'boolean',
+            displayName: 'Show numbers',
+            required: false,
+            description:
+                'Note: Show_Path and Show_Numbers default to 400 numbers',
+        },
+        numberColor: {
+            value: this.numberColor,
+            forceType: 'color',
+            displayName: 'Number color:',
```
liammulh, line 256: Kate: Since the colour dialogs pop up to the left of the displayName, remove the colon here and in 'Background color:'

See https://github.com/numberscope/frontscope/pull/164#discussion_r1046399515 for the original comment.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+        startingIndex: {
+            value: this.startingIndex,
+            displayName: 'Starting Index',
+            required: false,
+            description: 'Note: This does not accept negative numbers.',
+        },
+        pathType: {
+            value: this.pathType,
+            from: PathType,
+            displayName: 'How path moves about grid',
```
liammulh, line 242: Kate: I suggest displayName 'Grid path' (to avoid text wrapping)

For the original comment, see https://github.com/numberscope/frontscope/pull/164#discussion_r1046400670.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            value: this.backgroundColor,
+            forceType: 'color',
+            displayName: 'Background color:',
+            required: false,
+        },
+    }
+
+    constructor() {
+        super()
+        for (let i = 1; i < MAXIMUM_ALLOWED_PROPERTIES; i++) {
```
liammulh, line 272: Kate: The later properties (beyond 1) are inside an indented gray box. This visual doesn't match the heirarchy; Properties should be at the top level, and their colour choices/Visualizations individually indented.

Glen: Hmm, the trouble here is that making a property optional automatically indents it, and property 2 is optional because it only shows when property one is set, etc. So rearranging this layout would require changes to the general setup of the parameter dialog, which is likely beyond the scope of this PR. I'd suggest we file an issue once this goes in, rather than trying to fix it here.

See https://github.com/numberscope/frontscope/pull/164#discussion_r1046403591 for the original comment.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+    }
+
+    return propertyVisualization
+}
+
+function getPropertyColor(value: string, index: number) {
+    const propertyColor = {
+        value: value,
+        forceType: 'color',
+        displayName: `Property ${index + 1} color:`,
```
liammulh, line 163: Kate: remove colon

See https://github.com/numberscope/frontscope/pull/164#discussion_r1046404099 for the original comment.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
```
liammulh, line 1: (I'm commenting on a file so we can hide the convo if it becomes resolved and PR won't look so cluttered.)

Glen and I had a discussion in the previous PR about performance.

Glen: Is it just me, or is this visualizer really slow? I used the natural numbers, and just one property "divisible by seven", and didn't touch anything else and the picture took quite a number of seconds to show up -- a really noticeable delay, which is bad for the user experience. But it seems to me that testing 40000 numbers for divisibility by 7 should not take anywhere near that long... How performant does it seem for the rest of you with those settings?

Liam: It is also slow for me. One simple way to test wall clock time would be to put `console.time` and `console.timeEnd` statements in the code and see what the bottleneck(s) is(are).

Glen: Right well once we work through the outstanding items, we can also just take a look for any obvious inefficiencies; and of course, we can also make the Visualizer not draw the entire canvas at once, but instead do a certain number of points per frame. The latter won't speed anything up, but it will vastly improve responsiveness, which is arguably more important than raw speed.

Glen: Let's wait on changing the default number of terms to 4096 until someone has a chance to look at the speed... if it turns out to be easy to speed up, we can leave the default as it is, since as far as I can tell the only motivation for changing the default was the performance.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: So far as I can tell, the visualizer is working correctly without the None option in the property visualization.

src/visualizers/VisualizerGrid.ts
gwhitney: Should be done in the latest commit. Someone please look if all is well and resolve this conversation if so. Thanks.

src/visualizers/VisualizerGrid.ts
gwhitney: Should be done in https://github.com/numberscope/frontscope/pull/230/commits/b8fdb093b701217c19d8a13a7886748bafdcf132. Someone please check and resolve this conversation if so.

src/visualizers/VisualizerGrid.ts
gwhitney: [This commit](https://github.com/numberscope/frontscope/pull/230/commits/b8fdb093b701217c19d8a13a7886748bafdcf132) should have changed this to look/read better. Someone please check and resolve this conversation if so.

src/visualizers/VisualizerGrid.ts
gwhitney: Should be done in https://github.com/numberscope/frontscope/pull/230/commits/b8fdb093b701217c19d8a13a7886748bafdcf132. Someone please check and resolve this conversation if so.

src/visualizers/VisualizerGrid.ts
gwhitney: Should be done in https://github.com/numberscope/frontscope/pull/230/commits/b8fdb093b701217c19d8a13a7886748bafdcf132. Someone please check and resolve this conversation if so.

katestange: Looks good, resolving.

src/visualizers/VisualizerGrid.ts
katestange: The wording is changed but on my screen it still wraps.
Screenshot from 2022-12-29 12-44-06

src/visualizers/VisualizerGrid.ts
katestange: looks good.

src/visualizers/VisualizerGrid.ts
katestange: looks good

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+        + ' properties.'
+        + ' One use of this tool is to put the natural numbers in a spiral'
+        + " and to highlight the prime ones, which is known as Ulam's "
+        + 'spiral. '
+        + 'Another use of this tool is highlighting properties such as'
+        + ' whether a number is abundant, polygonal, '
+        + 'or in some other sequence.'
+
+    //Grid variables
+    amountOfNumbers = 40000
```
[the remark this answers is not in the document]
katestange: I did this in commit [49bc3fd](https://github.com/numberscope/frontscope/commit/49bc3fdac8dab290bfa373526c667f3b60a1f8d3) as a way to check that my git setup is working.  I also changed the displayed name of this from "Amount of Numbers" to "Boxes" which seemed better.

src/visualizers/VisualizerGrid.ts
katestange: Oops, I didn't notice this when I was looking for a quick change to test my setup.  Now it's at 4096, so someone will have to change it back if we speed it up.

katestange:
Commit [9ed92d](https://github.com/numberscope/frontscope/commit/9ed92d4eed725f5ed46a5237a87dd26d25f9058a) makes it so the visualizer accepts and behaves appropriately for negative values (negative of a prime is considered prime, but cannot ever be a sum of two squares, etc.).  This involved tidying a little code in the property checking functions (including adding return type annotations), and removing duplicate code (to compute sum of divisors), and catching a possible undefined/fail-to-return on a function that should return boolean.  In future, this code should use seq.getFactors() instead of some of the existing code, but this commit doesn't do that.

katestange:
Sorry, left some changes out of that last commit somehow.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            displayName: 'Amount of Numbers',
+            required: false,
+            description:
+                'Note: There is an increased lag time over 10,000 numbers',
+        },
+        startingIndex: {
+            value: this.startingIndex,
+            displayName: 'Starting Index',
+            required: false,
+            description: 'Note: This does not accept negative numbers.',
```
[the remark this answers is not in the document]
katestange: Question:  with "sequence by formula" does it always begin with index 0?  In this case, the only way to test this is with OEIS sequences that start at -1 (but right now the OEIS connection isn't working).  Looking through the code, however, I don't see a problem with starting at -1 or even earlier.

src/visualizers/VisualizerGrid.ts
katestange: Ok, so I guess this comment is resolved then?  Or we should add an issue and then resolve it?

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+                let a = this.seq.first;
+                a < Math.min(this.seq.last - this.seq.first + 1, 200);
+                a++
+            ) {
+                if (this.seq.getElement(a) == num) {
+                    return true
+                }
+            }
+            return false
+        } else if (property === Property.Prime) {
```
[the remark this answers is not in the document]
katestange: Does "temporarily disable" mean comment out or fully remove, or just set a variable?

src/visualizers/VisualizerGrid.ts
gwhitney: Wow I am glad you checked. That is so weird. We are going to have to do a joint debugging session where you share screen, because it looks perfect on my end. Maybe in our next meeting if the agenda isn't otherwise too full.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Maybe "Grid cells" or "# of grid cells"? It may not immediately be clear what "Boxes" are, at least to my eyes...

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Yeah, I think sequences by formula start at index 0. Why isn't the OEIS connection working? We definitely do need to test this.

src/visualizers/VisualizerGrid.ts
gwhitney: Yeah, I guess that would be best practice, for someone to file an issue that there should be separate control of optionality and indentation, and then close this conversation once that's there.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: I think Liam prefers to just delete the code that wouldn't be used if we are not handling multiple sequences for now, and I am fine with that approach.

src/visualizers/VisualizerGrid.ts
katestange: done, resolving

src/visualizers/VisualizerGrid.ts
liammulh: We're going to change "Shape of grid path:" and "Path in grid:" and "Rows_Reset_Augmented_Rows" to "Offset Rows".

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,951 @@
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
```
liammulh, line 1: This PR needs documentation before merging.

liammulh: We also need to add Olivia's examples to the docs.

liammulh: <details>
<summary>Olivia's Presets Notes</summary>

```
Presets
1 -Natural numbers – Primes in a spiral with Numbers
-Text: These are the natural number put in a spiral, with each of the prime numbers

highlighted.
2 -Natural numbers – Primes in a spiral
-Text: When there are lots of numbers, it becomes evident that the primes form long
diagonal lines. These diagonal lines are quadratic equations, namely x^2 + c, x^2 + 2x + c, x^2 -2x + c,
and x^2 + 4x + c.
3 -Natural numbers – Primes in rows with numbers
-Text: These are the natural numbers put in rows, with each of the prime numbers

highlighted.
4 -Primes in rows
5 -Natural numbers – Primes in rows that reset and augment with numbers
-Text: These are the natural numbers put in rows, but each row the sequence resets with

each number one more than the number above it.
6 -Natural numbers – Primes in rows that reset and augment
7 -Natural numbers – N^2 Primes in rows that reset and augment with numbers
-Text: These are the square numbers put in rows, but each row the sequence resets with

each number one more than the number above it.
8 -Natural numbers – N^2 Primes in rows that reset and augment
-Text: When there are lots of numbers, long diagonal lines become apparent. The diagonal
lines that go up and to the right each have a corresponding diagonal line that goes down and to the right
because numbers repeat when they are arranged this way. The quadratic equation for which each of these
diagonals corresponds is x^2 – x + C
9 -Natural numbers – Abundant numbers in spiral
-Text: These are the natural numbers in a spiral. The abundant numbers are the black
squares, and the non-abundant numbers are the white squares. Abundant numbers are numbers for which
the sum of the factors is greater than the As you can see, the abundant numbers have a lot more structure
than the abundant numbers.
10 -Natural numbers – Abundant numbers in rows
-Text: The structure of the abundant numbers becomes even more evident when they are

put in rows.
11 -Natural numbers – Abundant numbers and primes in spiral
-Text: The abundant numbers are black, and the primes are red. When they are show

together, primes apear less sporadic as they seems to fit around the abundant numbers.
12 -Natural numbers – Polygonal Numbers in spiral
-This is the natural numbers in a spiral. The polygonal number are highlighted. The triangle
numbers are red, the square number are orange, the pentagonal numbesr are yellow, the hexagonal
numbers are green, the heptagonal numbers are blue, and the octagonal numbers are purple. Polygonal
numbers are the number of dots that can be arranged in the shape of that polygon. For example, 6 is a
triangle number because one can form a triangle with three dots at the bottom, two dots in the middle, and
one dot at the top.
13 -Natural numbers – Polyongal Number in rows
-These are the polygonal numbers highlighted in rows.
14 -Natural numbers – N^2 Polygonal Numbers in rows that reset and augment
-These are the polygonal numbers highlighted in rows with the top row being x^2 and each

of the squares being one more than the number above it.
15 -Natural numbers, Rainbow Colored in rows with numbers
-These are the natural numbers in rows wiht each digit colored a different color.
16 -Natural numbers, Rainbow Colored in spiral
-These are the natural numbers in a spiral with each digit colored a different color.
17 -Pi (A000796) – Rainbow Colored in rows (With 1000 numbers)
18 -Pi (A000796) – Rainbow Colored in rows (With 40000 numbers)
-These are the digits of pi in rows, with each digit colored a different color.
19 -Abundant Numbers (A005101) Colored By Last Digit Alternative in a spiral

-When the abundant numbers are put in a spiral and highlighted by their last digit, the

scarcit of odd abundant numbers becomes visually apparent. (With 1000 numbers)
20 -Abundant Numbers (A005101) Colored By Last Digit Alternative in a spiral (With 40000
numbers)

-When the abundant numbers are put in a spiral and highlighted by their last digit, their

structure becomes clealy not random.

Idea for preset that can't be implemented yet:
-Natural numbers – Composite (Dark Green), Abundant (Green), Perfect (Lime Green) Primitive
Abundant (Pink), Weird (Magenta), Highly Abundant (Small Dark Blue), Superabundant (Small Blue),
Colosally Abundant (Small Cyan)

-These are various properties related to divisibilility. A number is abundant if the sum of its
factors is greater than the number itself. The way that these numbers are highlighted emphasizes the
overlap of these properties. All perfect numbers are abundant, and all abundant numbers are composite.
Similarly, all colossally abundant numbers are superabundant, and all suberabundant numbers are highly
abundant. Lastly, all weird numbers are primitive abundant and all primitive abundant numbers are
abundant.
```

</details>

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: We're going to remove the note about not accepting negative numbers. We don't think there's anything blocking us from having negative indices, but we can't test it until https://github.com/numberscope/backscope/issues/50 is resolved.

src/visualizers/VisualizerGrid.ts
katestange: done in commit [cd447c6](https://github.com/numberscope/frontscope/pull/230/commits/cd447c694250537ab7694261e490f9a72832575d) ; if it looks good, please resolve.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: done in commit [cd447c6](https://github.com/numberscope/frontscope/pull/230/commits/cd447c694250537ab7694261e490f9a72832575d) ; if it looks good, please resolve.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: done in commit [cd447c6](https://github.com/numberscope/frontscope/pull/230/commits/cd447c694250537ab7694261e490f9a72832575d) ; if it looks good, please resolve.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: done in commit [cd447c6](https://github.com/numberscope/frontscope/pull/230/commits/cd447c694250537ab7694261e490f9a72832575d) ; if it looks good, please resolve.

src/visualizers/VisualizerGrid.ts
katestange: commit [89598f8](https://github.com/numberscope/frontscope/pull/230/commits/89598f88f60e344fb611a51c28af0c9861c69a0d) puts in the documentation, with the exception of the images of the examples (so this isn't resolved yet, but closer)

src/visualizers/VisualizerGrid.ts
gwhitney: Check!

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Looks good indeed.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: The input box label is OK, but now it is inconsistent with the warning; maybe the warning text should change to: "Warning: Over 10,000 cells causes display lag." or something like that.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Looks good indeed.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: Also, should we mention that this will be rounded to the nearest square number?

src/visualizers/VisualizerGrid.ts
gwhitney: OK, took a look. Some thoughts/suggestions; some of these are a matter of taste, so feel free to take or leave. 

1. I think the parameters should go before the examples, esp since the example list will be very long once the pictures are there.
2. I think the credit should go after the examples, in a general spirit of modesty.
3. Formulas in examples 8 and 14 not yet set with MathJax
4. The headers break weirdly in examples 19 and 20
5. Do we really need 20 examples? Should we do a bit of judicious editorial pruning? (If they are all really cool, of course they all can stay.)
6. The link to the properties section is broken, and there appears to be a stray attempt at putting an ID on that heading.
7. The names of the properties in camelCase are not how they appear in the dialog. They should match.
8. I _think_ that the Rows_Offset re-starts the sequence from the specified starting index, but adds one more each row to the _values_ of the sequence. That should be checked and if I am right, the documentation should be updated.
9. I suggest we use "you" rather than "the user" in these docs, it's more friendly, I think.
10. Maybe want to mention that the "numberColor" parameter is only shown when show numbers is checked.
11. The header for "backgroundColor" breaks weirdly.
12. I think there is no longer a "None" property, since it was useless anyway.
13. Documentation for the "Property N Visualization" and "Property N Color" parameters is missing.

katestange: Thanks!  Some quick clarifications/questions: 
(2) and in general after the parameters (for other visualizers that have no examples section) -- so in the source file at the very end?
(5) good point; I was just copying the original list.  I'll wait until the images are in and see if some can be combined/eliminated
(6) odd, I thought I fixed that!
(7) I was following the format of the ShowFactors visualizer when I did this, which matched the internal variables.  (Or maybe ShowFactors actually has the same internal variables as dialog names?)
Thanks again, very helpful.

katestange: Also, @gwhitney there's a question in Slack about where to store the image files for docs in the directory tree.  Should they go in with the source (where the source for the relevant doc is) or outside ```src``` over in ```assets``` or something?

gwhitney: On (2), that's my feeling.
On (6), it might again stem from the fact that only the "staged" changes to your files get committed; the current versions of files in your working directory might differ from the "staged" versions.
On (7) dunno what ShowFactors does, I may have erred. But clearly documentation should match the identifiers visible to the person reading the docs.

On location of image files, seems like there's a precedent for src/assets/(VISUALIZER_NAME); I don't see any particular reason to change that pattern now, but if there's a reason to prefer a different location, I am not wedded to it, either.

katestange: [newest commit](https://github.com/numberscope/frontscope/pull/230/commits/c2de3a2778875ce18ec04e6ce65438a30ae33691) should handle all of this, I hope!

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: done

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1191 @@
...
+    Sum_Of_Two_Squares,
+    Abundant,
+    Perfect,
+    Deficient,
+    Semi_Prime,
+}
+
+enum PropertyVisualization {
+    Fill_Box,
+    Small_Box,
```
katestange, line 134: My newest commit changed the names here, so it is "Property Style" and you choose from "Fill_Box" and "Small_Box" (instead of Property Visualization and you choose from Primary and Secondary).  This felt much more self-explanatory to me once I was writing the docs.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,959 @@
...
+            return isPolygonal(num, 7n)
+        } else if (property === Property.Octagonal_Number) {
+            return isPolygonal(num, 8n)
+        } else if (property === Property.Abundant) {
+            return isAbundant(num)
+        } else if (property === Property.Perfect) {
+            return isPerfect(num)
+        } else if (property === Property.Deficient) {
+            return !isPerfect(num) && !isAbundant(num)
+        } else if (property === Property.Semi_Prime) {
```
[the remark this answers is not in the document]
katestange: The problem with this is that currently the Property ```enum``` is used to populate the drop-down menu automatically.  So I'm not sure what effect changing this would have on the drop-down menu... advice?

gwhitney:
I am getting a dozen or more warnings from `npm run build` in the mkdocs section about VisualizerGrid.md containing links to missing files. However, when I do `npm run preview`, all of the images actually show up on the documentation page, **except** for the very first one at the top. So the first link or image needs to be fixed, and the warnings need to be eliminated (it seems as though since there are not warnings for FactorHistogram, there must be some workable way of specifying image links that won't generate warnings).

gwhitney:
Of course if you would like me to take a stab at such fixes, just let me know.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: I should clarify it is done in the docs, not in the dialog.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: The key here is to replace a huge if/then/else sequence with a single call to a function selected from a dictionary of implementations of the different properties.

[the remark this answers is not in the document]
gwhitney: (This is not related to the drop-down in the dialog, but rather to the cleanliness of the code implementing the properties.)

src/visualizers/VisualizerGrid.ts
gwhitney: Will look again when the images/warnings are fixed.

src/visualizers/VisualizerGrid.ts
gwhitney: Maybe the input label for this in the dialog could just be "Display as:"? With the options "Fill_Box" and "Small_Box", it should be pretty understandable, and would avoid the need to repeat "Property N"... And should that first option actually be "Fill_Cell"? (To keep the language around cells consistent?) For that matter, should the label on "Property 1 Color" be changed to just "Color" (since after all, it is in the Property 1 "section")?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: No, I think it is related?  The ```Property``` enum is automatically used to populate the drop-down menu.  So if I change it into something else, where will the drop-down menu populate from?  (I understand the code cleanliness issue, but this ```Property``` enum is playing two roles, so if I change it into functions, I'm not sure what happens to the drop down.)

katestange:
> I am getting a dozen or more warnings from `npm run build` in the mkdocs section about VisualizerGrid.md containing links to missing files. However, when I do `npm run preview`, all of the images actually show up on the documentation page, **except** for the very first one at the top. So the first link or image needs to be fixed, and the warnings need to be eliminated (it seems as though since there are not warnings for FactorHistogram, there must be some workable way of specifying image links that won't generate warnings).

That's weird, the first image works fine on my computer.  I will note that I sometimes have to rebuild the docs to get the images to load, which I find weird.... but that's using `npm run doc:serve`.  Also, an idea: is it because the path to doc files that mkdocs uses isn't including the new directory?

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: The enum should still exist. There should just also be an object whose keys are Property enum instances and whose values are the corresponding functions for computing the property. Then that whole if/then block becomes something like `propertyIndicatorFunction[property](num)`. Nice, eh?

src/visualizers/VisualizerGrid.ts
katestange: Oh, nice idea!  I'm not sure about "Fill_Cell" only because then "Small_Box" is less understandable.  I thought of "Fill_Cell" and "Mark_Cell" but I thought that wouldn't be clear either.  Maybe "Fill_Cell" and "Fill_Center"?  I don't find any of these options super great.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: Ok great, that's the advice I was looking for.

src/visualizers/VisualizerGrid.ts
gwhitney: Maybe "Fill_Cell" and "Box_In_Cell"? Alternatively, we could change it to "Indicator size:" and have that be a decimal from 0.1 to 1.0, for the fraction of the size of the cell taken up by the (centered square) indicator of this property. That way you _could_ do three overlaid, for a large enough screen and few enough cells... I am just brainstorming, please go with whatever seems reasonable.

katestange: > Of course if you would like me to take a stab at such fixes, just let me know.

I fixed the missing image file now.  But the warnings I took a stab at and wasn't able to figure out why the histogram would be different (the code all looks the same to me).  If you could take a stab at the warnings, that would be great.  In the most recent commit, I also fixed the small things to do with the params dialog, and I changed a name again:  I replaced Rows_Offset with Rows_Augmented because this is closer to the meaning of that option, now that you explained it to me.

katestange: Why did my reply end up in this comment chain?  Github must have got confused.  I was responding to the docs warnings comment.

katestange: > Maybe "Fill_Cell" and "Box_In_Cell"? Alternatively, we could change it to "Indicator size:" and have that be a decimal from 0.1 to 1.0, for the fraction of the size of the cell taken up by the (centered square) indicator of this property. That way you _could_ do three overlaid, for a large enough screen and few enough cells... I am just brainstorming, please go with whatever seems reasonable.

That's a cool idea.  Maybe I'll try that.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
gwhitney: OK this is fixed. Resolving.

gwhitney:
>  If you could take a stab at the warnings, that would be great.

Happy to. Just let me know a time when you are not actively working on it, so that we don't have to deal with merges and the like (i.e., you can push everything you have, then I can do a pull, commit, and push, and then you can pull that).

gwhitney:
Also, I don't know why your comment on images ended up in the wrong conversation, sorry. But I do see all of the images now. I did want to mention one other thing: in the spots where images wrap to be on two "lines" there is no vertical whitespace between vertically adjacent images, making them a bit hard to see near their common boundary. Some vertical whitespace should be added. I can do it when I work on the warnings, if you like.

katestange:
> Also, I don't know why your comment on images ended up in the wrong conversation, sorry. But I do see all of the images now. I did want to mention one other thing: in the spots where images wrap to be on two "lines" there is no vertical whitespace between vertically adjacent images, making them a bit hard to see near their common boundary. Some vertical whitespace should be added. I can do it when I work on the warnings, if you like.

If you have time now, that would be awesome.  I'm working on something else for a bit.

gwhitney:
The third image in the "digit colourings" section of the documentation can't be the digits of pi in rows; the color pattern does not match 314159265358979 nor 14159265358979. Should we delete it or attempt to produce a correct image? It also makes me suspicious of the fourth image, which is supposed to be a zoomed out version of the first. So same question about it. Also, I am going to change to the American spelling of "color", OK?

gwhitney:
OK, the image links (there were just some extra `../`s) and spacing (just needed `margin:` instead of `margin-left:` and `margin-right:`) should be fixed. I also tweaked some residual wording/grammar things. But the two content issues in the images I mentioned above: the digits of pi image cannot be correct, and the two digits of abundant numbers images use different settings, are not resolved: they need to be. Anyhow, I am done working on this for today so anyone else should feel free to take the next crack.

src/visualizers/VisualizerGrid.ts
katestange: Ok, I'm going with Fill_Cell and Box_In_Cell rather than adding functionality at this point; would like to get this PR done.

katestange: done; please take a look

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
katestange: Done; please take a look.  The only issue is that there's an unused variable lint rule I sidestepped because I don't know how to solve it.  Advice on resolving that would be appreciated.

src/visualizers/VisualizerGrid.ts
katestange: This conversation got stale and github hid it.  I just discovered it.  I'm responding so it is fresher and visible.

katestange:
The most recent commit fixes the digit images, renames to use Box_In_Cell and refactors to create a dictionary from enum Property to functions that check the property.  As mentioned in that conversation, I had one lint error I couldn't resolve.  I bypassed it so I could commit and ask for help with that.  Done for tonight! :)

src/visualizers/VisualizerGrid.ts
katestange: Regarding this issue, right now the way the code is set up, property checking (e.g. `isPrime()`)  takes in the *value* passed as a `bigint`, not the *index* of the sequence term.  So to use `getFactors()` would require more refactoring than it would seem at first.  That is, the code would have to be changed so the property checking functions all take in an index, not a value.  This seems reasonable to me; seem reasonable to you?

katestange: Separate issue:  this will mean writing temporary code to use `getFactors()` to primality check.  It will be changed again once the backend can serve up factoring and the sequence module has a function `isPrime()`.   I guess I'm in favour of this (writing a little temporary code to finish this PR instead of waiting for the extra functionality), but wanted to double check that's the intention.

gwhitney:
Two options on the lint issue: since in javascript it's ok to call a function with extra arguments, perhaps you could widen the type of the indicatorFunction object so that `()=> false` would be an ok value for the None key. If you can't get that to work, then by naming the argument with a name starting with an underscore, like `_ignored`, lint will know you intend for it not to be used and won't complain. Either of those is better than the comment.

katestange:
The most recent commit refactors so `isPrime()`, `isSumOfSquares()` and `isSemiPrime()` use `getFactors()`.  It required lots of refactoring.  In the process, I found a subtle bug where it was offset the indices, so all the pictures were off by one square, and fixed that.  This should vastly speed up the drawing for those properties.  I also renamed an internal variable to be simpler.  We need to check that the example pictures are correct (because of the offset bug) but this commit doesn't do that.

src/visualizers/VisualizerGrid.ts
katestange: the most recent commit does this, please take a look

src/visualizers/VisualizerGrid.ts
katestange: Hi all, so I took a look at this.  I tried to move things from the iteration into the setup.  I think this helped a bit (e.g. no longer drawing background squares each time, but just setting a background first).  I put a console message at the beginning of the iteration, and observed something very weird.  I used Glen's suggestion of divisible by 7 for 40000 numbers.  The message appeared immediately (meaning the `setup()` was fast) and then there was a long delay.  Then the message appeared _again_ and it drew almost immediately.  What is going on?  It appears the thumbnail card is calling `draw()` twice (there's a `noLoop()` in `draw()`), and the first time is taking the long time, but the second time is fast.  Going to the big canvas is fast too

katestange: Ok, update:  the bottleneck is `seq.getElement()` in Glen's example.  When I made a dummy to replace that (just making all terms `=5n`), it was 10x faster.  So that explains why the first time through `draw()` is long and subsequent times are fast.  I tried using sequence A000005, which has 100000 terms in the OEIS, to see if `seq.getElement()` is faster there.  But I can't seem to get the backend to cache it all.  I'll check back later.

katestange: Ok, I was finally able to verify with a023900, which has 20000 terms, that the visualizer itself is twice as fast on an OEIS sequence than on a sequence by formula; that's what's slow.  I think this isn't an issue with this PR, so I suggest we discuss and file an issue, then resolve this convo.

src/visualizers/VisualizerGrid.ts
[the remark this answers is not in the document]
liammulh: It looks like this was done in https://github.com/numberscope/frontscope/pull/230/commits/f23f1284c6ce2320bbc9e669032e5ca11f7ac73e and https://github.com/numberscope/frontscope/pull/230/commits/4fe5d6c2065ca757cff1b71534e0facacad4164a.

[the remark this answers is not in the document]
liammulh: Looks like it went from 63 lines of code to 35! That's great. I did a bit of manual regression testing, and from what I can tell this is working as intended. Resolving.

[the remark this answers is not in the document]
liammulh: Wait, I want to take a look at that unused variable issue. Un-resolving...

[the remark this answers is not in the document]
liammulh: Ah, it looks like @katestange fixed the lint error in https://github.com/numberscope/frontscope/pull/230/commits/fa6bff641013846306dd508cc2ecb76f30e920d1.

src/visualizers/VisualizerGrid.ts
liammulh: It looks like this was done in https://github.com/numberscope/frontscope/pull/230/commits/71ddfcd4ad3e73adedfca36d09290fb0dc74011c.

liammulh: The code in https://github.com/numberscope/frontscope/commit/71ddfcd4ad3e73adedfca36d09290fb0dc74011c looks good. I am not sure what the best way to test it would be. I think I'd feel more comfortable if we had unit tests for the "property functions" like `isPrime`, `isSumOfTwoSquares`, etc.

src/visualizers/VisualizerGrid.ts
liammulh: I don't understand what the issue is, so I'd like to discuss this on Monday. We can write up an issue and resolve this thread if it's not an issue with this PR.

src/visualizers/VisualizerGrid.ts
liammulh: Looks like `Fill_Cell` and `Box_In_Cell` were added in https://github.com/numberscope/frontscope/pull/230/commits/bf0ceed0fdad013675d43d859d418b4b6db2ccbb and https://github.com/numberscope/frontscope/pull/230/commits/f23f1284c6ce2320bbc9e669032e5ca11f7ac73e, respectively. Resolving.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1220 @@
...
+                && factors[1][1] == 1n
+                && factors[2][1] == 1n) // negative of semi-prime
+        ) {
+            return true
+        } else {
+            return false
+        }
+    }
+}
+
```
katestange, line 1073: still a bug with a095102, which seems to miss the top right disc when using the default settings

src/visualizers/VisualizerGrid.ts
gwhitney: Maybe somehow caching and recompiling is slow on sequences by formula? Like there's some accidental quadratic term in sequences by formula? I don't know that's true. Just trying to brainstorm a counterintuitive result.

src/visualizers/VisualizerGrid.ts
gwhitney: Disc?

katestange: I commented on the wrong PR, oops, I'll move it to the right place.  Resolving.

src/visualizers/VisualizerGrid.ts
katestange: yeah, I did a lot of by-hand testing...

gwhitney:
The overlaid numbers when "show numbers" is checked are not centered in their cells. Should this be fixed before merge, or filed as an issue after merge?

gwhitney:
In the latest revision, properties that depend on factorizations do not work with Rows_Augment path, because the factorization being returned is that of the sequence value, but the number being displayed is the sequence value plus the row number. I'll work on this.

katestange:
> In the latest revision, properties that depend on factorizations do not work with Rows_Augment path, because the factorization being returned is that of the sequence value, but the number being displayed is the sequence value plus the row number. I'll work on this.

oh good catch... and not obvious how to fix, since for the augmented values we need to rely on naive factoring I guess.

katestange:
> The overlaid numbers when "show numbers" is checked are not centered in their cells. Should this be fixed before merge, or filed as an issue after merge?

I'm fine with it being for later.  For the sake of finishing up this PR.

gwhitney:
Correct, I am almost done with a fix; I fall back to simpleFactor when no factorization is available, and throw an error if simpleFactor fails (rather than show false data).

gwhitney:
OK, that should take care of the Rows_Augment problem. I am going to make the divisor, last digit, and number of polygonal sides a separate param for those properties to vastly reduce the number of defined properties. Should I also make Box_In_Cell the default for the second property so that if you define two properties they (by default) will not overwrite each other?

katestange:
> OK, that should take care of the Rows_Augment problem. I am going to make the divisor, last digit, and number of polygonal sides a separate param for those properties to vastly reduce the number of defined properties. Should I also make Box_In_Cell the default for the second property so that if you define two properties they (by default) will not overwrite each other?

That would be wonderful about the last-digit -- will that mean you can have different bases?  As for Box_In_Cell, yes, that sounds good.

gwhitney:
No, I am making the divisor a parameter for Divisible_By a parameter, the last digit (base 10) a parameter for Last_Digit_Is, and the number of sides a parameter for Polygonal Number.  So still the last digit properties will just be for base 10. It's too much of a pain to make there be a different _number_ of parameters for different values of the property (it was already an addition to allow a parameter to be different for different values of the property). OK will also change the default for property 2.

gwhitney:
Since the grid has to be square anyway, should we be specifying the side of the grid rather than its number of cells (which then entails rounding down to a square)?

gwhitney:
OK, as far as I can see, the only open points are the performance on sequences by formula, which is not really within the scope of this PR, and how to specify the size of the grid, which is probably a matter of taste anyway, and the centering of the numbers which we decided would be a subsequent issue. Also, I am finally content with the structure of the code.

Thus, I propose that @liammulh  (when available) do one final PR review from scratch (since he is probably the least recent to get into the nitty gritty of this) and if all seems well, to go ahead and merge this visualizer.

katestange:
> OK, as far as I can see, the only open points are the performance on sequences by formula, which is not really within the scope of this PR, and how to specify the size of the grid, which is probably a matter of taste anyway, and the centering of the numbers which we decided would be a subsequent issue. Also, I am finally content with the structure of the code.
> 
> Thus, I propose that @liammulh (when available) do one final PR review from scratch (since he is probably the least recent to get into the nitty gritty of this) and if all seems well, to go ahead and merge this visualizer.

I agree.  It would be great to wrap it up!  I'm on the fence about specifying the grid width vs. number of values.  In a way, number of values is easier because you might know how many terms you have access to.  So I would just leave it, I guess?

gwhitney:
Yeah, that's what I am proposing -- to leave this be feature-wise as is, have one of us do a final review as if from scratch, and merge when it's up to par.

liammulh:
> Yeah, that's what I am proposing -- to leave this be feature-wise as is, have one of us do a final review as if from scratch, and merge when it's up to par.

Starting my review now.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1040 @@
...
+            displayName: (d: Property) => propertyAuxName[Property[d]] || '',
+            required: false,
+            visibleDependency: `property${index}`,
+            visiblePredicate: (d: Property) => Property[d] in propertyAuxName,
+        },
+    }
+}
+
+/*
+ *   FUNCTIONS TO CHECK FOR PROPERTIES
```
liammulh, line 272: I will add a commit that strips the unnecessary leading white space from this comment and adds more information about the property functions.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1040 @@
...
+    switch (factors.length) {
+        case 1:
+            return factor > 1n && power === 1n
+        case 2:
+            return factor === -1n && factors[1][1] === 1n // negative of prime
+    }
+    return false // two or more prime factors
+}
+
+//Adapted from Geeks For Geeks :
```
liammulh, line 287: I will fix these comments so there is a space after the `//`.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1040 @@
...
+        }
+    }
+    if (legendre === 1) {
+        return true
+    } else {
+        return false
+    }
+}
+
+//Modification of Geeks for Geeks :
```
liammulh, line 328: I will fix these comments so there is a space after the `//`.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1040 @@
...
+    Abundant: isAbundant,
+    Perfect: isPerfect,
+    Deficient: isDeficient,
+    Semi_Prime: isSemiPrime,
+}
+
+class VisualizerGrid extends VisualizerDefault {
+    name = 'Grid'
+
+    //Grid variables
```
liammulh, line 421: I will fix these comments so there is a space after the `//`.

liammulh (commented):
Aside from the few comments I made, I am happy with the structure of the code as well. I will add some commits to fix comments.

I will also do another round of manual testing trying to be as thorough as possible. I will also test the docsite.

src/components/SeqVizParamsModal.vue
```diff
@@ -322,16 +322,39 @@
                     if (this.params[name].visibleDependency) {
                         const dependsOn =
                             this.params[this.params[name].visibleDependency]
-                        if (
-                            dependsOn.value !== this.params[name].visibleValue
-                        ) {
-                            continue
+                        if ('visiblePredicate' in this.params[name]) {
```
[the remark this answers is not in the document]
gwhitney: OK, let me know when you are done with the other commit(s) you are planning based on your review and I will work on a commit that will add an explanatory comment for each of the computed properties, and try to simplify the code in visibleParams so that it will be clearer and linebreak better.

src/visualizers/VisualizerGrid.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/230/commits/9cddb1cbd00ffae47823820f576cc37b78e76f91.

src/visualizers/VisualizerGrid.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/230/commits/9cddb1cbd00ffae47823820f576cc37b78e76f91.

liammulh:
In https://github.com/numberscope/frontscope/commit/9cddb1cbd00ffae47823820f576cc37b78e76f91, I accidentally altered a file that wasn't in the scope of this PR, so I reverted it in https://github.com/numberscope/frontscope/pull/230/commits/8ea3459d6bfb87a94eb13be9fec8c01053ef5910.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1041 @@
...
+        if (prime === -1n) {
+            // negative numbers are never sums of squares
+            return false
+        }
+        // otherwise update legendre for every prime 3 mod 4
+        if (prime % 4n === 3n && exponent % 2n === 1n) {
+            legendre = -1 * legendre
+        }
+    }
+    if (legendre === 1) {
```
liammulh, line 322: I will simplify this.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1041 @@
...
+            this.propertyObjects.push(ithPropertyObject)
+            Object.assign(
+                this.params,
+                getPropertyParams(i, ithPropertyObject)
+            )
+        }
+    }
+
+    checkParameters() {
+        const status = super.checkParameters()
```
liammulh, line 621: I think we can just return `super.checkParameters()`, which I will do.

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1041 @@
...
+    }
+
+    showNumber() {
+        const currentNumberAsString = this.currentNumber.toString()
+
+        if (this.showNumbers) {
+            this.sketch.fill(this.numberColor)
+            this.sketch.text(
+                currentNumberAsString,
+                this.x + 0 * this.scalingFactor,
```
liammulh, line 836: Why are we multiplying the scaling factor by 0? Why are we adding 0 to `this.x`?

liammulh (commented):
I found a few more things. Two I will fix, one is a question.

src/visualizers/VisualizerGrid.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/230/commits/ac7ac7363ad696a4c17ea965e5f52b43c836f09b.

src/visualizers/VisualizerGrid.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/230/commits/4e85af069d6f9c3c915403272a00a8cef809b2ff.

liammulh:
@gwhitney, I am done making commits, so when convenient please take a look at the commits I added, and add some explanatory notes for `visibleParams`.

src/visualizers/VisualizerGrid.ts
gwhitney: Agreed. Resolving.

src/visualizers/VisualizerGrid.ts
gwhitney: Also looks good

src/visualizers/VisualizerGrid.ts
```diff
@@ -0,0 +1,1035 @@
...
+            this.propertyObjects[i].aux = this.params[`prop${i}Aux`]
+                .value as bigint
+        }
+    }
+
+    setup(): void {
+        this.setPresets()
+        this.setOverridingSettings()
+
+        console.log('Background', this.backgroundColor)
```
liammulh, line 639: This needs to be removed.

gwhitney: will do in the commit I am working on. Thanks.

gwhitney:
OK, think I took care of the outstanding items in the commit I just made. Over to you =)

src/visualizers/VisualizerGrid.ts
liammulh: Looks good!

src/components/SeqVizParamsModal.vue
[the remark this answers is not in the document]
liammulh: @gwhitney added some comments in https://github.com/numberscope/frontscope/pull/230/commits/e4b27f474aee11070f3dbb3ef83677915682a385.

liammulh:
I did my final review of the code, did a manual test of the visualizer, and read through the documentation for the visualizer. I am okay with merging at this point.

gwhitney:
Why then please do!

Issue #120, opened 2022-05-22
visualizers enable mouseClicked(); visualizers access their canvas


katestange:
The visualizer should be able to use the [mouseClicked()](https://p5js.org/reference/#/p5.Element/mouseClicked) feature of p5.js.

Visualizers should be encouraged to use it as associated to a canvas, so the user has to click in the canvas (not just anywhere).  In order for this to happen, the visualizer has to have access to its own canvas, as in the examples at the link above.  See also #118.

liam-mulhall:
I believe this is part of @katestange's work on https://github.com/numberscope/frontscope/issues/115. (Or at least she made this issue while working on #115.)

liam-mulhall:
This blocks progress on https://github.com/numberscope/frontscope/issues/115.

katestange:
This may be helpful: [vue-p5 project by Kinrany](https://github.com/Kinrany/vue-p5)

gwhitney:
So the canvas you refer to is present in the p5 object as (in our case) `this.sketch._renderer` -- however, the privateness of this property as suggested in JavaScript by the name (beginning with underscore) is _enforced_ in TypeScript, so there is no way to access the property without generating a TypeScript error. Presumably the reason the property is private is because the p5 folks feel it may change in the future. So I don't think it's a good idea to circumvent TypeScript here to access it.

So there are a couple of ways I have thought of to proceed:

1) The drawingContext (presuming you are now able to access that as I was per my recent comment in #118) does contain a reference back to the canvas, which  you can access via `this.sketch.drawingContext.canvas`. But note that obtaining it in this way yields a "raw" `HTMLElement` that you must manipulate using the basic JavaScript interfaces (as documented for example at  https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement). It is lacking the fairly extensive convenience wrapper that p5 puts around such things. As a particular example, you cannot call `.mouseClicked(handler)` on this object, you have to call `.addEventListener('click', handler)`. By no means terrible, but  it doesn't exactly provide a simple/consistent programming environment for visualizer writers not coming from a web-techie background.

[...]

PR #231, opened 2022-12-20
feat: add NumberGlyph visualizer

liammulh:

<hr/>

This is @LeongHouHeng's sequence color visualizer PR. We are continuing it based on a branch in this repository. The original PR, which was submitted from a fork is https://github.com/numberscope/frontscope/pull/191.


src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+            value: this.n,
+            forceType: 'integer',
+            displayName: 'number of terms',
+            required: true,
+        },
+    }
+
+    private currentNum = 0
+    private X = 0
+    private Y = 0
```
katestange, line 40: When I submitted my Chaos Visualizer for a PR, @gwhitney encouraged me to use the p5.Vector class which has attributes x and y. See the chaos visualizer code [here](https://github.com/numberscope/frontscope/blob/bb157252d6ce8dfb603bbb0860531daee34a3053/src/visualizers/VisualizerChaos.ts#L236). That might be nice for some of your x and y coordinate variables.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+            // Check if drawing finished
+            if (this.currentNum >= this.n) {
+                this.firstDraw = false
+            }
+            this.sketch.noStroke()
+        } else {
+            // Monitor keyboard events after finishing drawing
+            // Bug: multiple clicks detected when only one click happened
+            this.keyboardEvents()
+        }
```
katestange, line 110: I'm finding the following behaviour: When I initially create a bundle, it draws the small version with coloured balls. Then when I click "Draw" and go into the main canvas, it is black/blank. The left key works to show/hide the legend. But no circles appear until I hit the right key for "refresh". So maybe it needs to have an automatic refresh somehow when moving from bundle card to main canvas and back? Would a change of this.firstDraw bool go into setup()? I'm not exactly sure.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+/** md
+# Sequence Color Visualizer
+In this visualizer, the brightness of each circle represents how fast $1/a^n$, 
+where a is a number in the sequence, 
+approaches zero as n increases.
+Each prime number is assigned a color, 
+and the color of each circle represents 
+the color combination of its prime factors. \s
```
katestange, line 14: Maybe you could explain that the first circle is associated to the first term of the sequence, and the second circle to the second number in the sequence, etc. You could also say that the fuzziness of the ball is a visual indicator: a brighter, less fuzzy ball means that term is bigger.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+                this.boxIsShow = true
+                this.drawBox()
+            }
+        }
+
+        // Show label when right arrow key is pressed
+        if (
+            this.sketch.keyIsDown(this.sketch.RIGHT_ARROW)
+            && this.firstDraw == false
+        ) {
```
katestange, line 129: What about negative terms? Right now it draws black. I suggest you just take the absolute values of terms, so that the sequences will converge. If you do this, you should also document this behaviour in the documentation above.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+            if (this.showLabel == false) {
+                this.showLabel = true
+                this.drawLabel()
```
katestange, line 132: The visualizer will hang pretty quickly on very large numbers (of which there are plenty in the OEIS). You might want to have a cut-off for a, where if you see a number bigger than that cutoff, you just "fake it" by drawing a pretty non-fuzzy bright circle (I'm assuming that as the numbers get very big, the eye won't detect the "fake" here; maybe I'm wrong?).

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+import type {VisualizerInterface} from '@/visualizers/VisualizerInterface'
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+/** md
+# Sequence Color Visualizer
+In this visualizer, the brightness of each circle represents how fast $1/a^n$, 
+where a is a number in the sequence, 
+approaches zero as n increases.
+Each prime number is assigned a color,
```
katestange, line 12: Another thought: since a lot of sequences (including just the natural numbers) get to a bright full ball fairly quickly (in other words, all balls look pretty bright), maybe we could somehow "tune down" the brightness (e.g. run the given algorithm on a/10 or on log(a)) until we can see more a visual difference between, say natural numbers and squares?

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    draw() {
+        if (this.firstDraw == true && this.currentNum < this.n) {
+            const currentElement = Number(
+                this.seq.getElement(this.currentNum++)
+            )
+            this.drawCircle(currentElement)
+
+            this.changePosition()
+
```
katestange, line 100: Another thing that might be nice is to be able to toggle the labels (the numbers 1/a^n) on and off with a key. That way, when we turn them off, we can have a nice pure visual. You could even have them toggle on/off along with the legend (left key). So when legend is on, numbers show. When legend is off, numbers are gone.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+        if (this.sketch.keyIsDown(this.sketch.LEFT_ARROW)) {
+            if (this.boxIsShow == true && this.subG != null) {
+                this.boxIsShow = false
+                this.undrawBox()
+            } else {
+                this.boxIsShow = true
+                this.drawBox()
+            }
+        }
+
```
katestange, line 124: I've noticed what I think is a bug. Try the visualizer with the sequence from formula n, n+1, n+2, n+3 etc. The circles should just be "shifting left" between each of these. But the first circle is not fuzzy, it looks like a single disk, e.g. n+2. So something funny is happening with the first circle. (I've added this comment near where the disc is calculated, but I don't know what the bug is, I may be in the wrong place in the code.)

katestange (changes requested):
This is transferring the review comments from the old branch.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
+import type p5 from 'p5'
+import {VisualizerDefault} from './VisualizerDefault'
+import type {VisualizerInterface} from '@/visualizers/VisualizerInterface'
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+/** md
+# Sequence Color Visualizer
```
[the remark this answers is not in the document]
gwhitney: Do the colors have anything to do with the sequence itself? Or are they simply indicating which divisor sequence a particular disk is using, i.e. are the same colors regardless of the sequence? If they are always the same, then maybe the colors are not really "part of" the visualization in some sense, i.e., the colors aren't telling you anything about the sequence you're analyzing, and then shouldn't be part of the name? So maybe just "Growth Charts"?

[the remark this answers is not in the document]
katestange: The colours are the prime factorizations of the terms.  I think basically the colour represents prime factorization and the brightness represents size in a way.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+            this.changePosition()
+
+            // Check if drawing finished
+            if (this.currentNum >= this.n) {
+                this.firstDraw = false
+            }
+            this.sketch.noStroke()
+        } else {
+            // Monitor keyboard events after finishing drawing
+            // Bug: multiple clicks detected when only one click happened
```
katestange, line 108: This bug needs resolving.  I gave it a go, however, it looks like it requires `keyPressed()` from p5.js, and this isn't working; maybe similar to #120 ?

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+        for (let i = 2; i <= Math.sqrt(n); i++) {
+            if (n % i === 0) {
+                return false
+            }
+        }
+        return n > 1
+    }
+
+    //return a number which represents the color
+    primeFactors(n: number, totalPrime: number) {
```
katestange, line 296: `isPrime()` and `primeFactors()` should use `sequence.getFactors()`

src/visualizers/SeqColor.ts
katestange: I can no longer reproduce this behaviour.

katestange: Actually, with A095102, for example, the refresh (right arrow) does alter the image.  So there still is some version of this bug going on.

src/visualizers/SeqColor.ts
katestange: done; check it out

src/visualizers/SeqColor.ts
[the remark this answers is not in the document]
katestange: changed; I went for "Primes and Sizes"

src/visualizers/SeqColor.ts
katestange: done; you can take a look (repeats code from PR #230 (isPrime) but this is temporary I suppose)

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,390 @@
+import type p5 from 'p5'
+import {VisualizerDefault} from './VisualizerDefault'
+import type {VisualizerInterface} from '@/visualizers/VisualizerInterface'
+import {VisualizerExportModule} from '@/visualizers/VisualizerInterface'
+import type {SequenceInterface} from '../sequences/SequenceInterface'
+
+/** md
+
+# Primes and Sizes Visualizer
```
katestange, line 9: The visualizer needs to be fully documented.

src/visualizers/SeqColor.ts
katestange: This is actually already implemented, so this comment is no longer relevant.  Resolving.

src/visualizers/SeqColor.ts
katestange: In a recent commit, I made it so the user can decide the growth function by formula, which should alleviate this (and besides, now it would be very difficult to do this).  So I would like to simply resolve this conversation -- no action suggested.  Sound ok?

src/visualizers/SeqColor.ts
katestange: This is effectively resolved by the recent commit where I added a formula box so you can control the growth by formula.  So I think this is resolved.  If you agree, please resolve.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,439 @@
...
+            this.canvasSize.x / this.n + 30,
+            this.canvasSize.x / this.n + 50
+        )
+
+        // Obtain all prime numbers from the sequence
+        for (let i = this.seq.first; i < this.n; i++) {
+            const checkCurrentPrime = this.seq.getElement(i)
+            if (
+                this.isPrime(i)
+                && !this.primeNum.includes(checkCurrentPrime)
```
katestange, line 149: this is odd, because it takes the prime values of the sequence, not prime divisors?  is that what we want?  those are the only primes which contribute to the colouring of the sequence, as the code stands.

src/visualizers/SeqColor.ts
katestange: This took a lot to track down!  It was because we needed to call `this.sketch.noStroke()` -- it was drawing the circle but with a black border that overwrote the colour of the last, larger circle each time.  So it was there, but very hard to see.  Somewhere else `noStroke()` gets set before it draws the next circle.  This is fixed in the most recent commit.  Please check it out.

src/visualizers/SeqColor.ts
katestange: fixed in a recent commit (except not yet documented).  now it takes the absolute value of any growth function

src/visualizers/SeqColor.ts
katestange: This seems to be fixed with the other bug fix.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,452 @@
...
+            this.changePosition()
+        }
+    }
+
+    // Remove all labels by drawing circles again
+    undrawLabel() {
+        this.redrawCircles()
+    }
+
+    redrawCircles() {
```
katestange, line 285: When we toggle the labels and the legend-box, sometimes the redraw doesn't clear the background, so we end up with circles over top of the white label box.

src/visualizers/SeqColor.ts
katestange: fixed; take a look

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,454 @@
...
+            this.initialPosition.y
+        )
+        this.firstDraw = true
+        this.currentIndex = this.seq.first
+        this.sketch.redraw()
+    }
+
+    drawCircle(ind: number) {
+        const numberNowBigint = this.seq.getElement(ind)
+        const numberNow = Number(numberNowBigint)
```
katestange, line 299: For discussion: should we try to make this visualizer work with bigint?  It allows for arbitrary formulas in terms of the input values based on @gwhitney 's code, so I don't know how feasible that is.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,566 @@
...
+First, the non-negative integers, with 
+the default settings. Next, with growth rings with 
+modulus 25. Finally, with growth rings with 
+modulus 1.  The final example shows some interesting 
+effects because the rings
+actually occur more rapidly than once per pixel.  The first
+two discs appear to be missing because they have no prime factors
+to color them.
+
+######
```
katestange, line 558: Still to do on documentation:  add more examples.  Note that once the grid vis is in, the latex here will work also.

src/visualizers/SeqColor.ts
katestange: This should be resolved with the new docs.

src/visualizers/SeqColor.ts
katestange: mostly done now; still needs maybe another interesting example or two

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,566 @@
...
+const colorMap = new Map()
+
+class SeqColor extends VisualizerDefault implements VisualizerInterface {
+    name = 'Primes and Sizes'
+    n = 64
+    ringsBool = false
+    modulus = 25
+    formula = 'log(n^x)'
+
+    params = {
```
katestange, line 43: still a bug with a095102, which seems to miss the top right disc when using the default settings

katestange: fixed now

src/visualizers/SeqColor.ts
katestange: an example of semiprimes added, it's probably enough for now.

liammulh:
Oh no! I just realized that when I transferred the old comments a couple weeks ago I didn't click the "Finish your review" button, and I guess maybe that made it so that other people couldn't see that I had transferred the comments from the old PR over? It looks like @katestange also transferred the comments. I'll try to delete any comments that are duplicates.

liammulh:
Okay, I think I got rid of all the duplicate comments.

gwhitney:
This PR uses a non-conventional name for the visualizer: SeqColor.ts, rather than VisualizerPrimesSizes.ts. Either it the implementation of this visualizer should have its filename renamed in the "standard" way, or we should rename all of the others (as in just Chaos.ts, Turtle.ts, Differences.ts, etc.). There is definitely an argument for the latter, as having "Visualizer" in the name of every source file in the "src/visualizer" directory is kind of redundant. We should discuss tomorrow which way to go and make it so.

gwhitney:
The image files for the documentation have not yet been included in any commit in this PR. Could whoever has them please commit and push them? Thanks.

gwhitney:
Checklist of items that need to be taken care of in this PR, from group discussion today. If you are the one who verifies that one of these things has been done by a new commit to this PR, please check the item off. Note that this is not necessarily exhaustive, there may well still be additional items that need resolution in the conversations above.

- [ ] Rename to NumberGlyph visualizer
- [ ] make the parameter dialog initially have a single 'customize glyph generation' boolean parameter and the other other parameters are only shown when that is checked
- [ ] Commit the images needed in the documentation
- [ ] Rebase on PRs solving #120 and/or #241 when they are merged
- [ ] Change the default glyph-generating function to something that looks better, including defaulting growth rings to true
- [ ] Make code check for safety of conversion to number (note that sequence `2^n%17` was not coming out periodic, probably because of this bug, although it may not be something that can be handled in this way because the problem there is _inside_ mathjs; the inputs and outputs are OK numbers, but they will eventually be wrong because of a internal overflow in mathjs).
- [ ] Warn in docs about needing to fit into number (for now)
- [ ] Deal with integers with no prime factors so that they are not just blank black spaces (maybe use grey?)
- [ ] Once #120 is fixed, improve situation so that toggling the legend doesn't "flicker"

katestange:
My most recent commit attempts to handle possible bigint errors but it doesn't do a great job... suggestions appreciated.

src/visualizers/SeqColor.ts
katestange: we decided to document its failings; see [comment here](https://github.com/numberscope/frontscope/pull/231#issuecomment-1377967936)

src/visualizers/SeqColor.ts
katestange: I'm quite happy with this now; ready for review

src/visualizers/SeqColor.ts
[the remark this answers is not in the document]
katestange: See  [comment here](https://github.com/numberscope/frontscope/pull/231#issuecomment-1377967936) -- it was renamed to glyphs.

katestange:
I believe everything is dealt with here except for the issue that is blocked by issue #120 .  Please take a look and resolve the open conversations or give feedback.

katestange (approved):

gwhitney:
I took the liberty of fitting the square of glyphs to the canvas, and I found a way in this case of avoiding the `{} as p5.Vector` TypeScript hack. So just went ahead and committed those minor changes.

src/visualizers/SeqColor.ts
gwhitney: Moot now with keyboard controls removed.

src/visualizers/SeqColor.ts
gwhitney: Vectors seem mostly appropriately used; resolving.

src/visualizers/SeqColor.ts
gwhitney: The current docs seem to describe this OK, even if it is still just slightly tricky to follow exactly what's going on in glyph generation.

src/visualizers/SeqColor.ts
gwhitney: Doesn't seem to work; if I make a growth function that returns all negative values (like multiplying the default one by -1), I get a completely black canvas. Please could you do the abs-ing claimed here and document this aspect? If you don't get to it before I get back from my morning bike ride, I will do it when I get back, no problem.

src/visualizers/SeqColor.ts
gwhitney: OK, resolving.

src/visualizers/SeqColor.ts
gwhitney: Moot since key interaction has been removed. Resolving.

src/visualizers/SeqColor.ts
gwhitney: Seems OK now.

src/visualizers/SeqColor.ts
gwhitney: Overall documentation is good.

src/visualizers/SeqColor.ts
gwhitney: Again, moot because keyboard interaction has been removed. Resolving.

src/visualizers/SeqColor.ts
gwhitney: Agreed.

src/visualizers/SeqColor.ts
```diff
@@ -0,0 +1,344 @@
...
+        }
+    }
+
+    keyboardEvents() {
+        // Show description box when the left arrow key is pressed
+        if (this.sketch.keyIsDown(this.sketch.LEFT_ARROW)) {
+            if (this.boxIsShow == true && this.subG != null) {
+                this.boxIsShow = false
+                this.undrawBox()
+            } else {
```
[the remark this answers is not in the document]
gwhitney: Moot now that keyboard interaction has been removed.

gwhitney:
OK, all seems well except for negative growth function values, and I have to check a couple errors I left unresolved above but don't have time this second (they  are probably fine though). Once that's all done I will merge.

src/visualizers/SeqColor.ts
gwhitney: Agreed.

src/visualizers/SeqColor.ts
gwhitney: OK, going from n+2 to n+3 doesn't just shift left, it also changes colors because the primes are colored in the order they occur in the sequence, not in numerical order. But I think that is intentional, and I think it is working. Resolving.

src/visualizers/SeqColor.ts
gwhitney: Yes, seems to take the primes that occur as factors, not just those that occur as elements of the sequence.

src/visualizers/SeqColor.ts
gwhitney: OK, at this point A095102 is producing a nice rainbowy sequence of glyphs and all 64 appear with the default settings, so closing.

gwhitney:
OK, have tested all of the other open points and resolved them, so I will now take care of the negative growth function values.

gwhitney:
Alright, now just the absolute value of the brightness function is used (and I dropped the outer `abs` from the default brightness function). I don't see anything else, so merging.

PR #245, opened 2023-01-11
feat: add E2E test for differences visualizer

liammulh:

<hr/>

This PR adds the ability to open the Cypress GUI and run an E2E test for the differences visualizer. The E2E test works by programmatically making a visualizer, taking a screenshot of the visualizer, and comparing the screenshot to an expected screenshot. This PR contains documentation on how to add further E2E tests for visualizers.

We will still need to integrate the E2E tests into our CI. I think that is outside the scope of this PR. I just want to ensure the test is working properly, the code seems reasonable, and the docs are up to snuff. Then we can integrate into CI.


cypress/afterScreenshot.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * The hook that runs after screenshots are taken. It verifies that
+ * screenshots are close, and it optionally cleans up the screenshots
+ * directory.
+ */
+const afterScreenshot = (options: {
+    actualPath: string // The path to the "actual" screenshot.
+    expectedPath: string // The path to the "expected" screenshot.
+    shouldCleanup: boolean // Whether we should clean up the screenshots dir.
+    screenshotClosenessPercent: number // How close screenshots should be.
```
liammulh, line 73: I think the comment here is misleading. It probably needs to be "Percent difference allowed between screenshots." Or something along those lines.

gwhitney:
OK, I took an initial look and I am confused. I pulled this branch, did `npm install`, and then `npm run test:e2e:open`. To my surprise, a window popped up. So I clicked on "E2E testing" and I got another screen with a mustard-colored box saying "Warning: Cannot Connect Base Url Warning". Forging ahead, I clicked Firefox and then "Start E2E Testing in Firefox". And that opened a new window in my previously running instance of Firefox. And in that window, everything seemed to have failed, with a lot of messages like "connect ECONNREFUSED 127.0.0.1:5173".

So that leads me to three questions/requests:

1. I was not expecting to have to interact with a _test_. Can this all be automated for me and just respond "pass" or "fail" (with diagnostic output in the latter case, ideally)?

2. If it's not going to work that way, then please add documentation also for _running_ tests in addition to creating them, since I am clearly doing something wrong.

3. If it's not going to work that way, then what will we do for continuous integration? We need something automated on the git actions that verifies that you can actually generate a visualization... that doesn't mean we should merge this current thing, but there will still be a need for something else then, I guess.

I will continue with the review when we get these things cleared up. Thanks!

liammulh:
> To my surprise, a window popped up.

In the description of the PR, I wrote: "This PR adds the ability to open the Cypress GUI and run an E2E test for the differences visualizer."

> So I clicked on "E2E testing" and I got another screen with a mustard-colored box saying "Warning: Cannot Connect Base Url Warning".

I believe that's because it needs something to test, i.e. you need to be running frontscope locally using `npm run dev` prior to `npm run test:e2e:open`.

> Forging ahead, I clicked Firefox and then "Start E2E Testing in Firefox". And that opened a new window in my previously running instance of Firefox. And in that window, everything seemed to have failed, with a lot of messages like "connect ECONNREFUSED 127.0.0.1:5173".

I think this is the same issue. You need to `npm run dev` prior to `npm run test:e2e:open`.

> I was not expecting to have to interact with a test. Can this all be automated for me and just respond "pass" or "fail" (with diagnostic output in the latter case, ideally)?

Yep, if we wanted to run it without the GUI, the command we would invoke would be `cypress run`. I was thinking that for this PR, we'd just do the bare-minimum to get Cypress into the code base, and then in subsequent PRs, we could add the "headless" tests where we run Cypress without the GUI. There are some advantages to the GUI, which I think is why the Cypress docs introduce you to it first. For instance, the GUI alerted me to the fact that there are canvases being drawn on the page where we make bundle cards.

> If it's not going to work that way, then please add documentation also for running tests in addition to creating them, since I am clearly doing something wrong.

Do you think adding info the the help section in `package.json` would suffice? Alternatively, we could make the script `npm run dev && cypress open`, but then the name of the script is somewhat misleading. I'm open to suggestions.

> If it's not going to work that way, then what will we do for continuous integration? We need something automated on the git actions that verifies that you can actually generate a visualization... that doesn't mean we shouldn't merge this current thing, but there will still be a need for something else then, I guess.

No, I think it can work that way.

cypress/afterScreenshot.ts
```diff
@@ -0,0 +1,84 @@
...
+import {PNG} from 'pngjs'
+import {readFileSync, readdirSync, rmSync} from 'node:fs'
+import {resolve} from 'node:path'
+
+const SCREENSHOT_DIR = resolve(process.cwd(), 'cypress', 'screenshots')
+
+/**
+ * Verifies that two screenshots are close.
+ */
+const screenshotsAreClose = (options: {
```
liammulh, line 20: Rereading this, I think `actualPath` and `expectedPath` should be passed explicitly rather than passed in the options object.

cypress/afterScreenshot.ts
```diff
@@ -0,0 +1,84 @@
...
+        rmSync(resolve(SCREENSHOT_DIR, screenshot), {force: true})
+    }
+}
+
+/**
+ * The hook that runs after screenshots are taken. It verifies that
+ * screenshots are close, and it optionally cleans up the screenshots
+ * directory.
+ */
+const afterScreenshot = (options: {
```
liammulh, line 69: Again, I think `actualPath` and `expectedPath` should be passed explicitly rather than passed in the options object.

gwhitney:
Thanks for all of the clarification. My two cents is that we don't necessarily need a package manager script just for firing up an interactive test environment. We can just put how to do that in the appropriate place in the documentation. What we need the script for is running the automated tests to see if all is well. In other words, the interactive tests would likely be used when someone is doing more exploratory, development-type work (such as investigating the canvases) and for that, one would want to be more up to speed on the particular tools we are using (and hence it's reasonable to expect the person to consult the documentation). But for just checking that a PR hasn't harmed anything, a person just wants a yes or no answer from an easy-to-invoke test, for which a package manager script is perfect.

So my recommendation is: change the `test:e2e:open` script to `test:e2e` that automatically/headlessly runs all of the cypress tests that have been defined, and add docs in the appropriate place about running Cypress interactively, which would include mentioning that you first have to run the `dev` script. Thoughts? We can of course hash this out at the meeting tomorrow.

gwhitney:
I have no objection to any of your self-review comments, so if you want to add a commit doing those things and then resolve those conversations, that's fine by me.

liammulh:
> So my recommendation is: change the `test:e2e:open` script to `test:e2e` that automatically/headlessly runs all of the cypress tests that have been defined, and add docs in the appropriate place about running Cypress interactively, which would include mentioning that you first have to run the dev script. Thoughts?

That seems reasonable. I will:

- [ ] Implement @gwhitney's above suggestion.
- [ ] Add docs for running the Cypress GUI.
- [ ] Fix the percent difference comment: https://github.com/numberscope/frontscope/pull/245#discussion_r1067428818.
- [ ] Pass `actualPath` and `expectedPath` explicitly: https://github.com/numberscope/frontscope/pull/245#discussion_r1084355036.

cypress/afterScreenshot.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/245/commits/da9377bf7cac1f377b45ba8b2cafb8e1a077c5a7.

cypress/afterScreenshot.ts
liammulh: Done in https://github.com/numberscope/frontscope/pull/245/commits/da9377bf7cac1f377b45ba8b2cafb8e1a077c5a7.

liammulh:
Okay, @gwhitney, I just finished everything I outlined in https://github.com/numberscope/frontscope/pull/245#issuecomment-1400741136. The `test:e2e` script runs the dev server concurrently (see https://github.com/numberscope/frontscope/pull/245#issuecomment-1400864285), so you might see a warning that Cypress will retry connecting to the dev server. It just depends on how quickly the dev server starts up. I think possibly seeing a warning is a reasonable trade-off for having Windows support.

liammulh:
I added help comments about the Cypress warning I mentioned in https://github.com/numberscope/frontscope/pull/245#issuecomment-1400888863 to `package.json` in https://github.com/numberscope/frontscope/pull/245/commits/b2ede0b279cee16c96aa86161d75ad321bb118e2.

liammulh:
We need to look into whether `cypress run` runs the tests in a browser, and if so we need to determine which browsers we want to run the tests in.

gwhitney:
OK, well the output says it is running in Electron. If that's configurable, I think we want to change it to Firefox or Chromium (or run twice, once with each). However, it may be that for the headless testing, it has to be Electron, in which case I am fine with it.

However, for me the test fails. Output is pasted verbatim below. Can you investigate? If you're not able to figure out what's going on from the information there, let me know and I will try to look into it here. Thanks.

```
[1]   (Run Starting)
[1] 
[1]   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
[1]   │ Cypress:        12.3.0                                                                         │
[1]   │ Browser:        Electron 106 (headless)                                                        │
[1]   │ Node Version:   v19.1.0 (/usr/local/bin/node)                                                  │
[1]   │ Specs:          1 found (differences-visualizer.spec.cy.ts)                                    │
[1]   │ Searched:       cypress/e2e/**/*.cy.{js,jsx,ts,tsx}                                            │
[1]   └────────────────────────────────────────────────────────────────────────────────────────────────┘
[1] 
[1] 
[1] ────────────────────────────────────────────────────────────────────────────────────────────────────
[1]                                                                                                     
[1]   Running:  differences-visualizer.spec.cy.ts                                               (1 of 1)
[1] 
[1] 
[1]   differences visualizer
[1]     1) can produce a visualization
[1] 
[1] 
[1]   0 passing (2s)
[1]   1 failing
[1] 
[1]   1) differences visualizer
[1]        can produce a visualization:
[1]      Error: Image sizes do not match.
[1]       at pixelmatch (/home/glen/code/frontscope/node_modules/pixelmatch/index.js:21:15)
[1]       at screenshotsAreClose (/home/glen/code/frontscope/cypress/afterScreenshot.ts:36:27)
[1]       at afterScreenshot (/home/glen/code/frontscope/cypress/afterScreenshot.ts:81:10)
[1]       at Object.handler (/home/glen/code/frontscope/cypress.config.ts:34:36)
[1]       at invoke (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/child/run_plugins.js:43:18)
[1]       at <unknown> (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/util.js:59:14)
[1]       at tryCatcher (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/bluebird/js/release/util.js:16:23)
[1]       at Function.Promise.attempt.Promise.try (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/bluebird/js/release/method.js:39:29)
[1]       at Object.wrapChildPromise (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/util.js:58:23)
[1]       at RunPlugins.execute (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/child/run_plugins.js:180:21)
[1]       at EventEmitter.<anonymous> (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/child/run_plugins.js:279:12)
[1]       at EventEmitter.emit (node:events:513:28)
[1]       at EventEmitter.emit (node:domain:489:12)
[1]       at process.<anonymous> (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@packages/server/lib/plugins/util.js:33:22)
[1]       at process.emit (node:events:513:28)
[1]       at process.emit (node:domain:489:12)
[1]       at process.emit.sharedData.processEmitHook.installedValue [as emit] (/home/glen/.cache/Cypress/12.3.0/Cypress/resources/app/node_modules/@cspotcode/source-map-support/source-map-support.js:745:40)
[1]       at emit (node:internal/child_process:944:14)
[1]       at processTicksAndRejections (node:internal/process/task_queues:83:21)
[1] 
[1] 
[1] 
[1] 
[1]   (Results)
[1] 
[1]   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
[1]   │ Tests:        1                                                                                │
[1]   │ Passing:      0                                                                                │
[1]   │ Failing:      1                                                                                │
[1]   │ Pending:      0                                                                                │
[1]   │ Skipped:      0                                                                                │
[1]   │ Screenshots:  0                                                                                │
[1]   │ Video:        false                                                                            │
[1]   │ Duration:     2 seconds                                                                        │
[1]   │ Spec Ran:     differences-visualizer.spec.cy.ts                                                │
[1]   └────────────────────────────────────────────────────────────────────────────────────────────────┘
[1] 
[1] 
[1] ================================================================================
[1] 
[1]   (Run Finished)
[1] 
[1] 
[1]        Spec                                              Tests  Passing  Failing  Pending  Skipped  
[1]   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
[1]   │ ✖  differences-visualizer.spec.cy.ts        00:02        1        -        1        -        - │
[1]   └────────────────────────────────────────────────────────────────────────────────────────────────┘
[1]     ✖  1 of 1 failed (100%)                     00:02        1        -        1        -        -  
[1] 
[1] cypress run exited with code 1
--> Sending SIGTERM to other processes..
[0] npm run dev exited with code SIGTERM
```

liammulh:
> OK, well the output says it is running in Electron. If that's configurable, I think we want to change it to Firefox or Chromium (or run twice, once with each). However, it may be that for the headless testing, it has to be Electron, in which case I am fine with it.

I can configure this.

> However, for me the test fails. Output is pasted verbatim below. Can you investigate? If you're not able to figure out what's going on from the information there, let me know and I will try to look into it here. Thanks.

The error you're seeing seems to come from `frontscope/node_modules/pixelmatch/index.js`:

```js
function pixelmatch(img1, img2, output, width, height, options) {

    if (!isPixelData(img1) || !isPixelData(img2) || (output && !isPixelData(output)))
        throw new Error('Image data: Uint8Array, Uint8ClampedArray or Buffer expected.');

    if (img1.length !== img2.length || (output && output.length !== img1.length))
        throw new Error('Image sizes do not match.');
```

In `frontscope/cypress/afterScreenshot.ts`:

```ts
const screenshotsAreClose = (
    actualPath: string,
    expectedPath: string,
    options: {
        screenshotClosenessPercent: number
        pixelmatchThreshold: number
    }
) => {
    const actualContents = readFileSync(resolve(process.cwd(), actualPath))
    const expectedContents = readFileSync(
        resolve(process.cwd(), expectedPath)
    )
    const actualScreenshot = PNG.sync.read(actualContents)
    const expectedScreenshot = PNG.sync.read(expectedContents)
    const {width, height} = expectedScreenshot
    const diff = new PNG({width, height})
    const numDiffPixels = pixelmatch(
        actualScreenshot.data,
        expectedScreenshot.data,
        diff.data,
        width,
        height,
        {threshold: options.pixelmatchThreshold}
    )
```

Somehow `actualScreenshot.length` and `expectedScreenshot.length` don't match or `actualScreenshot.length` and `diff.length` don't match. (Or both pairings don't match.) Maybe some logging and/or debug statements are in order?

If we can't have consistent behavior across computers, do we abandon this approach, or do we create an `expectedScreenshot` that will work for the CI server but not for our personal computers?

liammulh:
> I think we want to change it to Firefox or Chromium (or run twice, once with each). However, it may be that for the headless testing, it has to be Electron, in which case I am fine with it.

Cypress docs:

> The "browser" argument can be set to `chrome`, `chromium`, `edge`, `electron`, `firefox` to launch a browser detected on your system. Cypress will attempt to automatically find the installed browser for you.

I think we should pick a subset of the browsers to run the tests on. I think a good argument can be made for Chromium and Firefox. Since Chrome, Edge, and Electron browsers are all Chromium-based, presumably the tests would also work on those browsers.

gwhitney:
> Maybe some logging and/or debug statements are in order?

Sure, and/or can we just explicitly specified the desired screen size that the browser should suppose it is running on, especially given it is actually running headless?

liammulh:
@gwhitney and I discussed this today, and we think we might be able to specify the viewport for the screenshot by doing something like

```cy.viewport(preset, orientation, options)```

with a preset like `macbook-16`.

liammulh:
In https://github.com/numberscope/frontscope/pull/245/commits/61eebe005a68429200eecb5890d8f1a8f0e5f886, I set the viewport before taking the screenshot, and I added a new screenshot that was taken with the viewport setting code in place. @gwhitney, when you get a chance, would you please try to `nom run test:e2e` and see if the test is still failing for you.

I tried to use Firefox instead of Electron, and my test failed. I wonder if this is because the expected screenshot needs to be taken with the same browser. I'm not sure why using a different browser would change the PNG.

gwhitney:
Just reiterating here that no, unfortunately the PNGs did not match (even in size) on my machine.

gwhitney:
Status: suggested path forward is to proxy the p5 class to record all requests to display and then compare that log between a known correct log and the test run.

PR #261, opened 2023-04-05
fix: Reduce error alerts by increasing caching tolerance

### LINKED ISSUE #260: Make sequences/Cached lazier with regard to factoring
Once https://github.com/numberscope/backscope/pull/67 is merged, it seems that a remaining issue that causes "still caching" errors in the front end is the time it takes to factor large sequence values, even when the factorizations will not be used. Therefore, the caching of factorizations should be decoupled from the caching of values, and not executed until there is actually a request for a factorization.

gwhitney:

<hr/>

Before this PR, all cached sequences loaded factorizations into the cache in strict lockstep with values. But many visualizers do not use the factorizations, and they require extra computations and/or http requests that can be wasteful if the results will not be used. This PR decouples the two caches, so that values can be cached without factorizations (although not vice versa, since you may need the values to obtain their factorizations).

This created a new issue: since the factorizations are not cached until their first request, every single OEIS sequence was reporting caching errors in the ShowFactors visualizer. That was because the ShowFactors simply loops over every value and gets its factorization. The first call to getFactors was triggering the cache fill, but for OEIS sequences the cache fill is _asynchronous_, so it returns immediately without having completed the operation. That meant the second call to getFactors was _guaranteed_ to trigger an error that caching was in progress, since it completely certain was in fact in progress at that point.

The solution to this conundrum was to take advantage of p5's display loop. In the draw method, such caching errors are caught, and trigger a "progress spinner" cursor to appear. But of course p5 then simply gets to the next frame (after a delay dependent on the frame rate) and tries to access the cached values again. After a (typically very small) number of frames, the cache fill has completed, and the sequence is able to supply the factorizations without any CachingErrors. At this point, the draw() method clears the cursor and all seems well.

Running this PR in the frontend with https://github.com/numberscope/backscope/pull/67 in a local backend is producing really robust visualizer results for brand-new OEIS sequences for me. Hopefully the same will prove true in the review.

Resolves #260

### DIFF
--- a/src/components/BundleCard.vue
+++ b/src/components/BundleCard.vue
@@ -27,6 +27,7 @@
 <script lang="ts">
     import {defineComponent} from 'vue'
     import p5 from 'p5'
+    import {CachingError} from '../sequences/Cached'
     // we need a unique id for each canvas
     // see https://github.com/vuejs/vue/issues/5886#issuecomment-308647738
     let cid_count = 0
@@ -44,7 +45,17 @@
                     viz.setup()
                 }
                 sketch.draw = function () {
-                    viz.draw()
+                    try {
+                        viz.draw()
+                    } catch (e) {
+                        if (e instanceof CachingError) {
+                            sketch.cursor('progress')
+                            return
+                        } else {
+                            throw e
+                        }
+                    }
+                    sketch.cursor(sketch.ARROW)
                     if (sketch.frameCount >= 50) {
                         sketch.noLoop()
                     }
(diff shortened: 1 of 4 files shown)


katestange:
Just a note that this appears to work perfectly on my machine, when paired with backend pr [67.](https://github.com/numberscope/backscope/pull/67)

katestange:
I read through the code and it seems clear to me, for the most part.  No particular questions.  Should I merge?

PR #269, opened 2023-09-27
Grid modify sequence capability

OliviaBrobin:

<hr/>

This pull request adds a modify sequence capability that allows the user to change the base sequence to the natural numbers. The user can select what number the natural numbers start at. There are also the properties sequence_number and not_sequence_number which allows the user to highlight numbers that either are or are not in the original sequence. This only works for increasing numbers so that the code can be efficient.

This is framed as modifying the sequence because I have other ideas related to how to modify the sequence.

\<Please provide a high-level description of your PR.\>

### DIFF
--- a/src/visualizers/Grid.ts
+++ b/src/visualizers/Grid.ts
@@ -89,6 +89,11 @@ enum PathType {
     Rows_Augment,
 }
 
+enum ModifySequence {
+    No,
+    Change_To_Natural_Numbers,
+}
+
 enum Direction {
     Right,
     Left,
@@ -107,6 +112,9 @@ const leftTurn: Record<string, Direction> = {
 
 enum Property {
     None,
+    Equals,
+    Sequence_Number,
+    Not_Sequence_Number,
     Prime,
     Negative,
     Even,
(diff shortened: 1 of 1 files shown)


gwhitney:
Thanks for the PR! Since of course you can run the GridVisualizer on the natural number sequence, it seems to me that the only new capacity of this PR is to add a property of whether the elements of one sequence occur in another. It would therefore be much preferable to do a prototype, perhaps initially limited to this visualizer before being allowed in other visualizers, of having a visualizer that uses two sequences, rather than just one. If you recall our past discussions, we really want sequence modifications to stay on the Sequence side of things, so any visualizer could use them, and have visualizers focus on, well, visualizing.

Looking forward to the other concepts you have in mind!

katestange:
In preparation for the new UI work starting next week, this PR is being closed.  It interacts with plans to create the capability for two sequences to be used simultaneously, which is in the works.  Once the frontend has that capability, it will make sense for this new capability of the grid visualizer to be built around that.  So this can be reopened and reworked once we have that in place.  See #276.

PR #297, opened 2024-04-22
fix: Tolerate visualizer inhabit()ing a new div before removal from old

### LINKED ISSUE #295: Order of components mounting/unmounting is nondeterministic.
MEA CULPA -- we don't have an explicit test that checks the result of npm run build, because we have no end-to-end tests, just the guideline in the review checklist that one should always try to run the result of the build. With the flurry of checkins lately, I neglected to always do that (`npm run build; npm run preview`). And now we have gotten into a state in which frontscope works fine in dev mode, but the built version interacts fine until you hit draw, and then there is a blank white area where the visualization should be, for any visualizer. Ugh ugh. I will try to debug. If someone wants to bisect to see where this happened, it could be helpful. I will be that it was the version bump to vue though. Not sure how we would back that up since that necessitated almost everything bumping along with it.

gwhitney:
Refactors the VisualizerInterface to require matching inhabit() and
  depart() calls, both taking an HTMLElement where its view will reside.
  A second inhabit() before any depart() automatically calls depart()
  on the prior view.

  Resolves #295.

### DIFF
--- a/src/components/BundleCard.vue
+++ b/src/components/BundleCard.vue
@@ -42,8 +42,8 @@
             this.viz.show()
             setTimeout(() => this.viz.stop(), 500)
         },
-        unmounted() {
-            this.viz.dispose()
+        beforeUnmount() {
+            this.viz.depart(document.getElementById(this.cid) as HTMLElement)
         },
         methods: {},
         props: {
--- a/src/components/CanvasArea.vue
+++ b/src/components/CanvasArea.vue
@@ -46,8 +46,10 @@
             )
             this.activeViz.show()
         },
-        unmounted() {
-            this.activeViz.dispose()
+        beforeUnmount() {
+            this.activeViz.depart(
+                document.getElementById('visualizer-here') as HTMLElement
+            )
         },
     })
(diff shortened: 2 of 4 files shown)


gwhitney:
@katestange @Vectornaut I think reviewing/merging this is absolutely top priority because the current Numberscope in main is uninstallable.

katestange:
I have done `npm run dev` and `npm run build` `npm run preview` and stress tested it by hand without causing any weird behaviour.

katestange (approved):
Tested and looked over, looks fine to me, and more importantly seems to work fine in dev and production.

gwhitney:
I personally am fine with merging. The only point under discussion in the issue is whether it is an error or a no-op to depart() a visualizer that is not currently inhabiting anything. That is something we could always change in a later PR, and the current error behavior is "conservative" in that if it ever happens we will find out about it and can respond then.

gwhitney:
So if it makes sense to you that this point doesn't have to be resolved before merging, please merge but leave the issue open. Thanks!

PR #303, opened 2024-04-22
doc: Update contributors and current leadership of project

gwhitney:

<hr/>

Just trying to get the about page up-to-date with new folks coming in.

### DIFF
--- a/doc/about.md
+++ b/doc/about.md
@@ -28,6 +28,7 @@ You can find [Numberscope on GitHub](https://github.com/numberscope) or
 -   Aaron Fenyes
 -   Brendan Heaney
 -   Steven Hristopoulos
+-   Clyde Kertzer
 -   Jennifer Leong
 -   Ang Li
 -   Theodore Lincke
--- a/doc/about.md
+++ b/doc/about.md
@@ -12,9 +12,9 @@ source projects.
 You can find [Numberscope on GitHub](https://github.com/numberscope) or
 [contact the organizers](mailto:numberscope@colorado.edu).
 
-## Current Technical Lead
+## Current Design Committee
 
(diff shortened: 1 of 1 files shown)


katestange (changes requested):
Please make it Katherine E. Stange (more official?) and add Clyde Kertzer?

gwhitney:
OK, as you are mentioned twice, I left the first one as Dr. Katherine E. Stange, and changed the second to just Katherine Stange. Does that seem OK? Also, I looked and as far as I can see, no code from Clyde is actually merged yet. So the time to add him would be when something is merged (and there is actually an item in the PR checklist for doing that, if we were just taking the time to go through the checklist every time instead of being in such a blazing rush).

katestange:
> I looked and as far as I can see, no code from Clyde is actually merged yet. So the time to add him would be when something is merged (and there is actually an item in the PR checklist for doing that, if we were just taking the time to go through the checklist every time instead of being in such a blazing rush).

The list partially pre-dates the move to github and includes all the students that worked on the project in any capacity, even if they don't have code in the current version right now.  Unless we want to sort that all out retroactively and (re)move some names, I think it's fair to include Clyde, because he is in the same boat at the moment.  These other students who haven't checked in code have contributed in various ways including brainstorming, forming ideas, using numberscope, participating in discussions, creating prototypes etc. etc.  So I want to honour that.  If "contributors" means actual code, then we may need two lists, "code contributors" and "numberscope project contributors" or something.

gwhitney:
All good. Definitely no reason to be stingy with credit. I just didn't realize the membership of this list.

katestange (approved):
Looks good now!

PR #293, opened 2024-04-22
Add a template for p5 visualizers

Vectornaut:
By submitting this PR, I am indicating to the Numberscope maintainers that I have read and understood the [contributing guidelines](https://numberscope.colorado.edu/doc/CONTRIBUTING/) and that this PR follows those guidelines to the best of my knowledge. I have also read the [pull request checklist](https://numberscope.colorado.edu/doc/doc/pull-request-checklist/) and followed the instructions therein.

<hr/>

This update adds a new visualizer directory, `src/workbench`, whose contents don't normally show up in the visualizer list. To use the visualizers in this directory, you serve Frontscope in "workbench mode" by calling `npm run dev:workbench`.

If this pull request looks generally okay, I'll go on to document the workbench feature.

Right now, there's only one workbench visualizer: "p5 Visualizer Template", also added in this update.

PR #310, opened 2024-04-23
refactor: Improve uniformity of VisualizerInterface and P5Visualizer

gwhitney:
Three related interface changes for the Visualizer hierarchy, as
   discussed in #293. These are all designed to make some of the
   details of defining a Visualizer less confusing and more concise.

   (A) In P5Visualizer, draw() should be abstract and hence required
       on runnable derived Visualizers, since any (additional) boilerplate
       needed by all P5Visualizers would go in the function that wraps
       this method in the default inhabit() method. Thus implementation
       draw() functions no longer need to call super.draw().
   (B) Since setup() is the default p5 initialization method and can only
       be called once per sketch anyway, clarify that generally initialization
       should be carried out in setup() rather than inhabit(), in derived
       Visualizers.
   (C) Since `name` is used as an instance property in SequenceInterface
       indicating the details of the particular interface, re-designate
       that information as the return value of a `visualization()` method
       of an entity implementing the VisualizationInterface, and suggest
       that it should only depend on the class of the entity, not the
       instance. Implement this in the P5Visualizer hierarchy with
       the visualizationName static property, and modify the
       VisualizerExportModule constructor to look for it there, so that
       the visualizationName will not have to be stated twice in every
       module implementing a Visualizer. At such time as there are other
       base classes for concrete Visualizers, these conventions should
       probably be moved one level up to a common base class of all
       visualizers; there is nothing specifically related to p5 about this
       naming scheme.

### DIFF
--- a/src/components/BundleCard.vue
+++ b/src/components/BundleCard.vue
@@ -2,9 +2,12 @@
     <div class="col-4-sm">
         <div class="card" style="width: 18rem">
             <div class="card-body">
-                <h5 class="card-title">{{ seq.name + ' + ' + viz.name }}</h5>
+                <h5 class="card-title">
+                    {{ `${viz.visualization()} of ${seq.name}` }}
+                </h5>
                 <p class="card-text">
-                    This is a bundle of {{ seq.name + ' + ' + viz.name }}.
+                    Display {{ seq.name }} using a {{ viz.visualization() }}
+                    visualization.
                 </p>
                 <div class="card-preview" :id="cid"></div>
                 <div class="card-buttons">
--- a/src/visualizers/Grid.ts
+++ b/src/visualizers/Grid.ts
@@ -416,7 +416,7 @@ const propertyIndicatorFunction: {
 }
 
 class Grid extends P5Visualizer {
-    name = 'Grid'
+    static visualizationName = 'Grid'
 
     // Grid variables
     amountOfNumbers = 4096
@@ -660,7 +660,6 @@ earlier ones that use the _same_ style.)
     }
 
     draw(): void {
-        super.draw()
         this.currentIndex = Math.max(this.startingIndex, this.seq.first)
         let augmentForRowReset = 0n
 
@@ -887,9 +886,8 @@ earlier ones that use the _same_ style.)
 }
 
 export const exportModule = new VisualizerExportModule(
-    'Grid',
     Grid,
-    'Puts numbers in a grid.'
+    'Puts numbers in a grid, highlight cells based on various properties.'
 )
 
 /** md
--- a/src/visualizers/Grid.ts
+++ b/src/visualizers/Grid.ts
@@ -887,7 +887,7 @@ earlier ones that use the _same_ style.)
 
 export const exportModule = new VisualizerExportModule(
     Grid,
-    'Puts numbers in a grid, highlight cells based on various properties.'
+    'Puts numbers in a grid, highlighting cells based on various properties'
 )
 
 /** md
--- a/src/visualizers/ModFill.ts
+++ b/src/visualizers/ModFill.ts
@@ -80,4 +80,7 @@ class ModFill extends P5Visualizer {
     }
 }
 
-export const exportModule = new VisualizerExportModule(ModFill)
+export const exportModule = new VisualizerExportModule(
+    ModFill,
+    'A triangular grid showing which residues occur, to each modulus.'
+)
--- a/src/visualizers/ModFill.ts
+++ b/src/visualizers/ModFill.ts
@@ -18,8 +18,8 @@ occur by watching the order the cells are filled in as the diagram is drawn.
 ## Parameters
 **/
 
-class VizModFill extends P5Visualizer {
-    name = 'Mod Fill'
+class ModFill extends P5Visualizer {
+    static visualizationName = 'Mod Fill'
     modDimension = 10n
     params = {
         /** md
@@ -72,7 +72,6 @@ class VizModFill extends P5Visualizer {
     }
 
     draw() {
-        super.draw()
         this.drawNew(this.sketch, this.i, this.seq)
         this.i++
         if (this.i == 1000 || this.i > this.seq.last) {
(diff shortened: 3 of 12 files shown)


gwhitney:
(A) Oops, I accidentally pushed to origin rather than my fork when creating this PR. I will try to be more careful.

(B) There were likely some doc-page changes that should go with this, but my belief is that Aaron will merge or rebase this into #293 where there are overlapping doc changes and straighten it all out. @Vectornaut if that's not right, let me know, and I can add a doc commit to this PR as well.

src/visualizers/Grid.ts
```diff
@@ -397,8 +397,8 @@ const propertyIndicatorFunction: {
     [key in PropertyName]: key extends FactorPropertyName
         ? (factorization: Factorization) => boolean
         : key extends ValuePropertyName
-        ? (() => boolean) | ((value: bigint, aux?: bigint) => boolean)
-        : never
+          ? (() => boolean) | ((value: bigint, aux?: bigint) => boolean)
+          : never
```
katestange, line 401: why spacing?

src/visualizers/ModFill.ts
```diff
@@ -81,8 +80,4 @@ class VizModFill extends P5Visualizer {
     }
 }
 
-export const exportModule = new VisualizerExportModule(
-    'Mod Fill',
-    VizModFill,
-    ''
-)
+export const exportModule = new VisualizerExportModule(ModFill)
```
katestange, line 83: the other vis's have a description at the export line.  For this one a reasonable desc would be 'A dynamic histogram of modular values.' (If I'm recalling the concept correctly.)

src/visualizers/NumberGlyph.ts
```diff
@@ -403,11 +402,7 @@ The default value is 25.
     }
 }
 
-export const exportModule = new VisualizerExportModule(
-    'Number Glyphs',
-    NumberGlyph,
-    ''
-)
+export const exportModule = new VisualizerExportModule(NumberGlyph)
```
katestange, line 405: Again, the others have a desc, maybe something like 'Colorful glyphs representing properties of the first few terms'

src/visualizers/Turtle.ts
```diff
@@ -167,4 +161,4 @@ class Turtle extends P5Visualizer {
     }
 }
 
-export const exportModule = new VisualizerExportModule('Turtle', Turtle, '')
+export const exportModule = new VisualizerExportModule(Turtle)
```
katestange, line 164: maybe 'Terms of the sequence control a colourful walk on the screen.' or 'Turtle walks according to terms in the sequence, creating a trail.'

katestange (changes requested):
Works in production and dev mode.

src/visualizers/Grid.ts
gwhitney: Actually, this reflects a change in how our automated tools reformat Grid.ts that apparently resulted from the giant dependency update we have been going through. I am not quite sure how it filtered through to be a diff in this PR, but now that I have rebased, it is gone as a difference.

src/visualizers/ModFill.ts
gwhitney: OK. The description is optional, and is unused in the current app, but it could be like a tooltip or something in a revamped UI. The fact there is not one here just reflects that there was not one before for ModFill.  But I totally get how this might be good juncture to fill all the rest in. I will add a commit doing that. Do you also want me to make the description mandatory?

gwhitney:
OK, all visualizers have a description. Should we enforce this going forward, or leave it optional as it has been to date?

katestange:
> OK, all visualizers have a description. Should we enforce this going forward, or leave it optional as it has been to date?

I don't know... Yes, I'm imagining this might be a tooltip text or populate some type of menu or explorer or some type.  I leave it to you to decide if it's useful to make mandatory here.  It could be something that depends on the UI revamp.

src/visualizers/ShiftCompare.ts
```diff
@@ -117,7 +116,6 @@ class VizShiftCompare extends P5Visualizer {
 }
 
 export const exportModule = new VisualizerExportModule(
-    'Shift Compare',
-    VizShiftCompare,
-    ''
+    ShiftCompare,
+    'A grid showing pairwise congruence of sequence entries, to some modulus'
```
katestange, line 120: whether or not they have a period should probably be consistent.  right now some do, some don't

katestange (approved):
tiny comment

gwhitney:
OK, I will remove any periods from descriptions that have them since I think they are mostly not full sentences. And I will make descriptions mandatory, since we now have them all and they seem like they might be handy in a UI revamp, if we ever have a chance to do one ;-)

katestange:
looks good!

gwhitney:
OK, Aaron, ready for you to push the merge button whenever it works for you.

Vectornaut:
> Implement this in the P5Visualizer hierarchy with
the visualizationName static property, and modify the
VisualizerExportModule constructor to look for it there, so that
the visualizationName will not have to be stated twice in every
module implementing a Visualizer.

I like this simplification!

src/components/BundleCard.vue
Vectornaut, line 1: The new messages are much clearer to me.

src/components/BundleCard.vue
```diff
@@ -2,9 +2,12 @@
...
         <div class="card" style="width: 18rem">
             <div class="card-body">
-                <h5 class="card-title">{{ seq.name + ' + ' + viz.name }}</h5>
+                <h5 class="card-title">
+                    {{ `${viz.visualization()} of ${seq.name}` }}
+                </h5>
                 <p class="card-text">
-                    This is a bundle of {{ seq.name + ' + ' + viz.name }}.
+                    Display {{ seq.name }} using a {{ viz.visualization() }}
+                    visualization.
```
Vectornaut, line 10: These new messages are much clearer to me.

Vectornaut (approved):

PR #359, opened 2024-06-26
Sequence visualizer switcher

kaldis-berzins:

<hr/>
This is merge number 6 in the PR merge chain.

# Sequence visualizer switcher
## Switchers
 - There are two separate files `ChangeSequenceModal.vue` and `ChangeVisualizerModal.vue`. We tried to combine these into one file, but that made for far less readable code, as the internals of each modal aren't shared. The optimal solution that you might want to explore here is to make a `Modal.vue` component with a slot, thus drastically reducing the amount of duplicate code, but keeping everything quite readable.
 - OEIS doesn't work at this point, that gets done in `final_fixes`
 - Changing to all other sequences/visualizers should work, provided they work themselves.
## Parameter Editor
 - It is updated to work with switching visualizers and specimens.
 - More specifically, when a paramable is assigned to (like in the case of switching a visualizer/specimen), the watch function gets called that makes sure to reset all the statuses of the parameters. This is so that no errors are carried over from the previous visualizer/sequence.
## Specimen bar
 - The buttons in the bar are refactored to be moved into a button container and the save button popup has been redone to fix some visual bugs
 ## Mobile
 -  Breakpoint changed to 850px, as the specimen bar occupies a lot of horizontal space on the screen.
## Scope
 - Some additional tab bugs are fixed here, where the dropzones sometimes don't have the correct status.
 - Mobile is fully working at this point.


gwhitney:
@kaldis-berzins @maxd0328 and colleagues, if any of you happen to still be paying attention to this repo: we don't understand in this PR why in src/App.vue, the `src` CSS property of the Infinity font needs to/should change but that of Inter does not. Any insight would be helpful. Thanks.

maxd0328:
I believe this may be an artifact of some of the edits we made and this change can probably be removed. If not, I'll ask @kaldis-berzins, since he was the one working with the fonts.

kaldis-berzins:
This is a change made by Gido, I actually don't think it's needed. If it works without the change you can leave it out.

gwhitney:
Once again, thanks so much for the help.

gwhitney:
Closing as superseded by #379

PR #360, opened 2024-06-27
Final fixes

maxd0328:

<hr/>

This is merge number 7 in the PR merge chain.

# Final fixes

## OEIS support
 - OEIS sequences no longer have any live parameters, so they can pull from Backscope only once.
 - The user can import an OEIS sequence with an OEIS ID, alias, cache block, and optional modulus in the sequence switcher.
 - Importing and OEIS sequence in the sequence switcher automatically applies it to the current specimen.
 - When a saved visualizer relying on an OEIS sequence is loaded, the OEIS sequence is automatically pulled from Backscope if it has not been already.
## Download button
 - The download button now functions, producing an image from the visualizer and downloading it for the user.
## Bug fixes
 - Clicking the Numberscope logo in the specimen editor now resets the visualizer.
 - Miscellaneous mobile support issues have been fixed.
 - Footer has been removed from the specimen editor and is now only present in the gallery.
 - Fixed crash when calculating erroneous elements in Formula.

NOTE: There is a bug that has not been fixed to a shortage of time, which has to do with the synchronicity of OEIS sequences. Populating OEIS sequences from Backscope is of course done asynchronously, but the rest of the application appears not to be aware of this, as even in old Numberscope there were no callbacks or awaits (so this bug existed previously as well). This would require a considerable refactor to make specimen loading asynchronous. Currently, when importing an OEIS sequence, there is a chance that the OEIS sequence may not be populated by the time the visualizer attempts to render. This causes the visualizer to hang, and requires the user to press the reset button in the specimen bar to get it to properly render.


gwhitney:
All of the content of this PR, with adjustments by the core group, is now present in the ui2 branch, except as described in #418. Hence closing.

PR #355, opened 2024-06-26
Specimen browser storage

kaldis-berzins:
By submitting this PR, I am indicating to the Numberscope maintainers that I have read and understood the [contributing guidelines](https://numberscope.colorado.edu/doc/CONTRIBUTING/) and that this PR follows those guidelines to the best of my knowledge. I have also read the [pull request checklist](https://numberscope.colorado.edu/doc/doc/pull-request-checklist/) and followed the instructions therein.

<hr/>
This PR is second in the PR merge chain.

# Specimen browser storage
This PR adds two new files that aren't used anywhere at this point: `browserCaching.spec.ts` and `browserCaching.ts`.  `browserCaching.ts` adds the interface for storing specimens in the browser. This file is updated in later iterations, but largely remains the same.

PR #362, opened 2024-06-28
Specimen browser storage [cleaned]

gwhitney:

<hr/>

This PR is second in the PR merge chain.
Specimen browser storage

This PR adds two new files that aren't used anywhere at this point: browserCaching.spec.ts and browserCaching.ts. browserCaching.ts adds the interface for storing specimens in the browser. This file is updated in later iterations, but largely remains the same.

[Maintainer's note: we would not usually allow a PR that simply adds unused files, but are making an exception in this case for the sake of the chain of related PRs resulting from the Delft project.]

Supersedes #355.

### DIFF
--- a/src/shared/__tests__/browserCaching.spec.ts
+++ b/src/shared/__tests__/browserCaching.spec.ts
@@ -1,0 +1,109 @@
+import {describe, it, expect, vi, beforeEach} from 'vitest'
+import {
+    getCurrent,
+    updateCurrent,
+    saveSpecimen,
+    deleteSpecimen,
+    openSpecimen,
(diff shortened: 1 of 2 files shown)


katestange (commented):
I've done what I can here to test manually.  For example, the URL is persistent even when turning off and reloading the front end, and I tried various combinations of parameters and visualizers.  Unit tests run fine.  I'll keep at it a bit (because it is fun), but I think you should assume functionality is fine.

gwhitney:
OK, this seems to be a reasonably-written, self-contained new facility with reasonably thorough unit tests. Merging.

PR #356, opened 2024-06-26
Resizing and url integration

kaldis-berzins:
By submitting this PR, I am indicating to the Numberscope maintainers that I have read and understood the [contributing guidelines](https://numberscope.colorado.edu/doc/CONTRIBUTING/) and that this PR follows those guidelines to the best of my knowledge. I have also read the [pull request checklist](https://numberscope.colorado.edu/doc/doc/pull-request-checklist/) and followed the instructions therein.

<hr/>
This branch is third in the PR merge chain.

# Resizing and URL integration
## Changes to tabs
 - Added buttons that let you control the tabs without dragging
 - Rewrote large sections of the tab resizing code. This was in order to support minimizing.
 - Added a highlight to the tab that is currently selected.
 - Initial mobile compatibility in the style section of `Tab.vue`
   - This is changed in later PRs, but the majority of this css remains.
## Resizing
 - Added a reset function to a specimen. This is a hard reset that is used when the specimen is resized and it doesn't override resized function.
 - Added a resized function to both specimens and visualizers. It is not required for visualizers as they can just not do anything and the specimen will hard reset them.
## Scope
 - Adds rudimentary mobile support. This doesn't work entirely, and gets changed substantially in later PRs, so reviewing here probably isn't worth your time.

PR #366, opened 2024-06-28
Resizing and url integration [cleaned]

gwhitney:

<hr/>

      Changes to tabs
        * Added buttons that let you control the tabs without dragging
        * Support minimizing of tabs.
        * Added a highlight to the tab that is currently selected.
        * Initial mobile compatibility in the style section of Tab.vue
    
      Resizing
        * Added a reset function to a specimen. This is a hard reset that is used
          when the specimen is resized and it doesn't override resized function.
        * Added a resized function to both specimens and visualizers. It is not
          required for visualizers as they can just not do anything and the
          specimen will hard reset them.
    
      Scope
        * Adds rudimentary mobile support.

Supersedes #356.

### DIFF
--- a/src/components/Tab.vue
+++ b/src/components/Tab.vue
@@ -10,7 +10,7 @@
                 <span
                     class="docking material-symbols-sharp"
                     @click="dockWindow">
-                    dock_to_right
+                    dock_to_side
                 </span>
             </div>
         </div>
--- a/src/components/Tab.vue
+++ b/src/components/Tab.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="tab" @click="selected" last-coords-x="0" last-coords-y="0">
+    <div class="tab" @click="selected">
         <div class="drag">
             <div class="buttons">
                 <span
@@ -103,13 +103,15 @@
 
                 const tab = event.target.parentElement
                 if (!(tab instanceof HTMLElement)) return
-                if (tab.getAttribute('docked') === 'none') return
+                const zoneName = tab.getAttribute('docked')
+                if (!zoneName || zoneName === 'none') return
 
                 const dropzone = document.querySelector(
-                    '#' + tab.getAttribute('docked') + '-dropzone'
+                    `#${zoneName}-dropzone`
                 )
                 if (!(dropzone instanceof HTMLElement)) return
 
+                tab.setAttribute('last-dropzone', zoneName)
                 positionAndSizeTab(tab, dropzone)
             },
         },
@@ -219,6 +221,13 @@
             tab.classList.add('minimized')
         }
     }
+    // helper to choose docking site
+    function zoneOrder(preferred: string) {
+        const zoneParts = preferred.split('-')
+        const otherVert = zoneParts[0] === 'top' ? 'bottom' : 'top'
+        return [preferred, `${otherVert}-${zoneParts[1]}`]
+    }
+
     function dockWindow(event: MouseEvent) {
         const tab = (event.currentTarget as HTMLElement).closest('.tab')
         if (!(tab instanceof HTMLElement)) return
@@ -273,21 +282,16 @@
         tab.setAttribute('last-coords-x', x.toString())
         tab.setAttribute('last-coords-y', y.toString())
 
-        const firstDropzone = document.querySelector('#top-right-dropzone')
-        const secondDropzone = document.querySelector(
-            '#bottom-right-dropzone'
-        )
-
-        if (!(firstDropzone instanceof HTMLElement)) return
-        if (!(secondDropzone instanceof HTMLElement)) return
-
-        // if top right dropzone is empty, dock there
-        if (firstDropzone.classList.contains('empty')) {
-            positionAndSizeTab(tab, firstDropzone)
-        }
-        // otherwise dock in the bottom right dropzone
-        else {
-            positionAndSizeTab(tab, secondDropzone)
+        const preferredZone = tab.getAttribute('last-dropzone') || 'top-right'
+        const zonesToTry = zoneOrder(preferredZone)
+        for (const zoneName of zonesToTry) {
+            const zone = document.querySelector(`#${zoneName}-dropzone`)
+            if (!(zone instanceof HTMLElement)) continue
+            if (zone.classList.contains('empty')) {
+                positionAndSizeTab(tab, zone)
+                tab.setAttribute('last-dropzone', zoneName)
+                break
+            }
         }
     }
     // select the tab when it is clicked
@@ -333,7 +337,7 @@
         }
         .tab {
             border: 1px solid var(--ns-color-black);
-            width: 300px;
+            width: var(--ns-desktop-tab-width);
             height: 200px;
             z-index: 50;
         }
--- a/src/components/Tab.vue
+++ b/src/components/Tab.vue
@@ -10,7 +10,7 @@
                 <span
                     class="docking material-symbols-sharp"
                     @click="dockWindow">
-                    dock_to_side
+                    dock_to_right
                 </span>
             </div>
         </div>
@@ -28,6 +28,8 @@
         positionAndSizeAllTabs,
         selectTab,
     } from '../views/Scope.vue'
+
+    const minimizedTabHeight = 110
 
     // every element with draggable class can be dragged
     interact('.tab').resizable({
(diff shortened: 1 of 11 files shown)


gwhitney:
I will review when next I can work on this but @katestange and/or @Vectornaut feel free to jump in in the meantime if you happen to have a chance.

src/components/Tab.vue
```diff
@@ -1,6 +1,19 @@
 <template>
-    <div class="tab">
-        <div class="drag"></div>
+    <div class="tab" @click="selected" last-coords-x="0" last-coords-y="0">
```
katestange, line 2: both tabs (visualizer and sequence) have the default "last undocked position" set to 0,0.  This means if you open up a fresh scope and immediately use the tab buttons to undock both tabs, they cover each other up in the upper left, which could be confusing.  It would probably be better to set one of them to default lower down.

src/visualizers/VisualizerInterface.ts
```diff
@@ -65,7 +70,20 @@ export interface VisualizerInterface<PD extends GenericParamDescription>
...
     depart(element: HTMLElement): void
-
+    /**
+     * This is called when the size of the visualizer should change, either
+     * because the window is resized, or the docking configuration has changed.
+     * Visualizer writers should take care to resize their canvas and to make
+     * sure that any html elements aren't wider than the requested width.
+     * The new size is the full available size cut down to fit in the requested
+     * aspect ratio.
+     *
```
katestange, line 80: I'm not sure where in the code to point this issue, but it has to do with resizing.  To cause it, resize the browser window on a fresh scope to something mobile-like (or use mobile view in the developer tools) in Chromium, and you get the first attached screenshot.  Then reloading while in that size configuration will produce the second attached screenshot.  Neither of them is good behaviour.  Not sure if we should fix this now or log it as a bug for later, since it's a pretty corner case for usage, unless this PR is meant to be working fully with mobile already.  Note that I think the footer is going away later in the PR chain, so this may resolve itself.

Screenshot from 2024-06-28 17-12-02
Screenshot from 2024-06-28 17-12-20

katestange (commented):
Aside from the comments mentioned, it seems very robust on desktop view.  The behaviour is pretty intuitive.

src/components/Tab.vue
gwhitney: OK, if @kaldis-berzins or colleagues don't have a chance to add a commit to this PR handling this point, I will take a crack at it when next I have the opportunity.

gwhitney:
> Aside from the comments mentioned, it seems very robust on desktop view. The behaviour is pretty intuitive.

Did you get a chance to try it with a variety of visualizers (by changing "ModFill" in Scope.vue to the names of other visualizers)?

katestange:
> Did you get a chance to try it with a variety of visualizers (by changing "ModFill" in Scope.vue to the names of other visualizers)?

Yes, although not every single one.  I'll try a couple more.

src/components/Tab.vue
katestange: Another corner case behaviour related to this.  If you open a fresh scope and then move both tabs from docked-right to docked-left, they also keep their "last docked" variable set to upper left (0,0) and are still registered as docked.  So then when you click the "dock/undock" button they don't return to the right side (as they would do if you moved both tabs to the middle).  They just go "up" on top of each other.  So given this, I actually think the best solution is for the `last-coords-x/y` variables to be defaulting to something in the middle of the screen, not top and bottom upper left as my original comment may have suggested.

gwhitney: Maybe sort of one tab width to the right of the left edge? With Sequence at the top edge and Visualizer near the bottom?

gwhitney: Here's another oddity that doesn't feel right: If you load the default visualizer per the code of this PR ("Random visualized by ModFill"), and then click the minimize button on the Visualizer tab, and then immediately click the "unminimize" button on Visualizer, it does not go back to the same size that it was -- after this operation Sequence is taller and Visualizer is shorter than at the beginning. Is that sufficiently poor behavior that we would like to address it at the level of this PR?

katestange: > Maybe sort of one tab width to the right of the left edge? With Sequence at the top edge and Visualizer near the bottom?

Maybe slightly more than one tab width, and slightly down from the top edge.  This failure to align/tile would make the visual impression that we are "floating" more immediate.  Also, we could have the second one float similarly, but slight further down and right, so they are overlapping a little.  That creates a strong visual impression of floating boxes.

gwhitney: Another observation: You cannot simultaneously minimize both Sequence and Visualizer. when docked. When one is minimized, the other is maximized to take the whole dock area. While ultimately that also does not feel "right" to me and I think we should revisit that behavior at some point, at the level of this PR I think it is the intended design choice and we should not seek to do anything about it in this current review. If anyone disagrees, please comment.

katestange: The problem with my suggestion being, of course, that the more we move them off the top left, the more it becomes a potential problem on a small screen.

gwhitney: > The problem with my suggestion being, of course, that the more we move them off the top left, the more it becomes a potential problem on a small screen.

How about Sequence a half tab width down and to the right from the top left, and Visualizer a full tab width down and to the left?

katestange: > Here's another oddity that doesn't feel right: If you load the default visualizer per the code of this PR ("Random visualized by ModFill"), and then click the minimize button on the Visualizer tab, and then immediately click the "unminimize" button on Visualizer, it does not go back to the same size that it was -- after this operation Sequence is taller and Visualizer is shorter than at the beginning. Is that sufficiently poor behavior that we would like to address it at the level of this PR?

I'd leave this in the category of "not important enough for slowing the PR chain".  Unless it's super easy/obvious to fix (which may be the case), in which case the slowing is negligible.  If we don't fix it, we should log an issue, but I don't think it is likely to cause much consternation compared to many other outstanding issues.

gwhitney: > > The problem with my suggestion being, of course, that the more we move them off the top left, the more it becomes a potential problem on a small screen.
> 
> How about Sequence a half tab width down and to the right from the top left, and Visualizer a full tab width down and to the right?

Here's a screenshot of this proposal (for "undocking" of both tabs immediately upon startup).
Screenshot_2024-06-28_17-28-00

katestange: > > The problem with my suggestion being, of course, that the more we move them off the top left, the more it becomes a potential problem on a small screen.
> 
> How about Sequence a half tab width down and to the right from the top left, and Visualizer a full tab width down and to the right?

If you're trying it out, I trust your instinct.  That sounds reasonable.

katestange: > Another observation: You cannot simultaneously minimize both Sequence and Visualizer. when docked. When one is minimized, the other is maximized to take the whole dock area. While ultimately that also does not feel "right" to me and I think we should revisit that behavior at some point, at the level of this PR I think it is the intended design choice and we should not seek to do anything about it in this current review. If anyone disagrees, please comment.

I agree.

katestange: > > > The problem with my suggestion being, of course, that the more we move them off the top left, the more it becomes a potential problem on a small screen.
> > 
> > 
> > How about Sequence a half tab width down and to the right from the top left, and Visualizer a full tab width down and to the right?
> 
> Here's a screenshot of this proposal (for "undocking" of both tabs immediately upon startup). !

Oh that looks nice.  Just to clarify, I think you mean it still starts  up docked, but if you immediately undock, you get this.

katestange: I assume the first one you undock is at back in the overlap?

gwhitney: Correct on both counts. I basically know how to do this now, but I still don't know how to make it that the coordinates are tied with the tab width so that they are intentionally 1/2 and one full tab width, respectively, as opposed to by coincidence being those amounts. So I need to figure out how to get the values of constants into the CSS _and_ into the component templates.

However, in trying this out I discovered another disturbing behavior: if you doc a tab to the left, and then undock it and immediately re-dock it, it docks to the right. That does not seem right. As with the others, is it sufficiently "un-right" that we should address it in this PR? It's the most "bug-like" of these docking behaviors that I have noticed so far.

katestange: About left/right docking.  I noticed that behaviour but I assumed it was intentional.  I suppose however, it would be better to dock to wherever you were docked last.  Pondering it a bit more, I think it would be good to fix this.  It's a bit finnicky, though... You can actually dock one on the left and one on the right, so it's a per-tab variable that would be needed.

gwhitney: That's not a problem, the last undocked coordinates are already per tab so no reason why the last dock position can't be.

gwhitney: OK, I have pushed a commit that implements those initial positions and also returns tabs to the last place they were docked when you hit the dock button. Please try it out and if it all looks good, please merge this PR! (I am happy with it at this point.)

katestange:
Works great, merging.

PR #398, opened 2024-07-24
feat: Allow adding and dropping of OEIS sequence cards

gwhitney:

<hr/>

This PR represents another cleaned portion of the seventh Delft PR (#360). It allows OEIS sequence cards to be added and deleted from the list that is shown in the sequence switcher. It stores this list of "active" OEIS IDs in browser localStorage so that it is persistent across page reloads, tab closures, and the like. This PR still has a completely non-functional OEIS Search bar; instead, it allows adding new OEIS sequences as follows:

Visit a URL for a specimen that uses an OEIS sequence (we should add at least one to the Gallery as well -- any proposals, @katestange ?) Then, go into your browser's URL bar and move your cursor to the end of the URL. You will see the OEIS ID there as part of the URL. Just type your new desired OEIS ID in place of the one that is there, and then visit the resulting URL. With this PR, that new URL will be correctly interpreted, creating a Specimen that uses that sequence, even though the system has not before encountered that OEIS ID. Moreover, the next time you open the Sequence Switcher panel, you will see that new OEIS Sequence card inserted at the beginning of the block of OEIS Sequences.

You can delete OEIS Sequence cards from the OEIS List by clicking on the garbage can icon when the Sequence Switcher is open. Deletion of non-OEIS sequences is not allowed (no icon is shown on those cards). The Gallery and Saved Specimens should work as before (although they deserve testing as the changes in this PR affected their code).

Once this PR is reviewed, improved as needed, and merged, the final planned piece of this Delft PR -- the OEIS search bar -- will be implemented in a following PR.

### DIFF
--- a/src/components/SpecimenCard.vue
+++ b/src/components/SpecimenCard.vue
@@ -14,7 +14,9 @@
                 v-if="!permanent"
                 v-on:click.stop="deleteSpecimen"
                 style="padding-right: 15px">
-                <span class="material-icons-sharp" style="user-select: none">
+                <span
+                    class="material-icons-sharp"
+                    style="user-select: none; margin-top: 1ex">
                     delete
                 </span>
             </div>
@@ -85,7 +87,7 @@
         align-items: left;
         text-align: left;
         display: flex;
-        align-items: center;
+        align-items: start;
         justify-content: space-between;
     }
     .card-title {
--- a/src/components/SpecimensGallery.vue
+++ b/src/components/SpecimensGallery.vue
@@ -7,7 +7,7 @@
             :subtitle="specimen.subtitle"
             :lastEdited="specimen.lastEdited"
             @specimenDeleted="removeSpecimen"
-            :permanent="!canDelete" />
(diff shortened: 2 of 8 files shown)


katestange:
Here are a couple pretty or interesting ones that could be gallery material:

http://localhost:5173/?name=OEIS+A004001&viz=Turtle&domain=0+1+&range=15+180+&stepSize=10&start=0+200&strokeWeight=1&bgColor=e0def7&strokeColor=5e8d85&seq=OEIS+A010060&modulus=2

http://localhost:5173/?name=Number+of+divisors&viz=Chaos&corners=8&walkers=8&first=1&last=100000&dummyDotControl=true&alpha=0.7&pixelsPerFrame=2000&seq=OEIS+A000005

http://localhost:5173/?name=Chaos&viz=Chaos&walkers=4&first=1&last=100000&dummyDotControl=true&pixelsPerFrame=1000&darkMode=true&seq=OEIS+A008475

katestange (approved):
I played with it a lot, no suggestions or changes!

PR #402, opened 2024-07-25
Enhancements to Turtle

katestange:

<hr/>

This PR provides enhancements to the turtle visualizer:

1. Instead of a "domain" input there's just a modulus and the sequence is automatically taken with respect to that modulus (just like chaos does), so you shouldn't have so many boring thumbnails that go two steps and die.
2. The number of elements in the "angles" list has to match the modulus; instead of causing an error, the parameter dialog will shorten or lengthen it automatically for you when you change modulus.  Similar for other lists.
3. The step size is now a list of step sizes, like the angles.
4. There's a path length option.
5. The default behaviour is to draw the full path from scratch in one frame, no animation.
6. If you input a growth rate in the Path Growth variable, it will go back to drawing the path at that speed, so you can watch the animation.
7. If you use the "Turning angle increments" list, the angles in the angle list will gradually change, causing the path to "fold" gradually over itself like a protein.  This is fun for finding other variations on a cool picture you've found.  This is also an animation feature.  You can combine it with growth, or use just one of them.

Note:  This turtle visualizer is not backward compatible to past URLs.  Is this something we need to add (seems difficult)?  Or should we just update the existing gallery items to the new parameter format.  This is perhaps a more general question about when visualizers are updated.

The main thing I have not implemented yet, but hope to later, is more colouring options.

To see what it can do, check out some fun ones (note that these URLs only work with this PR).  These are all static to begin, but you can animate them.

http://localhost:5173/?name=Beatty+DNA&viz=Turtle&domain=3&turns=79+0+45&steps=5+2+6&pathLength=6000&bgColor=e9eee3&strokeColor=4b7a81&start=0+200&seq=OEIS+A001951

http://localhost:5173/?name=Tau+Goldberg+Machine&viz=Turtle&domain=2&turns=121+0&steps=40+100&pathLength=6000&bgColor=e9eee3&strokeColor=814b7c&start=0+5&seq=OEIS+A000594

http://localhost:5173/?name=Free+Wheelin%27&viz=Turtle&domain=3&turns=127+10+105&steps=40+25+5&pathLength=6000&bgColor=ffffff&strokeColor=5f402b&seq=OEIS+A005228

http://localhost:5173/?name=Hofstader&viz=Turtle&domain=3&turns=137+0+105&steps=15+10+5&pathLength=5000&folding=0+0+0&bgColor=ffffff&strokeColor=7792bb&seq=OEIS+A005228


katestange:
Sorry, something with the rebasing isn't right.

PR #406, opened 2024-07-27
FactorFence Visualizer

katestange:

<hr/>

This is a factor visualizer.  It takes the terms of the sequence s(n) and creates a bar of height log(s(n)) at position n.  The bar is divided into pieces of height log(p) for each prime p in the prime factorization (with multiplicity).

Some features:

1. There's one "highlighted" prime, which is always shown at the bottom of the bars.  If you click off the graph, highlight is removed.  If you click on a prime, it switches that prime to the highlight role.  
2. If you mouse-over the graph, information on the bar and prime you are hovering above is shown below the graph
3. You can use the mouse scrollwheel to zoom, and the arrow keys to move around; there are a few other special keys
4. It treats signs separately (checkbox turns this on/off):  the sign is removed before taking log and slapped back on afterward.

A few interesting ones:

http://localhost:5173/?name=5350&viz=FactorFence&terms=1000&highlight=0&seq=OEIS+A005350

http://localhost:5173/?name=abc+hits&viz=FactorFence&terms=1000&highlight=0&seq=OEIS+A130510


gwhitney:
OK, I modified the types so that the lint/typescript special cases wouldn't be needed. But I am observing a significant issue in that now when I scroll the parameter tab for the FactorFence visualizer, the visualization also zooms. It should only zoom when the mouse cursor is over the canvas. Are you observing that same issue? If so, I think it needs to be fixed up before this is further reviewed. If you are having trouble with that, let me know and I will be happy to help.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+    keyPressed() {
+        this.resetLoop()
+    }
+
+    mouseMoved() {
+        this.resetLoop()
+    }
+
+    mouseWheel(event: MouseEvent) {
+        let scaleFac = 1
```
gwhitney, line 591: Oops I think I should have just made the argument to mouseWheel be a WheelEvent both in P5Visualizer.ts and here, then the instanceOf test would not be necessary. When you are making changes, could you fix that? Thanks.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            // then with primeIsHigh = false
+            // in this way the highlighted prime sits
+            // at the bottom of the stack
+
+            // loop through factor to draw for this term
+            // from smaller to larger
+            for (let facIndex = 0; facIndex < factors.length; facIndex++) {
+                // get the next prime factor
+                const factor = factors[facIndex]
+
```
gwhitney, line 497: As far as I can see, you're not using facIndex anywhere in this loop below here. In that case, just iterate `for (factor of factors) {` ... then you don't need facIndex identifier at all and makes your intention clearer.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            // suitable for looping through to make the bars
+            // format: [prime, log(prime)]
+            const factors: bar[] = []
+            if (facsRaw) {
+                for (const [base, power] of facsRaw) {
+                    if (base != -1n && base != 0n) {
+                        for (let i = 0; i < power; i++) {
+                            const logPrime = BigLog(base)
+                            const thisBar = {prime: base, log: logPrime}
+                            factors.push(thisBar)
```
gwhitney, line 304: Since it doesn't depend on i, best to calculate thisBar outside of the power loop. And since logPrime is only used once and its expression is short, you may want to skip the identifier and just put BigLog(base) in the expression for thisBar.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+        // collect some info on the sequence in order to decide
+        // how best to vertically scale the initial graph
+        this.collectFailed = false
+        const seqVals = Array.from(Array(barsInfo.numBars), (_, i) => {
+            let elt = 1n
+            try {
+                elt = this.seq.getElement(i + barsInfo.minBars)
+            } catch {
+                this.collectFailed = true
+            }
```
gwhitney, line 238: In this case, can't you just shortcut all the computation below and just return the height of a 1-bar? (I mean when the get element fails?)

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            let sig = 1n // sign of element
+            if (elt < 0n && this.signs) {
+                sig = -1n
+            }
+            // can I simplify this??
+            if (elt * sig < 1n) {
+                // in case elt = 0, store 1
+                elt = sig
+            }
+            // store height of bar (log) but with sign info
```
gwhitney, line 248: I am confused by this code. Suppose this.signs is false and the sequence is entirely negative. Then doesn't this return all bar heights equal to 1, and doesn't that throw off your scaling?

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+        barGradient.addColorStop(0, colorTop)
+        barGradient.addColorStop(1, colorBottom)
+        this.sketch.drawingContext.fillStyle = barGradient
+        this.sketch.rect(x, y - height, width, height)
+    }
+}
+
+// bigint logarithm base 10
+// would be good to put this in shared math later
+// from https://stackoverflow.com/questions/70382306/logarithm-of-a-bigint
```
gwhitney, line 728: Should be in new mathjs when we update

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            const factorizationPrimesPre = factorizationPrimes.filter(
+                factor => factor < this.mousePrime
+            )
+            const factorizationPrimesMouse = factorizationPrimes.filter(
+                factor => factor == this.mousePrime
+            )
+            const factorizationPrimesPost = factorizationPrimes.filter(
+                factor => factor > this.mousePrime
+            )
+
```
gwhitney, line 658: You could also use Array indexOf() and lastIndexOf and then take three slices if you wanted to avoid filtering the whole array three times.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            else scaleFac = 0.97
+        }
+        this.scaleFactor *= scaleFac
+        this.graphCorner.y = this.graphCorner.y / scaleFac
+        this.resetLoop()
+    }
+
+    // set highlight prime by click
+    mouseClicked() {
+        if (this.mouseOn) {
```
gwhitney, line 603: Seems to have the same issue that it fires when I click on the parameter tabs.

gwhitney (commented):
OK I had a little time this morning so I did some code review. Not sure it's complete but hopefully a helpful start.

katestange:
> OK, I modified the types so that the lint/typescript special cases wouldn't be needed. But I am observing a significant issue in that now when I scroll the parameter tab for the FactorFence visualizer, the visualization also zooms. It should only zoom when the mouse cursor is over the canvas. Are you observing that same issue? If so, I think it needs to be fixed up before this is further reviewed. If you are having trouble with that, let me know and I will be happy to help.

Ok, I agree with this and noticed it also.  I think probably this is a "global" issue, not something each visualizer should have to handle.  That is, mouse events should be captured and then routed to the canvas only if the mouse is over the canvas (*and* not over a modal), and the default browser behaviour prevented.  So I think this is probably a modification to frontscope, not FactorFence.  Should it still go in this PR?  Or a separate one?  I found a few relevant things, e.g. [here](https://stackoverflow.com/questions/10313142/javascript-capture-mouse-wheel-event-and-do-not-scroll-the-page).  It looks like the idea is to add an event listener to the canvas when it is created.

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+        )
+
+        // maximum bar to compute
+        const maxBars = minBars + numBars
+
+        return {
+            minBars: minBars,
+            maxBars: maxBars,
+            numBars: numBars,
+        }
```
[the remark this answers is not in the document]
katestange: cool; next commit should fix this

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    checkParameters(params: ParamValues<typeof paramDesc>) {
+        const status = super.checkParameters(params)
+
+        if (this.seq.last < this.terms) {
+            this.terms = this.seq.last
+        }
```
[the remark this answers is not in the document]
katestange: ok moving refreshParams(); also thanks about the number of terms, fixing

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+            }
+            this.factorizations[myIndex] = factors
+        }
+        return firstFailure
+    }
+
+    setup() {
+        super.setup()
+
+        this.refreshParams()
```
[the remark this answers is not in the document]
katestange: ok

src/visualizers/FactorFence.ts
```diff
@@ -0,0 +1,738 @@
...
+
+        // set up terms first and last
+        this.first = this.seq.first
+        this.firstFailure = this.first // set factoring needed pointer
+        this.last = this.terms
+
+        // warn the backend we plan to factor
+        try {
+            this.seq.getFactors(this.last)
+            // we only wanted to send info to the backend,
```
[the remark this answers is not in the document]
katestange: excellent, done

src/visualizers/FactorFence.ts
katestange: Done.

src/visualizers/FactorFence.ts
katestange: got it

src/visualizers/FactorFence.ts
katestange: done

src/visualizers/FactorFence.ts
katestange: cool

src/visualizers/FactorFence.ts
katestange: yes, we have a separate comment chain about this now, so could resolve this

src/visualizers/FactorFence.ts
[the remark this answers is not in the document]
gwhitney: OK, sounds like something unfortunate is going on. I will need to take a look. Let me know when you've pushed commits you're happy with for the moment (i.e. addressing whatever else you're dealing with now, but not this point) and I will take a look. We definitely want this to work in an intuitive way!

src/visualizers/FactorFence.ts
gwhitney: Let's wait, to make sure the eventual fix for the wheel also fixes this issue.

src/visualizers/FactorFence.ts
katestange: Question:  this is a trade-off, since that code (if I understand your suggestion) would assume the factors are listed in increasing order, while this code does not.  So it's less robust in some way.  Is that a good tradeoff?

src/visualizers/FactorFence.ts
katestange: good call

src/visualizers/FactorFence.ts
katestange: Such a sequence is A002819.  You're correct and I've refactored.

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
katestange: committed

src/visualizers/FactorFence.ts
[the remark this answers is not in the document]
katestange: committed

src/visualizers/FactorFence.ts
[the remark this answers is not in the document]
katestange: committed

src/visualizers/FactorFence.ts
gwhitney: Good point. It never occurred to me they wouldn't be increasing. Filtering thrice just felt so redundant. But I won't stress about this. Feel free to make a change or not, and just go ahead and resolve this when you are content. My comments, especially this one, are just suggestions. If you want to use indexOf, you could check the increasing condition when you make factorizationPrimes in the first place.

gwhitney:
> I tried this, but no luck so far. But we have the same issue on the navbar, which I don't think is a child of the canvas?

Weird. Maybe I am misuderstanding how this works. I will pull and take a look.

gwhitney:
> I will pull and take a look.

Aha. I just found the following in the p5 source code:
```
// Bind events to window (not using container div bc key events don't work)

    for (const e in this._events) {
      const f = this[`_on${e}`];
      if (f) {
        const m = f.bind(this);
        window.addEventListener(e, m, { passive: false });
        this._events[e] = m;
      }
    }
```
So there you have it. p5 is explicitly grabbing all events that happen anywhere on the entire window, not just ones that happen on its own element(s). That's pretty poor citizenship, if you ask me, and I would never have thought p5 would do that -- that's why I assumed we were just not propagating events correctly. We will need to work around this somehow, and I agree, the workaround should be in p5Visualizer.ts rather than in individual visualizers. I will think on this -- too tired to try to work it out now, I just finished a big construction event here at PCMI/IAS.

katestange:
> So there you have it. p5 is explicitly grabbing all events that happen anywhere on the entire window, not just ones that happen on its own element(s). That's pretty poor citizenship, if you ask me, and I would never have thought p5 would do that -- that's why I assumed we were just not propagating events correctly. We will need to work around this somehow, and I agree, the workaround should be in p5Visualizer.ts rather than in individual visualizers. I will think on this -- too tired to try to work it out now, I just finished a big construction event here at PCMI/IAS.

Maybe the poor citizenship is so you can do things like [this](https://editor.p5js.org/guillaume/sketches/r1tYp8B6m).  That shows how one can pay attention to mouse position from within p5.

src/visualizers/FactorFence.ts
katestange: Ok, I thought about this and decided to use the indexing.  Because there are already other places in the code where it is assumed the factors are in order.  This relies on the backend returning the factors in order.  Committed.

gwhitney:
> Maybe the poor citizenship is so you can do things like [this](https://editor.p5js.org/guillaume/sketches/r1tYp8B6m). That shows how one can pay attention to mouse position from within p5.

That example uses mouseenter and mouseleave events, with listeners manually added to the canvas element, rather than automatically added (to the entire window) by the p5 library itself. In fact, mouseenter and mouseleave are among the few events that p5 doesn't touch.

So I had some time to contemplate how to proceed here. I in fact considered the possibility of our adding our own listeners to the canvas. But (A) there is some p5 boilerplate wrapped around the sketch's mouseWheel() and other functions, and to make our environment as much like "ordinary" p5 as well, we probably want that boilerplate and we don't want to copy its code, because that might change out from under us in a future release of p5. And (B) p5 is still putting numerous event listeners on the top level window no matter what we do -- even for visualizers that define no event functions, those listeners are there and they fire on every wheel event, etc., just to repeatedly check that nope, this sketch still doesn't have a mouseWheel() function. (Another example of unimpressive coding in p5 -- why check on every event? Seems like it should check at the beginning when the sketch is created, and not bother putting a listener if nothing is going to happen. The documentation doesn't say anywhere that you can add a mouseWheel() function to a sketch _after_ it is created, and adding methods to objects after the fact is a very unusual thing to do, so I don't think anybody would miss it if p5 lost the ability to start calling a mouseWheel() function that was added after sketch creation.)

So anyhow, I think it would be confusing if there were the p5 event listeners plus our own, and the p5 ones are going to be there anyway, nothing we can do about it, so we may as well use them. I then wondered if the base class implementations (in the WithP5 class) could check if the event occurred in the canvas and if not, drop them. But that would mean that the individual visualizers would be obliged to call super.mouseWheel() -- not so bad -- and then somehow check the return code to see whether or not to process the event; but that same super function with its return code would have to work on its own as an event handler in the more common case when the individual visualizer doesn't define an event handler. So then that got too confusing for me also.

So my intention at this point is instead in our sketch creation function (inside the inhabit() method of the base P5Visualizer class factory), where for example it generates the draw function with special code to change the cursor in case of a cache miss, to similarly handle certain of the mouse event functions specially and ignore them (i.e., never call the visualizer functions) if they occur outside the canvas. I think the ones that need special handling are mousePressed(), mouseClicked(), and mouseWheel(). I think we want to ignore all of those happening outside the canvas. I don't think we should do anything to mouseReleased(), mouseMoved(), or mouseDragged(), because if a visualizer gets a mousePressed it might legitimately want to know that the mouse is released outside the canvas; and it might want to track motions or drags into or out of the canvas. Any other events that perhaps need additional special handling? I guess there are key events -- I will have to check to see if they are "bleeding" into the visualizer as well, and if so, maybe check whether the canvas has focus and if not, ignore them.

Not sure when I will get to this, though, sorry.

src/visualizers/FactorFence.ts
gwhitney: Sounds fine to me. If you want to be extra careful about this, you could put code to check the factor order in the spot where you build the factorizationPrimes list. Or perhaps even better, just file an issue in backscope noting that primes must be in ascending order in factorizations, a fact that should (a) be added to the backscope documentation, and (b) explicitly tested in one of the backscope automatic tests.

katestange: I added an issue to the backscope.

katestange:
Ok, so this is ready to be reviewed or to get some help with, I'm not sure I can do more here.  I tried to work toward having the parameters dialog complain when the user-entered number of terms exceeds the number available, but I ran into two problems:  (1) I can't access `seq.last` and `seq.first` from inside the `params` object, so I can't display to the user the number of terms or check at that moment; and (2) if I add a validation in `checkParams()` then it produces all sorts of weird behaviour, probably related to the issues mentioned above about `refreshParams()`, so I kind of gave up...

gwhitney:
OK, I should have time later this week to look at/take care of the three outstanding items on this:

- [ ] Fix up P5Visualizer.ts per the above comment so that click and wheel handlers don't fire on events outside the canvas;
- [ ] Fix whatever weird thing is going on with refreshParams() in checkParameters()
- [ ] Get the allowed ranges into the description strings for parameters like the number of terms. (This may involve enhancing the framework so that those description strings can be updated dynamically, rather than just their initial value be used. However, that bit may work as things stand, I am not sure.

If anyone else sees/remembers any other aspects of this PR that need attention before this is merged, please post them here. Thanks.

gwhitney:
All right, I pushed a commit that prevents mouse presses, clicks, and wheel events that occur outside the canvas from being handled by the Visualizer. Please test and see if it works for you. However, I have not checked the item off in the checklist above because there is another wrinkle. Suppose you are running FactorFence, and you are currently viewing 1000 terms and you want to change that to 2000. So you click in the text entry box for the number of terms, and say you are not clicking so precisely, and the cursor ends up after the last "0" in "2000". You might naturally hit the left-arrow three times, but if you do, the cursor will move to just after the "2" _but_ the visualization will also scroll three bumps to the left. That behavior is a bit jarring to me. But how about if your last click happened to be in a blank area in the top bar that has the numberscope link and the name of the visualization etc. Should the visualization at this point respond to the arrow keys? If so, then it's not as simple as just checking whether the canvas has focus to decide whether a keypress should be paid attention to by the visualizer. Or maybe we don't mind the "double response" to arrow keys when used in the parameter tabs or to the "y" key if the person wants to name the visualization some name that includes a "y"? Anyhow, I need guidance as to how/when/where the keypresses should be responded to in order to finish this part off. Thanks for your thoughts.

katestange:
About keypresses.  Hmm, that's tricky.  I hadn't noticed it when I was messing around, so it's a fairly uncommon situation.  Three options I can think of:
(1) leave it alone; double behaviour is ok (jarring as you say)
(2) pay attention to keypresses only when mouse is on canvas (probably sometimes frustrating, i.e. an overcorrection)
(3) pay attention to keypresses only when canvas has focus (also probably sometimes frustrating/overcorrection)
Given the lack of other options, my current favourite is (1).  I'm worried that (2) and (3) or any combination thereof will result in confused users who are wondering why keypresses don't work.  There isn't any way to know if you are currently moving a cursor, is there, like "focus is in an entry field"?

gwhitney:
> (1) leave it alone; double behaviour is ok (jarring as you say)

I'm fairly against this because it feels like "we coded it wrong".

> (2) pay attention to keypresses only when mouse is on canvas (probably sometimes frustrating, i.e. an overcorrection)
> (3) pay attention to keypresses only when canvas has focus (also probably sometimes frustrating/overcorrection)
> Given the lack of other options, my current favourite is (1). I'm worried that (2) and (3) or any combination thereof will result in confused users who are wondering why keypresses don't work.

On the other hand, I don't disagree with you there, either. This seems like a difficult dilemma (and is presumably why p5 defaults to just handling all keypresses).

> There isn't any way to know if you are currently moving a cursor, is there, like "focus is in an entry field"?

I am not quite sure what you mean by "moving a cursor"? The issue occurs for just typing a "y" or an "h" when you are naming your visualizer, as well.

In our case, since we fully control the layout, we do in some sense have a list of all of entry fields, so in principle we can check each one to see if it has focus. And maybe that's what we have to do. But because these entry fields are in different components, that approach will keep the code from being as modular as it is now -- the visualizer will have to know about the name entry bar and the parameter tabs and so on, which right now it doesn't care about. And when we add something like the OEIS search bar, the list needs to be updated... We should keep brainstorming about how to handle these keypresses.

gwhitney:
> We should keep brainstorming about how to handle these keypresses.

There seems to be something called `document.activeElement` that could be helpful. One of us should put a console.log of that in the keypressed handler and see what it says in various situations in which we would, or would not, want the visualizer to be handling the keypress, and see if we can determine a criterion that could be used at the P5Visualizer level to decide whether a given keypress event should be passed to the visualizer to handle. If nobody else gets to this before I have a chance, I will do it.

gwhitney:
A comment to https://stackoverflow.com/a/30714894/5583443 suggests that
```
if (document.activeElement.tagName === "INPUT")
    console.log("An input is focused");
```
is our ticket, and noodling around, it does seem promising. When I get a chance, I will add a commit making P5Visualizers ignore keypressed events when an input is focused by the above heuristic, and we can see if that gives us intuitive behavior.

gwhitney:
OK, the keypresses seem OK to me now, so checking off the event handling above. Let me know if you agree. As I was working on this, I noted that if you enter '6' as your favourite prime, it accepts it, dutifully shows "highlighting prime: 6" under the bar graph, and then of course does not highlight anything. Feels a little inconsistent? Should it complain that 6 isn't prime, or highlight both 2 and 3, or something else? I will leave this to your decision, @katestange; please just let me know if you will be adding a commit to make a change to this behavior or if we will just forge ahead as is. Thanks.

gwhitney:
P.S. One could change the field to "highlight factors of" and then use the quick gcd algorithm to deal with an arbitrary integer input without ever having to factor.

katestange:
> P.S. One could change the field to "highlight factors of" (call the value `h`) and then use the ~quick gcd algorithm~ `h mod p === 0` to deal with an arbitrary integer input without ever having to factor.

Beautiful idea -- it extends functionality and at basically no cost.  I'll do this.

katestange:
Ok, pushed a few changes:
1. favourite number has to be positive, it will check for all factors
2. refactored the display of factorization under mouse 
3. the display of factorization now shows three colours matching the bar:  highlight, mouseover and plain
4. Change Y/H to I/K for stretching

src/visualizers/FactorFence.ts
gwhitney: Seems to be solved by the general event filters. Resolving.

gwhitney:
OK was looking into my second checkbook, about refreshparams in checkparameters, but then I b remembered that we had an in-depth conversation  about how we want to steer Numberscope away from forcibly changing your direct input to accepting any input that meets the formal validity requirements and then warning the user that it's actually going to _behave_ as though the different and more acceptable/usable number had been entered.

So I am reluctant to merge another example of the behavior -- altering the user's keystrokes -- that we are deprecating and will have to change anyway. So it seems to me that we will never need refreshParams to work from checkparameters, so there's not much benefit to spending time on that now.

But that leaves the question of how to change factorfence for merging now. First, clearly it should not change the terms param, but instead just behave as if the smaller number of terms had been entered. But it should then also provide feedback as to what it is doing. But there is not yet any "warn" status. So should we implement the "warn" status in this PR, or just not say anything about the actual terms being lower than the requested terms until a later PR that implements such warnings?

katestange:
> But that leaves the question of how to change factorfence for merging now. First, clearly it should not change the terms param, but instead just behave as if the smaller number of terms had been entered. But it should then also provide feedback as to what it is doing. But there is not yet any "warn" status. So should we implement the "warn" status in this PR, or just not say anything about the actual terms being lower than the requested terms until a later PR that implements such warnings?

Yes, exactly my concern also.  I am ok with just using fewer terms and then filing an issue to create the "warn" status very soon, but if you'd like warn implemented in this PR, I'm ok with that too (and could even try doing it).  Just let me know.

gwhitney:
All right well I don't mind using this as a vehicle to get that done. So if you want to take a stab, great. You will need to add a new ValidationStatus for warnings that leaves the whole thing "ok()" but which also displays the messages (say in orange rather than red, or something like that).

katestange:
> All right well I don't mind using this as a vehicle to get that done. So if you want to take a stab, great. You will need to add a new ValidationStatus for warnings that leaves the whole thing "ok()" but which also displays the messages (say in orange rather than red, or something like that).

Ok, I was being too ambitious.  I don't think I can figure out quite what's needed to do this!  Can you take a stab?

gwhitney:
OK, I have indeed taken a stab at it. Let me know what you think. I rebased on the new ui2 so that I would be able to search the OEIS while testing it. I encountered the following concerns that I think should be fixed before we merge this:

1. There seems to be an off-by-one error (I may have introduced it...): When I ask for three terms, I get two bars; for four terms, three bars; etc.
2. This situation (of point 1) is made slightly more confusing in that I believe that sequence values of -1, 0, 1 don't appear at all. Should you put dots or dashes at the baseline as placeholders for those terms or something like that?
3. With the switch to "favourite number" it seems that clicking on the canvas but not on a bar sets that favourite number to 0, highlighting everything (rather than nothing). I think it should set the favourite number to 1 (especially because on reload, the 0 is flagged as an invalid value for the "favourite number".

When you've made and pushed  whatever responses you feel are appropriate to these points, I will proceed with getting the available number of terms injected into the description string for the terms parameter.

gwhitney:
Ah, and point 3 above has revealed that the concern about which clicks get to a visualizer has not yet been fully resolved: if you have the FactorFence visualizer active and you open the sequence switcher and click in the OEIS Search entry area, the "favourite number" goes to 0, which does not seem appropriate. The point is that click is making it to the Visualizer because it is geometrically within the bounds of the visualizer canvas, even though it is inside a popup that is currently being displayed on top of that canvas. I will have to think on how to prevent this kind of click getting to the Visualizer, and can try to address it after @katestange has made whatever response she decides is appropriate to points 1,2,3 that I recently raised.

katestange:
Ok, I've pushed commits addressing 1-3.  I fixed the off-by-one error, I made outside clicks make favourite = 1, and I added a thin dash line to show a "placeholder" for terms that are 0,1,-1.

katestange:
Oh I see a tiny thing that needs fixing also, hang on.

katestange:
Ok, I fixed a bug in the colours of factors in the displayed factorization info.  @gwhitney you are welcome to do your next thing

gwhitney:
All right, I think I have fixed the issue that clicks on the popups (such as in the OEIS search bar) were bleeding into the visualizers. I also had to make some changes in terms of waiting in the Scope for sequences to fill their cache before trying to validate the visualizer, because a sequence doesn't know how many terms it has until its first cache fill, but that info is used in the checkParameters of some visualizers (including this one).

Try it out and see what you think. When I have a chance, maybe tonight, I will look into updating the description strings with the max number of terms.

src/visualizers/FactorFence.ts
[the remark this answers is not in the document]
gwhitney: OK, I think we have sidestepped this point about checkParameters and refreshParams(), so resolving.

gwhitney:
Oh, I also noted that when you use the `k` key to move the factor fence down, the caption with information about the keys you can use and the sequence value and its factorization does not move with it. So the bars start overwriting the captions. Is that intentional, or do you want to push a commit that moves the captions as well (or some other resolution to this point)?
   Thanks, Glen

gwhitney:
Actually, I had a few minutes to look at updating the prompt for the number of terms to show the maximum, and it turned out to be entirely straightforward. So I've pushed a commit that takes care of that. You can pull and try everything, and decide what you want to do about the captions moving or not when you move the bars up/down. Once you've let me know you're done with that (and have pushed anything you might want to, or decided it is OK as is on that score), I will do a final check/review and merge this visualizer -- there's nothing else on my radar for this PR.

katestange:
(1) I have a suggestion about the focus on textboxes:  enter should cause the textbox to lose focus.  I find myself changing a setting, then the arrow keys cease to work and I don't realize why and the only way to get back to having arrow keys working is to click on the canvas (which has its own effects, namely changing the favourite number, so not ideal), or to click somewhere outside an entry field on the parameters pane, which is a very specific action that I had to discover by trial and error.  But hitting enter to finish entering something in a text box is a very natural action.

(2) When I click a checkbox I also lose the ability to use arrow keys.  That's not a desired behaviour; I think checkboxes shouldn't be included in the "entry field focus stealing key activities" rule.

katestange:
May I suggest a wording change to the wording of the warning?  Currently it is: "Displaying all 16091 terms of sequence, fewer than the 1770000 requested."  How about "More terms requested than available; using only 16091."  It's shorter/punchier?

katestange:
> Oh, I also noted that when you use the `k` key to move the factor fence down, the caption with information about the keys you can use and the sequence value and its factorization does not move with it. So the bars start overwriting the captions. Is that intentional, or do you want to push a commit that moves the captions as well (or some other resolution to this point)? Thanks, Glen

Ok, I played around with it and I saw a few options.  Keep in mind the text can always be turned off:

(1) leave as is; it can overlap the graph
(2) have it scale with the graph; then it can end up partially off the screen
(3) have it disappear when the graph starts to overlap it; but disappearing might be annoying if you still want the text
(4) put a background box on the text so it obscures the graph when there's overlap; the size of the text changes with mouseover, so the box would always be popping around in size, which is unpleasant
(5) fill the entire bottom of the canvas with a background upon which to put the text, which the graph will slide "behind" when it moves around; this reduces screen real estate a fair bit.  It could be pushed low to make a sort of "footer" of text, but then your eye has to move further as you mouse over and play.
(6) use a font that is appropriate for movie captions (i.e. has some background or contrasting border); I think it looks worse
(7) have the text in a pop-up mouseover, like in histogram; I find this less visually appealing and too jumpy

I didn't have any other ideas, and decided that I personally prefer (1) ... and (5) is my second choice.  Any input or other ideas?

katestange:
> All right, I think I have fixed the issue that clicks on the popups (such as in the OEIS search bar) were bleeding into the visualizers. I also had to make some changes in terms of waiting in the Scope for sequences to fill their cache before trying to validate the visualizer, because a sequence doesn't know how many terms it has until its first cache fill, but that info is used in the checkParameters of some visualizers (including this one).
> 
> Try it out and see what you think. When I have a chance, maybe tonight, I will look into updating the description strings with the max number of terms.

This appears to work and block clicks.  Mouseover is still picked up by the canvas but in the case of this visualizer, I see no problem with that.  Perhaps some other visualizers would dislike the behaviour.

katestange:
The example image needed to be updated (it is now out of date with various changes).  Committed.

gwhitney:
> I didn't have any other ideas, and decided that I personally prefer (1) ... and (5) is my second choice. Any input or other ideas?

I guess I don't understand why the bars slide up and down with i/k but the text stays put; why not just slide the text up and down by exactly the same amount as the bars are sliding up and down?

katestange:
Ok, now it does.  It can still overlap the graph when using U/O however.  What are your thoughts on that?

gwhitney:
> It can still overlap the graph when using U/O however. What are your thoughts on that?

Well, it only happens when you turn on respecting signs, and then stretch the bars. So it feels less like a bug to me. I'm OK living with it.

But if it gives you discomfort, you could just calculate the minimum level and bar goes down to and make sure the legend is below that, unless that would put it off screen; otherwise it's just as low as it can be while still on screen (and then yes, it does overlap the bars, but at that point it's "forced" to -- there's no screen space left where it wouldn't overlap).

Since it seems like there will always be some chance of overlap, you might want to change the first line of the legend to use the green color from the palette rather than the purple. Just a thought.

I'll leave this point in your hands to wrap up or just to say that you're also OK with how it is now and you're leaving it be.

gwhitney:
OK, reworded the warning as suggested; text fields lose focus on enter; and boolean fields lose focus after every interaction.

However, I did encounter another concern: I got the "Tau Many Primes" visualization in a nice state with the bars the way I wanted and zoomed the right amount and then wanted to highlight the '2's to check if in fact this sequence is only odd when the index is the square of an odd number. But typing a 2 in the "favourite number" box reset the scaling and positioning. I think that's because by default updating a parameter always calls setup, and setup initializes all of those aspects. So that was frustrating. It seems to me that changing the highlight number should leave all the other aspects the same.

I see at least four ways to go here:

A) Leave the behavior as is --- maybe it's good to start with a clean slate whenever any parameter changes, even if sometimes people are a little frustrated with having to "get back to where they were".

B) Make the scale, left/right position, and up/down positiion (did I miss anything?) params as well. They would start with their default values, and the keystrokes would change them and use refreshParams() to show the new values, but you could type directly into the params instead/as well if you like.

C) Figure out where to initialize these internal values (maybe the constructor?) and then don't touch them in setup so that they will keep their state even when the favourite number (or number of terms, for that matter) is changed.

D) Override the parameter update function so that it can leave these internal values alone when the favourite number or number of terms is changed.

I'm leaning toward B because it will have the side effect of putting that info in the URL, making the exact view that someone has achieved reproducible when they share their visualization. But you may object to (B) on the grounds that t will complicate the ui in the visualizer parameters tab. We could minimize that complication by hiding all these view parameters under a checkbox (like with the "additional dot parameters" for the Chaos visualizer).

So anyhow, I will leave which way to go with this up to you. Just let me know how you'd like to proceed and if there are to be any changes, whether you would like to execute them or prefer that I do.

gwhitney:
P.S. A consideration about (B) from my last post is that it would cause all of those view aspects -- stretch, position, zoom -- to be kept when changing the sequence as well. You might prefer that whenever the sequence changes, the view resets to the "standard view". If so, that would argue for something like (C) or (D). I think (C) would work with the view variables initialized in presketch() -- I think that presketch() is called whenever the sequence changes but that it is not called when a param changes (someone would need to check on that to be sure).

katestange:
Regarding the pan/zoom parameters.  At first, I was leaning toward (B), but then it occurred to me that even if they are in the URL and shareable, the person who loads them next time may have a different canvas size or aspect ratio, so the effect might really be different anyway.  This is true of any visualizer, but it's particularly pertinent to these "positioning" parameters.  Then I thought about your comment on (C) and whether these are reset with a new sequence -- I definitely want to go back to 'standard' view with a new sequence.  And for that matter, loading the standard view is reasonable when we have a canvas resize or someone else loads the visualizer in a different aspect ratio.  The visualizer goes to some lengths to compute a good initial positioning based on the sizes of the terms and canvas, and giving that up would ruin a good thing and cause a lot of annoyance and messing about for the user.

So I guess I'd go with (C).  I also think this is a reasonable choice in the wider context of varied visualizers, because I think there may be other cases where we want to differentiate code that runs when a new sequence is loaded vs. when a parameter changes.

So to sum up, a possible proposal would be:

1. Reset to standard view on resize, share URL, or new sequence, but not on other parameter changes
2. Capability for visualizers to have code that runs on **new sequences** but not on parameter changes (possibly presketch() )
3. Capability for visualizers to have code that runs on **canvas resize** but not on parameter changes (does this exist?)

What do you think?  

Related question:  Does the current design allow for a visualizer to trigger certain code when parameter A changes but not when parameter B changes?

katestange:
I updated the text color as suggested.

katestange:
> OK, reworded the warning as suggested; text fields lose focus on enter; and boolean fields lose focus after every interaction.

This all seems to work as advertised.

gwhitney:
> So to sum up, a possible proposal would be:
> 
>     1. Reset to standard view on resize, share URL, or new sequence, but not on other parameter changes
> 
>     2. Capability for visualizers to have code that runs on **new sequences** but not on parameter changes (possibly presketch() )
> 
>     3. Capability for visualizers to have code that runs on **canvas resize** but not on parameter changes (does this exist?)
> 
> 
> What do you think?

1. All the points you make are reasonable, totally happy to go with this plan. I do not think presketch() will run on resize, but I think there is an explicit resize method that one can extend. So I think we can make a method "resetScaling" and call it in on presketch() and resize(). I am pretty sure that presketch() is called on both new sequence and new URL (and opening gallery/saved visualizer), so then I think that will fit the bill.
2. Yes, I think presketch() fits this.
3. Yes, I think there is a resize() method.

> Related question: Does the current design allow for a visualizer to trigger certain code when parameter A changes but not when parameter B changes?

Yes, there is a parameter update function, which currently just always calls reset(). But if you extend or override it, you can see what parameter has changed and take action accordingly. In addition, I am thinking we should have a "leave" option in parameter descriptions which, if specified to be true, means that adjusting that parameter (like the checkbox to reveal other parameters) does not reset the visualization. (But there is no such option at the moment, you basically need to roll your own using the parameter update function at the moment.)

So I think this can move ahead. I will assume that you would like me to do the implementation unless I hear that you want to take a stab at it.

gwhitney:
Whoops, nope, a quick experiment showed that unfortunately presketch _is_ called on any arbitrary parameter change. Ugh. So I will look into the framework and try to find an apt place to insert a method that _is_ called on new sequences but not called on an arbitrary parameter change, or else maybe I will change things so that presketch is not called on parameter changes, since to me it doesn't seem logical/reasonable that it should be called just for a parameter change. I am leaning toward that solution -- make sure presketch not called on parameter changes -- unless I hear feedback from anyone otherwise. But I will take care of the implementation.

gwhitney:
OK, I have adjusted the framework so that presketch(), as we had originally expected, is not called on resize or on parameter change (but it is still called for a new URL or a new sequence). It was a pretty small easy change, so if you don't think that's the right direction, it would be easy to back out. With that, let me know if you want to implement the planned changes to FactorFence in terms of when to standardize the view: on new sequence/URL and on resize, but not on other parameter changes.  (Since presketch() is not called on resize, there will need to be a "standardizeView()" or some such method that is called both from presketch() and in the resized() function. )

Some implementation notes:
* In presketch, this.sketch is not available, so in this plan in "standardizeView()", you can only do things that do not involve accessing this.sketch. Anything that does need sketch access has to be left in setup().
* In presketch, remember to `await super.presketch()`
* In your resized() function, you will presumably want to call "reset" after standardizing the view, to redisplay the visualization. By defining resized(), you will have prevented the default behavior on resize of just calling reset(). So you need to call reset yourself if you want it called.

I am fine either way with your making the changes about standardizing the view, or doing it myself if you prefer. Just let me know.

gwhitney:
P.S. Give me a sec to update the docs in making-a-visualizer.md about when presketch is called. Once you see a commit with the message "doc: Detail when presketch() is called" you can have at it, if you are planning to do the implementation.

katestange:
Ok, I can think of visualizers that would want to rerun preSketch() on change of parameters *if* that parameter affects the sequence precomps that take a while, e.g. suppose I added a "translate" parameter to the factor visualizer, so you could translate the sequence values by 1 (so t_n now becomes 1+t_n) and then factor -- all the factoring would need to be redone.   You might say, ok, that should really happen on the sequence side, we shouldn't have the visualizer controlling translation.  And you'd probably be right about that, certainly in this example.  Although perhaps there are situations where there's a more visualizer-based parameter change with a big precomp...  Is this a situation we want to worry about at all?

katestange:
> I am fine either way with your making the changes about standardizing the view, or doing it myself if you prefer. Just let me know.

If you go ahead that's great!  Thank you!

gwhitney:
> want to rerun preSketch() on change of parameters  ...  Is this a situation we want to worry about at all?

I think since this seems to be an unusual situation, it's not too onerous in such a case just to expect that the visualizer writer will call presketch() "manually" in their custom `parameterChanged()` function.

Getting going on the standardize view changes to FactorFence. Will let you know when I have pushed it and it is ready for you to try out.

katestange:
So.... my son and I were playing with factor fence and it occurred to me that.... the only actual purpose of the _Number of Terms_ parameter is to decide how far to pre-grab factorizations.  The visualizer doesn't actually draw all those bars until they are needed when they appear onscreen when zooming out.  And in fact, it's just annoying from the user's point of view to decide ahead of time how far they _might_ hope to zoom out, and have to up this parameter as needed. 
 It's just distracting that the parameter choice exists at all.  And something like the natural numbers as a formula or random number or whatever ought to let you zoom out as far as you want.

What it really ought to do is ask for more factorizations as you go (maybe staying ahead of you so you don't notice).  Why should it make the user wait a while at the beginning in order to factor a bunch of stuff that isn't even on screen to begin with?  

So maybe we should remove the Number of Terms parameter entirely, and implement some well-behaved background caching.  Should that be an issue to file for later after this PR?

(Side note:  This is a kind of funny thing to realize, because we did a lot of work to make that particular parameter behave well (e.g. the warnings capability), but of course that will be work well spent on other visualizers like turtle.)

Thoughts?

gwhitney:
OK, I believe I have the view settings reverting to the "standard initial view" when we want them to (new sequence, new URL, or canvas resize) and not when we don't want them to (parameter/checkbox change). Please give it a try and let me know what you think.

As far as the terms param goes, I agree that it serves no purpose in the context of FactorFence. It can't hurt to be able to scroll through all of the available terms. If you never get way out to the high-index terms, it won't hurt you that you could. And in fact, FactorFence is already being a very good citizen with respect for waiting for terms/factorization to show up. So the only code change needed is to rip out the terms param, and all should be well. As usual, let me know if you are going to make the modification or you'd like me to do it.

katestange:
I gave it to my six year old and he didn't succeed in crashing it!  Your commit seems to work well.  I noticed a few things that needed doing:
1. Check box parameters should be consistent in whether they use question marks (I took them out)
2. I removed the number of terms parameter and default to min(10000,available).  It's pretty hard to zoom or scroll enough to see 10000 terms, but *just* in case anyone does, it will load more terms as long as possible, which may be forever, e.g. when using random or formula.  By setting initialLimitTerms = 10 or something small, you can test this feature and experience the slight lag caused by extending the terms.
3. images needed to be updated again

katestange:
One final thing.  The Thue-Morse sequence (A010060), or any other sequence made of 0,1,-1, is a special case.  I noticed the text floats to the top in this case.  I added a line of code to fix this.

katestange:
Ok, as far as I know we are done?

katestange:
Darn my fix for Thue-Morse broke the Bernoulli numerators.  A027641.  Fixed the fix.

gwhitney:
> Ok, as far as I know we are done? (Assuming I didn't break anything new.)

When you bump up this.last, you should also call `this.seq.fill(this.last)` (but don't `await` it) so that in the background it starts factoring the new terms.

katestange:
Update:  I couldn't help myself, I added some info on the non-factored terms.  Now we should have the following:
1. if a term is -1,0,1, we see a dash placeholder
2. if a term wasn't factorable (too big), we see a question mark
3. now the factorization info will display as long as we are over the graph in a left-right way, and for dashes or question marks it will give the value of the term and say "(no factorization)"  (it used to be the mouse had to be on the graph itself, now it just has to be above the graph in horizontal regard, regardless of vertical positioning)

katestange:
The text wasn't totally optimal, so I improved the way it displays factorizations

katestange:
Ok, I really am done doing things to it.  I'm done.  I'm closing the editor.  You can take a look if you like!

katestange:
> When you bump up this.last, you should also call `this.seq.fill(this.last)` (but don't `await` it) so that in the background it starts factoring the new terms.

Oh, I just saw this comment.  Ok, I won't close the editor.  I'll fix now.

katestange:
Ok, now, I promise, hands off awaiting your feedback! :)

gwhitney:
Some questions and one recommendation:

1. Is it a positive or a negative thing that among all of the bars there are tiny places (horizontally)  where you can put the mouse such that you are between bars and no term is displayed at the bottom? Would it be better if when your mouse is  within the left-to-right region of the bar chart, then always some term of the sequence was selected as the current term to display?
2. In a similar vein, is it a positive or a negative that one gets these term displays when the mouse cursor is over the title/menu bar above the canvas?
3. Also, I think because (to be on the safe side) you prepare three or four more "bars" than can fit on the screen, when the parameter tabs are to the right of the canvas (which they are by default), then the term displays continue as you move the mouse horizontally into the parameter tab, about 1/3 of the way across the parameter tab, showing values and factorizations that you can't actually see on the screen, until it mysteriously stops at some point. How do you feel about that behavior?
4. Isn't there a direct calculation (using division by the bar width) you could do from the mouse position to tell you what sequence index it was on, so that all you need to do as far as detecting mouse overs on prime blocks is check the vertical positions of blocks vs the mouse pointer when you got to that one index, rather than comparing the mouse position to the geometry of every single prime block (where all of the comparisons for a given index that's not in the correct horizontal position are redundant)? Is it worth streamlining the calculation of term display/mouse highlighting in such a way?
5. In drawTerm, you don't need to recheck that `myTerm` is not -1, 0, or 1 in the bit that draws the question mark. Just put that part in the "else" clause from the part that checks whether to draw a dash.

As usual, I'll do a final review once I hear back that you've completed whatever you may do in response to these thoughts (or may not do, the only item where I feel strongly there should definitely be a code change is 5).

gwhitney:
PS would it be better for the text when you mouse over a question mark to be something like " (factorization unknown)" rather than "(no factorization)"?

gwhitney:
Oops sorry to pile up a few of these: When you mouse over a dash, you see e.g.
```
S(21) = 0
  = 0
```
I realize the second "= 0" is to fill the slot of the factorization, but maybe in these basic cases it's not really necessary to repeat in this way, and instead just say `S(21) = 0` with no second line?

katestange:
Ok, some responses implemented:
1. I left the tiny horizontal cracks... they are important aesthetically and I'm not sure it makes any sense for a mouse in that crack to show any particular factorization
2.-3. I now only show factorization when mouse is on canvas
4. I have not done anything on this one.  It would be a major refactor and it's complicated and not clear what is best.... I'm doing it the way it is because that way I can do the mouse check exactly when the bar is created, to avoid having to store all the heights somehow.  I do see that my originally fairly simple design choice complicated issues as I added functionality, though.  So I'm not totally sure how to reimagine this.  If you think it's important I will.
5. done
6. now (factorization unknown)
7. no more =0 =0 etc.
8. I aligned the factorization = sign with the original = sign for aesthetic reasons

katestange:
It suddenly occurs to me that instead of putting a question mark it would be prettier to put an empty bar (no colouring).  Then you still get the term size, just not the filled in info.  I'll do that.

katestange:
Ok, I think the empty factorizations are much prettier now, what do you think?

gwhitney:
> Ok, I think the empty factorizations are much prettier now, what do you think?

Yes, _much_ better. Thanks for a good idea. But the bare thin outlines are a bit hard to see, especially when I zoom out a bit. Maybe a uniform grey or light purple (in the hue of the regular bars) fill? Just a thought.

gwhitney:
If you try a uniform fill but like it better with them empty, I am OK with it.

katestange:
Ok, will try that.  I also noticed another small bug.

katestange:
Which do you like?
Screenshot from 2024-08-18 19-34-14
Screenshot from 2024-08-18 19-34-04
Screenshot from 2024-08-18 19-33-26

katestange:
Ok, fixed bug and went with pale non-outlined bars for non-factorization.  The reason I liked the empty ones was the clear implication that we don't know what's "in" them, i.e. we don't know the factorization.  But the pale bars are prettier.  I hope it's still clear the info is missing (not that the bars are all suddenly prime)?  I'm worried about whether they are different enough.

gwhitney:
> Ok, fixed bug and went with pale non-outlined bars for non-factorization. The reason I liked the empty ones was the clear implication that we don't know what's "in" them, i.e. we don't know the factorization. But the pale bars are prettier. I hope it's still clear the info is missing (not that the bars are all suddenly prime)? I'm worried about whether they are different enough.

I like the top one best and I think that's what you went with. I think the distinction from the ones with gradient is totally clear -- these look totally drained of life, very blah by comparison. I think it totally works and is prettier than the outline-only. I will do final review and hopefully merge when I can.

gwhitney:
Do we want to do anything about the numbers/factorizations  that are too long to fit on one line in the value/factorization display? E.g., put an ellipsis in the string of digits and indicate how many digits in all? Or split the representation onto multiple lines? Or leave them as they are, just extending off the right of the canvas into nothingness? Or...?

katestange:
> Do we want to do anything about the numbers/factorizations that are too long to fit on one line in the value/factorization display? E.g., put an ellipsis in the string of digits and indicate how many digits in all? Or split the representation onto multiple lines? Or leave them as they are, just extending off the right of the canvas into nothingness? Or...?

Sorry, I was lazy on this count, but you're right, ellipsis looks better.  I've done this now.  Note:  the way I implemented it, it keeps attempting to draw the rest of the factorization's primes after the first ellipsis occurs, but just fails until it runs out of primes.  I don't think this is a performance issue, but is it a code style issue?  It would be more complicated to add a flag to control flow and stop attempting to write the couple extra primes in rare cases.

gwhitney:
Very nice!  Two comments/questions:
1. It seems to me that by "it fails until it runs out of primes" you mean "it draws ellipses off screen". I don't see too much point in that: I'd recommend you either do nothing in textCareful if textLeft is already beyond this.sketch.width, or leave the loop in printing factors if textLeft gets bigger than this.sketch.width, whichever seems better to you.
2. Do you want to or want me to add to the ellipsization so that it shows the number of digits? There a lots of possible ways to do that. One tempting way that should be very recognizable  without special knowledge on the part of the viewer is to switch to floating point notation, i.e., write a 1345-digit number starting with 8675309 and so on as `8.675309...e1344` if that is all that fits. Or something like `~8.675309e1344` is slightly more compact, with the `~` supposed to mean that the value is no longer exact, if you think that is understandable. Or if you don't like introducing the decimal point, we could do `8675309...[1345 digits]` or `8675309...[+1338]` if you think the latter is clear enough (I think it's a bit cryptic myself, but couldn't think of anything clearer that is compact, other than floating point).

I am contemplating item 2 because I could imagine circumstances in which the magnitude is of interest, but the raw ellipses lose that information. Or we can just leave raw ellipses. Your call.

katestange:
> 1. It seems to me that by "it fails until it runs out of primes" you mean "it draws ellipses off screen". I don't see too much point in that: I'd recommend you either do nothing in textCareful if textLeft is already beyond this.sketch.width, or leave the loop in printing factors if textLeft gets bigger than this.sketch.width, whichever seems better to you.

Ok, thanks, that was my question.  I'll do that now.

> 2. Do you want to or want me to add to the ellipsization so that it shows the number of digits? There are lots of possible ways to do that. One tempting way that should be very recognizable  without special knowledge on the part of the viewer is to switch to floating point notation, i.e., write a 1345-digit number starting with 8675309 and so on as `8.675309...e1344` if that is all that fits. Or something like `~8.675309e1344` is slightly more compact, with the `~` supposed to mean that the value is no longer exact, if you think that is understandable. Or if you don't like introducing the decimal point, we could do `8675309...[1345 digits]` or `8675309...[+1338]` if you think the latter is clear enough (I think it's a bit cryptic myself, but couldn't think of anything clearer that is compact, other than floating point).

Concerns:  One concern is that this is doable for the raw value, but seems pretty complicated for the factorization, since each prime is its own number with its own floating point notation, presumably.  So that's a whole kettle of fish.  Also, even just for the raw value, the approximate size of the raw value is shown on the graph, and this situation of wanting the exact number of digits without the exact value is a getting a bit "niche".  

OTOH, the notation `8675309...[1345 digits]` seems the most widely understanding/accessible and most aesthetic, so I could be convinced to do that for the raw value and leave the factorization as is.  It's slightly finicky because you need to know the number of digits of the number of digits.  Maybe I'll try it and see.

katestange:
Show digits is committed.  Take a look.

gwhitney:
Looks good. I like it. I don't know of any other items, so I will do final review and merge if I don't see anything.

BTW on a small coding note: I advocate against using boolean arguments for two-way behavior switches (as opposed to say a logic function that takes a boolean value). That's because the call `doSomething(arg1, arg2, true)` is undecipherable for someone reading the code at a call site -- what does that `true` instruct `doSomething` to do? So if this textCareful function were part of an API, I would basically insist that the argument be an enum with values something like DIGITS and NONDIGITS. But for a little local function that's only used a couple places, it's OK as is.

gwhitney:
OK, I found a bug in that the digits of a very long value of a term were being miscounted because the characters of the `S(162) = ` part were being included in the digit count. But as I got to fixing that, I just couldn't resist reducing the number of comparisons of the mouse position to various rectangles. So I gave in to the temptation to just compute the mouseIndex directly from the mouseX coordinate, and then only do the "extract prime" checks in drawing that one stack of bars. Sorry for mucking in your code, I just couldn't shake my aversion to lots of redundant tests.

Anyhow, doing this work left me with two extremely minor questions:

1. There is a universal convention in the OEIS that in a given entry, the sequence that entry about is represented as a(n). In this visualizer's bottom text,  we use S(n) for essentially the identical thing. Do we want to change that to a(n) to match the OEIS convention?
2. Looking at a sequence that changes signs back and forth and has both 1 and -1 entries and other positive and negative values, it's visually clear that there is a midline to the bar chart and that positive terms are represented _above_ that midline, and negative terms _below_ it -- except for -1, which is represented identically to 1 and above that midline. So that representation creates a  bit of visually confusion. So I'd recommend that -1 terms be represented the same thickness as 1 terms, except just below the midline. But that proposal then begs the question of how to represent 0s? Unless you leave a pixel of vertical space between the positive and negative terms, there's no way to put it vertically between the positive and negative terms. So maybe twice the height of the 1s and -1s and centered on that midline? Maybe in grey or maybe dotted? But then it seems weird to have 0 be _thicker_ than 1 or -1. So... your thoughts on this point?

We could of course just leave things as they are, on both/either counts. Let me know and we can proceed accordingly.

katestange:
I love how detail oriented you are!  I actually was unhappy with S( ) but wasn't sure what else to use, but of course you are correct, lowercase a is totally the right thing!    As for the bars, I separated the positive and negative terms by one pixel, then used that one pixel for 0 terms, and put 1 and -1 above and below.  let me know how you think it looks.  I'm on the fence.

katestange:
Extra thought:  I was looking at your code changes, and I realized that maybe now that I have moved all negative terms down one notch, the mouseover detection will be off by one pixel?  Well, if you think the new approach looks good enough to keep, then maybe that needs fixing.  But I'll wait to hear from you first.  Feel free to just fix that and merge if you are happy, or tell me to do it.

gwhitney:
I don't see that moving the negative bars down a pixel has introduced an error in the prime sensing, because you adjust barStart to do that, and the prime sensing still uses barStart just like you had been doing. So it should all be consistent, and zooming way in it seemed fine behaviorally to me -- does it seem off to you? It can be tricky to know exactly where the "hotspot" of the mouse cursor is...

I like that all of the positive terms now line up, and the negative terms line up, and I like that in a sequence that has positives and zeros the zeros dip just slightly lower. So I am fine with leaving this behavior as you have it now, if you are. Let me know if in the end you are OK with this scheme for representing 1, 0, -1 and if so, I will do final review and hopefully merge.

gwhitney:
Oops, nope, found a bug, sorry: the ellipsization showing the number of digits is not playing properly with zooming. To reproduce, do FactorFence on A027641, the Bernoulli numerators. Pan over into the forest of tall unfactored bars. Mouse over one of the first ones that causes ellipsization. Now zoom in and out, keeping your mouse pointer on that term. As you zoom out, the type gets smaller so that the number would fit without ellipsization, but it still ellipsizes. And as you zoom in, the type gets bigger but it doesn't ellipsize enough, so it goes back to being just a bunch of numbers running off the edge of the screen.

katestange:
Thanks for finding the bug!  Fixed!

gwhitney:
Looks great. Here's a **very** tiny point, that I am not sure it is worth doing anything at all about: Suppose you view a sequence like Thue-Morse so that all you see are trival cases on screen. Then when you hit U/O, nothing seems to happen, This is by design, since after all the log of 1 is 0 and 0 times any scale factor is 0. But the person making the visualization just has the experience of hitting a key, that the legend says will "stretch", and yet nothing appears to happen. Should we do anything about this, and if so, what? Some possibilities:
* After each stretch keystroke, temporarily display the stretch factor, say in the text area, maybe to the right of the "U/O stretch legend" (and then I guess we would want to do something similar for zoom and pan? Or not because those always seem to do something?)
* After any of the keystrokes, temporarily highlight the corresponding letter in the guidance text.
* Provide temporary visual feedback icons for each of the operations: two arrows pointing away from the center up and down for stretch, four arrows pointing to/from the corners from/to the center of the canvas for zoom in/out, etc.

Or like I said, we could just do nothing as this is a very fine point. Let me know your thoughts.

gwhitney:
Hmm, just noticed another very odd behavior that I _think_ is a bug?: 

View A000045 V-F numbers in the standard view and zoom way out so that you can see the whole sequence up to where we start to lose factorizations. Now slowly scrub your mouse cursor horizontally across the body of the the of the collection of bars, somewhere above the baseline so that you are crossing a lot of bars. What I see when I do this is that from time to time as I move the mouse, the guide text "Click select; ..." flickers to lavender and back to green, whereas I think it should just always stay green. The flickering is a bit distracting and I'm not sure what it means if anything.

katestange:
I responded to your "tiny thing" by making key presses show visually as a colour change on the key instructions; and I improved the feel of the scaling (which was previously too fast on sequences with giant terms, and could also reverse positive and negative, which seemed silly) by moving from additive to multiplicative scaling.

gwhitney:
Great, the feedback on keystrokes is nice. And now only the "Click select;" part of the top line of the legend flickers back and forth between green and lavender when I scrub the mouse cursor along a zoomed-out V-F instance of FactorFence. Does that mean something, or is it a bug?

katestange:
Ok, ready for final review (I hope!)

gwhitney:
OK, I have completed another round of code/behavior review. Many observations. There is one at the beginning  that requires action, marked by an asterisk (`*`) that it would be nice if you were to do, and a couple  at the end that we might want to act on even though this has been a long saga, marked with a question mark (`?`) -- for those I am happy to do it myself if you agree we should take action, or equally happy with your doing it, and also fine with your saying "Enough already!" on these. The other items are just informational.

* `*` The new message that highlighted primes sink to the bottom was not in general true. For negative terms, they rise to the top. I put in new wording. Please check out what I wrote and edit as you see fit. And when you are happy with the text, please update the pic in which the text appears. Thank you.
* I realized that we were never really using modulo, just checking when one integer divided another. So I made a function for that in shared/math and used that instead.
* I noticed that == was being used everywhere but JavaScript convention these days is to use === except in the rare cases that what you mean is ==. So I changed those, and in so doing made isTrivial just an ambient function, since it has nothing to do with the FactorFence structure (doesn't need any properties of FactorFence).
* I had inadvertently introduced a bug earlier so that if you snuck the mouse off the canvas directly from a bar without crossing whitespace, the prime you were on would stay green. Fixed that.
* Removed some unused variables and function arguments.
* I thought it was odd/confusing that we were reordering the factors in two places to make the divisors of this.highlight come first but doing it in completely different ways with different code. So I wrote a single implementation for doing this and then used it in both places. This refactor shortened the code overall a bit.
* Since just about every call to textCareful was followed by adding the number of pixels drawn to textLeft, for convenience I just made textCareful return that number.
* In bottomText, I refactored to detect conditions when we were done and return from the function, rather than enclosing the rest of the function in the then-portion of an if-statement. This refactor reduced the number of spaces of indentation needed in that function by quite a bit, making it easier to read.
* `?` When a term is known to be prime, is it worth it to print its value out twice rather than to just label it as prime? I get the technical logical consistency with showing the factorizations of composites, but is it perhaps more immediately humanly understandable to  just put `(prime)` in parentheses after the (what's now the first) shown value (and get rid of the repetition)? Or if there is a worry about the `(prime)` going off screen for long primes, we could instead put `= a prime` or `= prime` or `(prime)` on the line below (instead of the repeated value).
* `?` I think mousewheel zoom is not optimally implemented, and the defect makes it pretty visually confusing for me. I think the usual convention these days in mousewheel zoom is that the current mouse position should be the center of the dilation, so that what your mouse is on stays put and everything expands away from or condenses toward it. I think this would be particularly useful when you are zoomed way out, and you want to inspect some particular bar. Right now, that's tricky, because as soon as you start zooming, the bar of interest moves away from where you were looking, This is particularly confusing when all of the bars in your neighborhood that's showing go off the edges of the screen vertically -- then there is no clue from the "shape" of the bar chart that it has in effect shifted left or right as a result of the zoom. Technically speaking, if we made this change, a zoom would be a pan (possibly with both vertical and horizontal components) together with what's now the zoom. Note I am not quite sure what the up- and down-arrow zoom should do.. perhaps they should be left just as they are now (using the origin as the dilation center)? Or they could check if the mouse pointer happens to be on the canvas and if so use that as the dilation center, otherwise use the origin as the dilation center.

katestange:
Thank you for all the review comments, those are wonderful!  And thank you for doing the code refactoring.  I had noticed that bit about two ways to make the highlighted prime come first but I have yet to break free of my "if it ain't broke, don't fix it" mentality and didn't refactor. :)  Thanks.

> * `?` When a term is known to be prime, is it worth it to print its value out twice rather than to just label it as prime? I get the technical logical consistency with showing the factorizations of composites, but is it perhaps more immediately humanly understandable to  just put `(prime)` in parentheses after the (what's now the first) shown value (and get rid of the repetition)? Or if there is a worry about the `(prime)` going off screen for long primes, we could instead put `= a prime` or `= prime` or `(prime)` on the line below (instead of the repeated value).

I think `(prime)` is a great suggestion.  Happy to do this.

> * `?` I think mousewheel zoom is not optimally implemented, and the defect makes it pretty visually confusing for me. I think the usual convention these days in mousewheel zoom is that the current mouse position should be the center of the dilation, so that what your mouse is on stays put and everything expands away from or condenses toward it. I think this would be particularly useful when you are zoomed way out, and you want to inspect some particular bar. Right now, that's tricky, because as soon as you start zooming, the bar of interest moves away from where you were looking, This is particularly confusing when all of the bars in your neighborhood that's showing go off the edges of the screen vertically -- then there is no clue from the "shape" of the bar chart that it has in effect shifted left or right as a result of the zoom. Technically speaking, if we made this change, a zoom would be a pan (possibly with both vertical and horizontal components) together with what's now the zoom. Note I am not quite sure what the up- and down-arrow zoom should do.. perhaps they should be left just as they are now (using the origin as the dilation center)? Or they could check if the mouse pointer happens to be on the canvas and if so use that as the dilation center, otherwise use the origin as the dilation center.

This is a very good point.  I think I probably did the easiest thing when I began and never came back to this, as I got used to it.  But you're totally right, it should zoom on the mouse pointer.

I think the up/down should continue to use the origin (I suspect if it zooms on the mouse pointer, a user won't realize that and just experience it as random zoom centering).

I'll try to do this, although there's a good chance my head will explode in the process ;)

katestange:
Ok I took care of * and the two ?'s, although my head did explode.  More seriously, I don't know if I did it the "right" way.

katestange:
Oh wait I forgot the pic again, will do that now.

katestange:
Ok, back to you!

katestange:
Hmm, it just occurred to me that when you drag it also selects, so probably we want to avoid that, and only select when you click without dragging.  I have to head home now, so if you are looking at this soon/now, you could go ahead on that, otherwise I'll try when I have time again.

gwhitney:
Awesome. The new mousewheel zoom is great, and the dragging is really nice. I will take care of making it so dragging doesn't highlight the bar you happened to be on.

gwhitney:
OK, I think I have the "click -> highlight, drag -> pan graph but don't highlight" adjustment working. But I definitely want someone else to bang on it before this gets merged, so let me know how it works. Especially, I put in a little "fuzz" on the decision as to whether it was a click or a drag, so that if the mouse happens to jitter a bit then it won't be detected as a drag. Let me know how that threshold feels in practice.

Thanks so much and I once again think/hope we are very close! Great idea on the dragging, that hadn't even occurred to me. And as a bonus, I have a touchscreen, and so I can now pan the bar chart around with my finger =)

katestange:
Thanks for doing that!  I am having fun finding congruence for Ramanujan Tau.  You can easily stumble on primes that seem to have regular behaviour like 2586714457, 2383, 3961159, ...  Anyway, it seems to work well, although it does slow down when you zoom out a heck of a lot.  Not sure there's too much we can do without avoiding gradients.  Anyway, in short I'm playing and have no further suggestions.  You can review again or merge or whatever you think!

katestange:
Ok, I'm getting a weird behaviour, can you verify?  Go to http://localhost:5173/?name=OEIS+A000594&viz=FactorFence&highlight=2411&signs=false&seq=OEIS+A000594 and then mouse scroll out a bit, then click the favourite number entry field.  Does the graph disappear for you?

gwhitney:
Yes actually I can reliably produce the "graph disappear" by simply loading the URL you gave, no need to zoom at all, just highlight all the text in the favourite number entry field, and then click just to the right of the text so the test entry cursor goes to the end of that entry field. The graph disappears.

I will try ignoring mousepresses outside the canvas.

gwhitney:
OK, give that a shot when you have a chance.

katestange:
It's working great!   In the famous words of children, "Are we there yet?" :)

gwhitney:
Well, the only thing I have noticed so far is that when you hit (for example) the change sequence button, and the popup with the previews of all of the sequences come up, the green highlighting of the moused-over prime continues happening with the FactorFence in the background as you move your mouse cursor around over the popped-up preview pane. It's a bit distracting. Does this need to be fixed before we merge FactorFence? While you contemplate that, I will continue beating on it.

gwhitney:
For that matter, the thumbnails also highlight green "live". Should they? Or should they be static? And it is possible to get you mouse in places where as you move it, both one of the thumbnails and the original FactorFence in the background are flashing different primes green/purple, which is especially distracting. So I feel like something should be changed before we merge this, but I am a little unclear as to what the most desirable behavior is for these situations. Please share your thoughts, thanks.

gwhitney:
Some other interface questions to contemplate as this is on the cusp of merging:
1. On reflection, it does not seem rational that the left/right arrow keys pan left/right, the up/down arrow keys zoom, and the I/K keys pan up/down. It would seem to me more consistent to have all four arrow keys pan, and (say) +/- keys zoom (those keys are very commonly associated with zoom, e.g. in inkscape and LibreOffice).
2. Given that thanks to you we now have drag, it no longer seems necessary to mention the keys for panning. I'd suggest we change the "microdirections" line in the legend to "Click to select, drag to pan, +/- keys to zoom, U/O to stretch". (But leave the arrow keys also for panning, just unadvertised.)

gwhitney:
Another:
3. When the favourite number is 1, should the orange legend change to something like "Not highlighting any factors", or "Not highlighting (favourite number is 1)"? I could go either way on this.

gwhitney:
Whoever next makes a commit to this PR needs to take out the "terms=1000&" part of the Gallery Tau Many Primes specification in the source code, as there no longer is any such parameter for the FactorFence visualizer.

katestange:
> Well, the only thing I have noticed so far is that when you hit (for example) the change sequence button, and the popup with the previews of all of the sequences come up, the green highlighting of the moused-over prime continues happening with the FactorFence in the background as you move your mouse cursor around over the popped-up preview pane. It's a bit distracting. Does this need to be fixed before we merge FactorFence? While you contemplate that, I will continue beating on it.

I noticed that, but didn't feel obliged to fix it.  I could go either way.

katestange:
> For that matter, the thumbnails also highlight green "live". Should they? Or should they be static? And it is possible to get the mouse cursor in places where as you move it, both one of the thumbnails and the original FactorFence in the background are flashing different primes green/purple, which is especially distracting. So I feel like something should be changed before we merge this, but I am a little unclear as to what the most desirable behavior is for these situations. Please share your thoughts, thanks.

Well, we could maybe just turn off mouse events getting to p5 for the thumbnails and while the pop up is open.  Is that something that's easy to do?  It might be better.

katestange:
> 1. On reflection, it does not seem rational that the left/right arrow keys pan left/right, the up/down arrow keys zoom, and the I/K keys pan up/down. It would seem to me more consistent to have all four arrow keys pan, and (say) +/- keys zoom (those keys are very commonly associated with zoom, e.g. in inkscape and LibreOffice).
> 2. Given that thanks to you we now have drag, it no longer seems necessary to mention the keys for panning. I'd suggest we change the "microdirections" line in the legend to "Click to select, drag to pan, +/- keys to zoom, U/O to stretch". (But leave the arrow keys also for panning, just unadvertised.)

Ok, so the main issue I have with +/- is that the browser responds to these by default -- I use that functionality with some frequency in general.  So we would have to override that if we wanted to use them, and I'm not sure we want to -- it seems useful to me actually to use them as browser keys, as applied to the numberscope page, they make the parameter panes vs. canvas bigger/smaller.  And people are expecting the browser behaviour.  

The arrow keys seem to me the most accessible and natural keys, so in principle they should do the most useful things people would like to do.  Zoom seems important. 

I am dissatisfied to use random keyboard keys for stretch etc., I admit.

open to further ideas...

Once we decide the keys, I agree, we could update the microdirections.

katestange:
> Another: 3. When the favourite number is 1, should the orange legend change to something like "Not highlighting any factors", or "Not highlighting (favourite number is 1)"? I could go either way on this.

Yes, agree.  Maybe the second of these is more informative.

katestange:
> Whoever next makes a commit to this PR needs to take out the "terms=1000&" part of the Gallery Tau Many Primes specification in the source code, as there no longer is any such parameter for the FactorFence visualizer.

Agree.  I'm going to be fairly busy today, so unlikely to have time before tonight.

gwhitney:
> arrow keys seem to me the most accessible and natural keys

So what about taking advantage of the drag, ditching pan keys altogether, and using left/right arrows to zoom and up/down arrows to stretch? Does that seem workable?

katestange:
> > arrow keys seem to me the most accessible and natural keys
> 
> So what about taking advantage of the drag, ditching pan keys altogether, and using left/right arrows to zoom and up/down arrows to stretch? Does that seem workable?

Hmmm, ok.  Then the instructions could be a more pleasing "click to select primes; drag to move graph; scroll to zoom; arrow keys to adjust"?

gwhitney:
OK, all these discussed changes are in. Let me know what you think, especially about the latest microdirections. Note I left U/I/K/O as pan keys, but didn't bother mentioning them in the microdirections because those are already kinda long and the pan keys are pretty unnecessary with dragging. But for completeness' sake, I did include U/I/K/O in the user guide documentation for FactorFence.

katestange:
I made two tiny tweaks.  I used IKJL now for panning, since that's the standard keys in vim, for example.  And I tweaked the wording of the instructions and updated the picture.  It seems very robust now and fun to play with.

gwhitney:
Awesome. The overall microinstructions are definitely going to go off screen for some reasonable browser dimensions. What shall we do about that? (A) nothing, (B) just always split to two lines, (C) implement our own line wrapping, (D) see if p5.js has a wrapping solution? I guess (D) sounds best to me, but it might not handle multicolor text. If that's the case, then we need to choose A, B, C or some other solution.

katestange:
I did (C) because (D) was not helpful with our multicolour setup.  But the multicolour setup was actually already pretty amenable to text wrap since the line was already broken into pieces.  I also made it so if the favourite number if huge, it can ellipsisize and stuff... check it out.

gwhitney:
Awesome again! I took the liberty of moving the semicolons to ends of "break units" because I don't think we should be breaking just before punctuation. (The small downside is now the punctuation also changes color -- is that OK?) I also changed the logic for not breaking just before the start of the second line so it wouldn't depend on the exact text -- that way we shouldn't have to worry about adjusting the breaking algorithm if we ever decide to re-word the text. Let me know what you think. I will admit that I am hoping we are just about done ;-)

katestange:
It looks good!  (I thought I had done that with the semicolons but must not have, and also thanks for the logic, I knew what I was doing was fishy, but I wasn't sure how to fix that, but I see what you did now.)  I can't think of anything else that needs doing!!

gwhitney:
I was getting weird text clipping artifacts when zoomed way out, so I tried always drawing the text at scale 1; that seems to have improved the issue and as a side benefit, the text now seems much more stable: the intracharacter spacing stays constant and so the text remains readable as you zoom in and out. Let me know what you think when you have a chance (definitely want someone else to test this before I merge!). Thanks.

katestange:
A little stress testing, I see!  I didn't see the original problem, but this version is showing no signs of trouble.  I did notice that in the text wrap commit I screwed up one space, though, so I fixed that.  I tried in several browsers.  It does get nice and laggy when you try to zoom out a lot, but no weird text behaviours.

katestange:
> > It does get nice and laggy when you try to zoom out a lot
> 
> Well, one way to deal with that is put a cap on the number of bars to draw per frame, and then instead of being laggy, it would only draw part of the image as you drag (for example). Not sure it's a win, and pretty sure we don't want to dive into that now. So unless I hear otherwise from you before I merge, I will just leave that concept be.

Yes, let's not dive into it now.  If you feel ready to merge, don't wait on me!

gwhitney:
As usual, trying to beat on this as much as I can. It was bothering me that the '(prime)' designator was not colored the way it used to be when it was a repetition of the prime. So I made it be that way, and in so doing found a nice little refactor that reduces the duplication of color-selection code. Let me know what you think; I'll wait for your feedback if any before I merge.

katestange:
That looks great and seems to work too.

gwhitney:
All right, tracked down and fixed a couple of regressions I created in providing the facilities FactorFence needed to work, including #415 (which can be closed now). The only tiny thing I noticed is that if you press the mouse button outside of the browser window, and then cross the browser window with it still pushed down, then the green bar highlighting and sequence term display does not operate (until you release the mouse button). This is fairly esoteric, and I am not even clear that those things should happen under these circumstances. Would you like the green highlighting/term display to work under these circumstances? Let me know -- if so, I will look into it; if not, I will merge :)

gwhitney:
Group discussion concluded that the current behavior with respect to click-dragging onto the Numberscope browser page is fine and in fact probably the desired behavior. So merging. (yay)

Issue #403, opened 2024-07-25
URL out-of-sync bug in parameter handling in ui2


katestange:
There's an important bug that I'm having to handle in enhancing turtle, but should properly be fixed in ui2, not in turtle.  Here's the situation:

Suppose that inside `checkParameters()` a parameter should be modified to conform.  For example, some visualizers will stop you from putting a "number of terms" that exceeds `this.seq.last - this.seq.first` (quite reasonably).  It would be nice to be able to simply replace a large number entered by the user with the maximum allowed by the sequence, and display this updated value in the parameter pane right away.  Note that no default value here is simultaneously reasonable and safe, as many OEIS sequences have very few terms.  So if you don't want the visualizer to load up in a default invalid state on such sequences, you want to just update the number of terms (including the default) if it is too large immediately to the sequence's allowable max.  This is also a favour to the user, who can't see the max in the UI at the moment.   So this is something we probably want to do, and I find myself wanting to do in Turtle.

Now, this check can't be performed earlier than `checkParameters()` since it needs access to `this.seq`.  So it happens inside `checkParameters()`, and then you have to call `refreshParams()` to display the modified value to the user, presumably at the start of `presketch()`.  This works.  

Except that in this scenario the URL *doesn't* update with the correction in `refreshParams()` and instead shows whatever too-large value the user typed (it appears as the user is typing).  So then the URL is out of sync with the visualizer's actual internal state, and hard reloading or sharing the URL will cause a crash (or whatever the bad value will cause) because a reload or fresh load doesn't seem to re-run `checkParameters()`.

gwhitney:
The description is totally understandable, and I think Chaos is already doing something similar with replacing its first and last indices with those of the sequence when they are out of bounds. Can the bug be exercised with existing Chaos visualizer? Generally speaking, I am happy to work on this, I just need an explicit "recipe" for making the bug occur.

gwhitney:
Ah, I see, it happens with A371958 under the current Chaos visualizer. That sequence only has 15 terms, so the "last" parameter shows 15. If you type 0 after that in the input field, the URL changes to include `last=150` even though the parameter box still shows 15. OK, I will take a look.

gwhitney:
@katestange would you be able to check if this has now been resolved in ui2, or if you can still reproduce it in this branch? If it seems resolved, feel free to close this issue; if not, please add any further details needed here, or at least a verification that it still exists. Marking this for alpha as it seems it is/was a bug in the new code.

[...]

Issue #401, opened 2024-07-24
checkParams() should run on defaults once sequence is chosen


katestange:
The default values for visualizer parameters may not be valid in the context of a particular sequence.  This could be checked by `checkParams()`, but it appears that right now `checkParams()` is not run until the user touches the parameters.  For example, if a turtle visualizer or a difference visualizer wants to use the first N terms by defalut, but the sequence at hand has fewer than N terms (there are some sequences with very few terms in the OEIS!), such a check may fail.  If it isn't checked, an error is thrown later on in the code when we try to fetch too many terms.

gwhitney:
Hmmm, I am confused. Scope.vue creates a visualization by calling Specimen.fromQuery. That calls validate() on first the sequence and then the visualizer. And validate() in turn calls checkParameters(). So I don't see how a validation can ever get on the screen without going through checkParameters(). Can you provide an explicit recipe for something to go wrong and an explanation of what exactly goes wrong? I tried Differences with all defaults (wants 20 terms)  on A371958 (only has 15 terms) and all seemed to go fine. Admittedly, Differences tolerates its parameter called "n" being larger than the number of entries in the sequence and just deals with it in draw(), so it doesn't seem like anything can really go "wrong" there, but maybe the issue you are identifying can cause real problems in another context?

katestange:
Ok, well, I noticed this while working on the `turtle-enhance` branch, so it's always possible it's actually something in that branch.  For now, here's how to reproduce:

1. On `turtle-enhance` branch
2.  Go to [http://localhost:5173/?name=errordem&viz=Turtle&pathLength=10000&growth=1&seq=OEIS+A156660](http://localhost:5173/?name=errordem&viz=Turtle&pathLength=10000&growth=1&seq=OEIS+A156660)
3. Observe that this has 10000 terms (which the sequence allows)
4. change OEIS sequence to A371606 (for me I've used this sequence before, so it's in the sequence chooser)
5. now the turtle ends after 16 steps, but it shows that it is using 10000 terms.  If you put the mouse in the 10000 terms input box and try to type, the box will auto-correct to 16.

My essential observation is that it doesn't auto-correct to 16 upon loading the sequence, only upon trying to update that box.  I'll try to look into this more when I have time.

gwhitney:
[...]

PR #410, opened 2024-07-29
feat: Operational OEIS search bar

gwhitney:

<hr/>

This commit makes the OEIS search bar operational, and reconfigures it according to group discussions. I don't think we ever discussed the details of the presentation of the returned results themselves, so I am very much open to feedback and suggestions on the workings of the bar.

This is the final planned PR for the sake of cleaning/merging #360, and as such, supersedes that PR.


katestange:
First main feedback:  The mouse is a "cursor" mouse over the part you need to click on to load the sequence.  It turns into a pointer finger only on the link to the OEIS url.  So I was confused trying to figure out where I can load the sequence.  Is this related to the fact that I get a `Failed to load resource: the server responded with a status of 404 (Not Found) http://localhost:5173/favicon.ico` error in the console?

katestange:
Another small observation:  There's no way to "back out" of a search except for deleting everything in the search box.... I kind of expected clicking somewhere else or something to back out.  I'm not sure if this is what you had in mind or not, and I'm not sure what is the best behaviour, but I thought I would mention that it caused me a little bit of "exploration click time" that didn't result in anything.

gwhitney:
> Is this related to the fact that I get a `Failed to load resource: the `

No, it's just that I need to add the CSS command to change the cursor. Will do as soon as I have the opportunity.

gwhitney:
> observation: There's no way to "back out" of a search

Happy to implement other "back out" gesture(s)  -- but just let me know what would be the intuitive action(s) to back out.

gwhitney:
> No, it's just that I need to add the CSS command to change the cursor. Will do as soon as I have the opportunity.

Yes, that was literally one line, which is now committed.

By the way, I meant to ask: should we bolden or highlight in some other way occurrences of the search string in the displayed names of the search results?

gwhitney:
I don't think I ever mentioned that what's now the second-most-recent commit here is supposed to fix #401.

src/components/OEISbar.vue
```diff
@@ -0,0 +1,166 @@
...
+            <p
+                v-for="item in results"
+                :key="item[0]"
+                @click="select(item[0])">
+                <a
+                    v-if="item[0].startsWith('A')"
+                    :href="`https://oeis.org/${item[0]}`"
+                    class="mono"
+                    @click.stop
+                    target="_blank">
```
katestange, line 36: Even with the pointer cursor over the whole search results, I'm finding myself clicking the A number to load the sequence and then surprised I get a new tab.  Maybe instead of making the A number the link, we could add  little "(more...)" link at the right hand side that's clickable and would open the OEIS entry.  Loading the sequence is probably the main activity the user wants.  Thoughts?  (This is a change I could make myself.)

katestange:
> > observation: There's no way to "back out" of a search
> 
> Happy to implement other "back out" gesture(s) -- but just let me know what would be the intuitive action(s) to back out.

I came back to it and played with it again today and I found myself just trying to click anywhere outside the search box/results area.  I think that would be a perfectly fine solution.  Then one thing to make sure is that clicking back in the search box, even without further typing, should make the results open again.

katestange (changes requested):
It does appear to fix the URL update issue.  It does appear to fix #403 and #401.

src/components/OEISbar.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+    <div id="search-bar">
+        <div id="search-wrapper" @click="doSearch($event)">
+            <div id="search-input">
+                <label for="oeis">
+                    <b>Search the encyclopedia</b>
+                    (<a href="https://oeis.org">OEIS.org</a>):
```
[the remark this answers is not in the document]
gwhitney: Completely agree and would be happy for you to add the attribute so that it will open in a new tab.

src/components/OEISbar.vue
gwhitney: Hmm. To me, a link is a link, and expect it to go elsewhere. Perhaps a little plus-in-dark circle button (or other icon) at the start or end of each line, to give you a visual target? (But still leave clicking anywhere on an entry working to add that sequence.)

gwhitney: I feel like all those "more...s" would look redundant and take up a fair amount of space. If you don't like the add icon idea, maybe we could just make the *ellipsis* at the end of the abbreviated description **be** the link -- or would that be too subtle?

src/components/OEISbar.vue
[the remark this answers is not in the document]
katestange: committed

src/components/OEISbar.vue
katestange: I tried an icon fix, as shown here:
Screenshot from 2024-08-11 13-43-14
Let me know what you think, it's committed to the branch.

gwhitney: How do you feel visually about the repetitive column of info icons? I take it you prefer it to making the ellipses at the end into links (and adding ellipsis-links to the descriptions that fully fit on the line)?

Here's another idea: Leave the OEIS IDs as links as they were originally, linking to the OEIS website. But: (a) style the rows as buttons rather than just rectangles (maybe slightly rounded corners, maybe that slightly raised pseudo-3D-ish look, that sort of thing) AND  (b) make them highlight or look activated as you mouse over them, inviting you to click on said button, AND POSSIBLY (c) even have a tooltip go up after a second or so of hovering over one of the buttons that says "click to add Axxxxx to your Sequence options" or something like that.

I mean, I am not set against your revision (thanks for submitting!) and if both of you think it is a good way to go we could do that. I just feel like it will be a bit visually tedious for the experienced user and that we should use some part of the layout that would naturally be there as the way to get to the OEIS, and everything else to add it as a Sequence option.

src/components/OEISbar.vue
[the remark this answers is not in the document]
gwhitney: This part looks good to me and seems to work, so please feel free to resolve this conversation if you are likewise comfortable (only not resolving it myself since you're the reviewer ;-)

gwhitney:
> I found myself just trying to click anywhere outside the search box/results area. I think that would be a perfectly fine solution. Then one thing to make sure is that clicking back in the search box, even without further typing, should make the results open again.

I was trying to implement this but found I had a question: Right now, when the search results box is popped up, clicking on one of the still-visible sequence preview cards closes the search results _and_ the Change Sequences modal and switches to the clicked sequence (even if that card is only partly visible, but the click is outside the search results. Should that behavior persist, OR while the search results is up, do you want a click outside the search results to close the search results and not do anything else? Thanks for letting me know.

katestange:
> I was trying to implement this but found I had a question: Right now, when the search results box is popped up, clicking on one of the still-visible sequence preview cards closes the search results _and_ the Change Sequences modal and switches to the clicked sequence (even if that card is only partly visible, but the click is outside the search results. Should that behavior persist, OR while the search results is up, do you want a click outside the search results to close the search results and not do anything else? Thanks for letting me know.

My reaction is that the latter would be less confusing.  But you can go with what seems to make sense to you.

src/components/OEISbar.vue
katestange: Ok, I see you proposing several options, and I'd like @Vectornaut 's input.  Options (with a few of my own thoughts on each):

(1) As in the screenshot above.  I see what you mean about visually repetitive and inefficient use of space.  The visual impact could be lessened by using the white icon instead of filled black one, and/or making it smaller.

(2) Ellipsis at the end of each title being a link.  This might be too subtle or easy to miss?

(3) Make rows into buttons.  Possibly complicated, both to program, and visually a little busy.

Thoughts, Aaron?

gwhitney: (3.5) We could leave the rendering of the rows as rectangles with text in them and just make the backgrounds highlight as you mouse over, and see if that's enough of an indication that you should just click on a row to select it.

But yes, we should discuss at meeting today.

gwhitney:
> > OR while the search results is up, do you want a click outside the search results to close the search results and not do anything else? Thanks for letting me know.
> 
> My reaction is that the latter would be less confusing. But you can go with what seems to make sense to you.

OK, I have committed code intended to result in this latter behavior. You can pull and try it out at your leisure.

gwhitney:
All right, I think that I've done everything open on this PR. From the latest commit message:

- Return to the OEIS ID being a link but add the Wikipedia
    "external link" icon
  - Highlight the description of each sequence as you mouse over it,
    to provide clue you can click on it (to add as a sequence). The
    color used is a lightened version of the "primary" Numberscope color.
  - Display a message when there are no sequences to show as a result
    of a search. Note this can occur because there are _too many_ matches,
    or _no_ matches, and there is no way to tell which is the case from
    the OEIS api response, unfortunately. The message reflects this.
  - Fixed bug that results were being cached under the wrong search term
    if the searcher typed more while the search was executing.
  - Make sure that enter/clicking on magnifier always fires search

Of course if you see anything else that needs changing/fixing or if there is anything in the code that needs to be improved, I am happy to develop this PR further. Thanks!

gwhitney:
P.S. In particular I don't love the phrasing I came up with for an "unsuccessful" search, namely

...... search for foobar gave no/too many results

So if anyone has better/clearer wording for this that is still relatively short, I would be happy to change.

katestange:
Two comments:

(1) clicking on magnifier still gives me a tooltip and stops search

(2) for the message, you could try '....... search for foobar failed (OEIS reports no results or too many results)'  One advantage of this phrasing is it attempts to pass the blame for the ambiguity ;)  But simpler may be '....... search for foobar failed (no results or too many results)'

Everything else works great!

gwhitney:
OK, I think I've taken care of both of those things.\

katestange:
When the search doesn't come back (axios/network/cors error) currently there's a popup and all interaction is interrupted for the user... but I think in the context of search especially, we should just continue (the same as if the user clicks "ok" on the popup) because it generally resolves itself at that point.  At least this is what's happening for me right now.  This might be appropriate to this PR or it might be something for post delft todos discussion?  Ideas?

gwhitney:
> I don't think you pushed the commits.

Oops, embarrassing. I have now.

katestange:
> > I don't think you pushed the commits.
> 
> Oops, embarrassing. I have now.

Ok, it looks good now.  The tooltip pops up mouseover for the magnifier, which is good.  Clicking causing search is essentially unverifiable at the moment for me, since my search is returning fast enough on typing (meaning, by the time I move my mouse and click, the search already fired).  But that's ok, I don't see any suggested change.

I think the only outstanding issue is the popups on network errors (comment earlier today), which maybe should be disabled during search?  Or maybe even entirely.  Those popups haven't served a purpose for me other than interrupting things that later fix themselves.

gwhitney:
> but I think in the context of search especially, we should just continue (the same as if the user clicks "ok" on the popup) because it generally resolves itself at that point. At least this is what's happening for me right now. This might be appropriate to this PR or it might be something for post delft todos discussion? Ideas?

OK, I will create and push a commit that catches network errors in the OEIS search and adds a message that the search was temporarily unavailable, please try again, and will _not_ cache that outcome.

katestange:
> > but I think in the context of search especially, we should just continue (the same as if the user clicks "ok" on the popup) because it generally resolves itself at that point. At least this is what's happening for me right now. This might be appropriate to this PR or it might be something for post delft todos discussion? Ideas?
> 
> OK, I will create and push a commit that catches network errors in the OEIS search and adds a message that the search was temporarily unavailable, please try again, and will _not_ cache that outcome.

Oh perfect, that sounds like a good solution.

gwhitney:
OK, I have added such a commit. I will be honest though: I spent a while mashing the keyboard and could not invoke the new error-handling code. Maybe the time of day? (Less load on OEIS server?) So let me know how it works for you when you try it.

katestange:
> OK, I have added such a commit. I will be honest though: I spent a while mashing the keyboard and could not invoke the new error-handling code. Maybe the time of day? (Less load on OEIS server?) So let me know how it works for you when you try it.

Ok, I just tried it and was able to provoke a server error in the console that was not popped up and not apparent in the browser/user experience.  So that's perfect.  I don't know why I'm better at making the server mad. :)

katestange:
Ok, I believe that everything is in order now with this, shall I squash and merge?

gwhitney:
You're the reviewer ;-) anyhow, fine by me if you squash and merge -- I don't know of anything else outstanding.

PR #473, opened 2024-10-26
refactor: make overall status and status of each parameter part of Paramable

### LINKED ISSUE #112: Validation errors should appear at location of error
In the context of the parameter validation errors, when validation errors occur, it would be nice if the instructions appeared not up at the top, but next to the offending input boxes (before or after). Then at the top, there could be a generic message to "see errors below" and the errors below could be highlighted with a red background highlight. Or something along those lines. (A la standard web form behaviour.) In fact, I'm a bit on the fence about whether to include a message up top at all. I'm thinking ahead to the panel that will hot-modify parameters in the draw view, and there, you don't want the panel to "jump downward" to give an error somewhere up above -- you just want the error to appear next to where the user is playing. But you probably want a top message in the pop-up model we have right now.  This was discussed in PR #108 where Glen said: "That does seem like a good idea, but it would require an extension to the ValidationStatus interface -- basically there would need to be a place to communicate the information of what the offending parameter is. So I think it is actually out of scope for this PR -- it would be great if you can file a separate issue that raises this. Also, I think any solution to this new point will have to resolve #58."

### LINKED ISSUE #113: Do pre-release test with "dumb" inputs
Before we go live / release frontscope, we should do a manual test where we supply weird/dumb inputs to the params. For example, we could provide way too many numbers (1111111111111111111111111111111111111111111), etc.

Related to https://github.com/numberscope/frontscope/pull/108.

gwhitney:

<hr/>

This PR puts new properties in every Paramable object, representing the ValidationStatus of every parameter and the overall ValidationStatus of the Paramable as a whole.  These statuses are then used directly in the user interface. These changes allow Sequences and/or Visualizers to

(a) Make sure that their messages show up in the desired place: at the top of the tab for the object, or adjacent to the proper parameter entry field; and
(b) Update warnings at any time during operation, not just during checkParameters().

These facilities are also used to implement a proper warning in ModFill for Mod Dimension being too large. The location of errors and warnings are improved throughout the visualizers, and in the Formula sequence.

Resolves #112.
Resolves #113.


gwhitney:
Looks like I need to update the Github CI test snapshots, so marking this draft until I do. I will also check in instructions for how to update those snapshots, into doc/code-tests.md

gwhitney:
OK, detailed instructions as well as new snapshots in place. Ready for review.

katestange:
Bug:  Try this url: http://localhost:5173/?name=ModFill&viz=ModFill&modDimension=1000&seq=Formula&formula=n%21 and then try to change the mod dimension in modfill.  It reinitializes but the mod dimension drawn doesn't change.  However the parameter box and url changes.  A hard browser reset or the soft header bar reset will result in the modfill being actually changed in the visualizer as expected.

katestange:
Review question:  Is it ok if I just trust you on the CLI snapshot instructions until next time I'm doing a visualizer overhaul and have to do that myself (which will definitely happen before alpha)?

gwhitney:
> Review question: Is it ok if I just trust you on the CLI snapshot instructions until next time I'm doing a visualizer overhaul and have to do that myself (which will definitely happen before alpha)?

Well, I wrote down exactly what I did, so it definitely works for me, so sure I don't have any issue putting off verifying it works the same way for someone else.

[looking into the other bug you found, thanks]

gwhitney:
> It reinitializes but the mod dimension drawn doesn't change. However the parameter box and url changes.

Good catch. I got overly ambitious with how much stuff I put in presketch -- I got confused about how presketch() doesn't re-run on typical parameter changes but setup() does, which is on purpose so you can control what gets reset and what doesn't. The framework is still working as expected, I just arranged the initializations wrong when I refactored ModFill. Should be working now.

katestange:
> These facilities are also used to implement a proper warning in ModFill for Mod Dimension being too large.

I get a warning like "Running with maximum modulus 14176, since 600000 will not fit on screen."  But when I resize, the number 14176 doesn't change.  Only a reset (browser or header button) after resizing will cause the error number to change.  Also, it seems to cap out at 14176 and even when I resize larger & reload it doesn't update this number.  I thought it would be the canvas width?

katestange:
> Good catch. I got overly ambitious with how much stuff I put in presketch

Works now, thanks.

katestange:
> Also, it seems to cap out at 14176 and even when I resize larger & reload it doesn't update this number. I thought it would be the canvas width?

Oh I think I see why this happens.  The canvas wants the line of slope one to go corner-to-corner so the boxes are actually rectangles.  After a certain point in my resize, my height has reached max, and this artificially caps the width and just stretches the boxes.

gwhitney:
> But when I resize, the number 14176 doesn't change. Only a reset (browser or header button) after resizing will cause the error number to change.

Yup, you've found another problem. Looking into it.

doc/code-tests.md
```diff
@@ -139,6 +139,66 @@ command `npm run test:e2e -- --update-snapshots`. This will produce new
 expected snapshots in the directory /e2e/tests, which you will need to add as
 part of your commit.
 
+### Updating snapshots for the Github Continuous Integration (CI) testing
+
+Sometimes the new behavior will also affect the snapshots for tests that are
+run automatically when pull requests (PR)s are created or updated, or new code
+is merged into main. (Note there is a separate set of snapshots for such
+tests, because the images for some of the tests end up slightly different in
+the Github CI runtime environment than on our local machines.)
```
katestange, line 148: It's not clear from this whether this applies to all snapshots or just those that involve the parameter panes, or....?  Because I don't recall having this issue with Turtle.ts.  Can you explain a little more?  And maybe say how to tell whether this is why the github tests aren't passing?  (Is it just that things pass on your machine but not on github?)

gwhitney:
OK, here's my take on this: presketch() explicitly takes a size argument that gives what the size of the canvas is going to be. That implies to me that if the size was going to be something different, then the framework would be obliged to call presketch again, because maybe I made some decision based on the size, that I would then need to adjust for a different size. (In fact, that's exactly what was biting ModFill.) But the framework was not necessarily calling presketch again when the size changed; sometimes it would, sometimes it wouldn't. So I changed the code to make sure that the framework calls presketch() again when the size changes. If that all seems appropriate to you, we should be good to go for further review. Your review so far has been super productive, thanks!

src/sequences/Formula.ts
```diff
@@ -2,15 +2,40 @@ import {Cached} from './Cached'
...
         required: true,
+        validate(f: string, status: ValidationStatus) {
+            let parsetree = undefined
+            try {
+                parsetree = math.parse(f)
+            } catch (err: unknown) {
+                status.addError(
+                    `Could not parse formula: ${f}`,
+                    (err as Error).message
+                )
```
katestange, line 23: I'm thinking we should also catch an empty box with "Could not parse formula".  Right now backspacing to empty the box and start afresh (something that will happen plenty) causes an interrupting popup with an unhelpful message (pic attached).  So can we watch out for an empty box before math.js (or whoever is throwing this message) gets involved?
Screenshot from 2024-10-26 16-35-44

katestange: Also weird side-effect of this popup is that the numberscope gets in a sensitive state and pops this up more often, even after you've cleared it.  For example, changing n to n+log(2) is fine with a fresh scope, but after this has happened, that same change will cause a pop-up on the final parenthesis.

src/visualizers/Chaos.ts
```diff
@@ -127,6 +147,11 @@ const paramDesc = {
             'Alpha factor (from 0.0=transparent to 1.0=solid) of the dots.',
         visibleDependency: 'dummyDotControl',
         visibleValue: true,
+        validate(a: number, status: ValidationStatus) {
+            if (a < 0 || a > 1) {
+                status.addError('must be between 0 and 1, inclusive.')
```
katestange, line 152: above you didn't use a period

src/visualizers/Chaos.ts
```diff
@@ -171,41 +199,18 @@ class Chaos extends P5Visualizer(paramDesc) {
...
-                'The number of walkers must be a positive integer.'
+                'The highlighted walker must be less than '
+                    + 'the number of walkers.'
             )
-        }
-        if (params.gradientLength < 1) {
-            status.addError(
-                'The colour cycle length must be a positive integer.'
+            this.statusOf.highlightWalker.addWarning(
+                'must be less than the number of walkers'
```
katestange, line 208: these two also mis-match on whether period is included... would be nice to be consistent, whichever way we go

src/visualizers/NumberGlyph.ts
```diff
@@ -160,6 +160,9 @@ class NumberGlyph extends P5Visualizer(paramDesc) {
         try {
             parsetree = math.parse(params.formula)
         } catch (err: unknown) {
+            // TODO in overhaul: make a utility somewhere that both this
+            // SequenceFormula.ts can use; also move this into the individual
+            // param validator for the 'formula' param
```
katestange, line 165: "both this and"?

src/visualizers/ShiftCompare.ts
```diff
@@ -31,6 +32,9 @@ const paramDesc = {
         displayName: 'Modulo',
         required: true,
         description: 'Modulus used to compare sequence elements',
+        validate(m: number, status: ValidationStatus) {
+            if (m <= 0n) status.addError('Modulo must be positive')
```
katestange, line 36: "Modulus"

src/visualizers/Turtle.ts
```diff
@@ -361,9 +359,9 @@ class Turtle extends P5GLVisualizer(paramDesc) {
                 }
             }
             if (nIn < threshold) {
-                status.addWarning(
-                    `Only ${nIn} of the first ${toTest} entries are in the `
-                        + `domain (${params.domain}); consider adjusting.`
+                this.statusOf.domain.addWarning(
+                    `Only ${nIn} of the first ${toTest} sequence entries `
+                        + 'are in this domain; consider adjusting.'
```
katestange, line 364: Above, "Domain" was capitalized a few times; here it is not.

katestange (changes requested):

katestange:
> If that all seems appropriate to you, we should be good to go for further review. Your review so far has been super productive, thanks!

That sounds good to me.  Should it be clarified in the "make a visualizer" doc?

src/visualizers/NumberGlyph.ts
gwhitney: Well, given that you've also made a very reasonable request to adjust the Formula validation, rather than make that change in both places I am actually going to go ahead and do this "TODO". There are two ways to go about it.

1. Add a `ParamType.FORMULA` in src/shared/ParamType.ts, or
2. Add a convenience function `validateFormula` in src/sequences/Formula.ts that can be separately imported other places that want to have `STRING` parameters that really should be formulas.

Thoughts/preferences on which way to go? I guess it depends on how common we expect it will be that people will want formula parameters and whether the additional ParamType will be more complicated/confusing/entry-barrier-creating than useful.

doc/code-tests.md
gwhitney: I will expand this portion of the doc, but you did not have this issue with Turtle because GitHub can't run WebGL tests :-( -- that's why PR review must always include running tests on the reviewer's local machine, not relying just on Github automatic test runs, sadly.

src/visualizers/NumberGlyph.ts
katestange: Hmmmm, I think a ParamType.FORMULA seems useful (we've already got one place we need this).  I could imagine using it for a colour range on Turtle path for example.... and for the shiftcompare, you could have any formula for whether to turn a pixel on... I can think of various uses!

gwhitney: OK, working on it as a ParamType.

src/visualizers/Chaos.ts
gwhitney: OK, I tried to make all of the categories of text related to parameters and validations each consistent:
* parameter description fields should now all be complete sentences, with capital letters and periods.
* individual-field validation messages should now all be uncapitalized phrases without periods.
* object-wide messages that appear at the top of the parameter tab rather than by a specific field should now all be complete sentence, w/ caps and periods.

Do those seem like reasonable conventions? Should we put those conventions in the documentation somewhere? If so, do you have a suggestion for where?

src/visualizers/NumberGlyph.ts
gwhitney: OK, now there is a ParamType.FORMULA. Note that if the inputs are anything but the usual `n`, they must be specified by an `inputs` property; see NumberGlyph `growthFormula` parameter for an example.

gwhitney:
OK, I believe I have made all of the requested changes. I expect that the CI tests will fail because the NumberGlyph snapshot has changed. I will make another commit when the CI tests finish, if so. But it should be fine to continue your review any time.

doc/code-tests.md
```diff
@@ -139,6 +139,74 @@ command `npm run test:e2e -- --update-snapshots`. This will produce new
...
+Github CI runtime environment than on our local machines.
+
+If you're not sure whether you might be in this situation, you can tell after
+you create a PR: if `npm run test:e2e` succeeds on your machine, but the
+GitHub CI tests fail, and when you look at the details of the tests (see just
+below for the procedure to find those details), the failures are caused by
+snapshot mismatches for snapshots with file names that include `ci_snaps` in
+the full path to the "expected" image.
+
+To go ahead and update thse CI snapshots, the first challenge is to obtain the
```
katestange, line 157: these (typo)

src/visualizers/Chaos.ts
katestange: Those seems reasonable to me.

katestange: For documentaion, maybe in the example visualizer code?

src/visualizers/NumberGlyph.ts
katestange: Nice!

gwhitney:
OK, how's that?

katestange (approved):

Issue #475, opened 2024-10-26
Handling of large formula numbers in factorfence


katestange:
This URL: 
http://localhost:5173/?name=Formula&viz=FactorFence&highlight=17&seq=Formula&formula=2%5En-1&first=1
Produces this erroneous image (factorizations are wrong from the 54th term onward):
![Screenshot from 2024-10-26 14-35-57](https://github.com/user-attachments/assets/f85ee79a-8c19-4673-892a-7f018bd86139)
The problem appears to be

katestange:
Interestingly it is fine on n factorial?  It greys out the bars in that case, as it should:
![Screenshot from 2024-10-26 14-44-47](https://github.com/user-attachments/assets/8e66be04-530a-45a1-b473-c136753e2c40)

katestange:
Also, this may or may not be related, but when you click "sequence" from the n factorial view above, you see Turtle doing something bad:
![Screenshot from 2024-10-26 15-01-05](https://github.com/user-attachments/assets/024eded7-c0f5-4ec9-b3c7-0a326e7e3ef1)
However, if you click turtle it comes up okay.

gwhitney:
Yes, definitely a bug. And a weird one, too, as the in-browser factorizer is supposed to know its own limitations and get this sorted out. Marking for alpha.

katestange:
This may be related and might provide clues:

http://localhost:5173/?name=FactorFence&viz=FactorFence&seq=Formula&formula=gcd%285%5En-1%2C3%5En-1%29

This just doesn't look right.  There seem to be three phases going on.  On the left, it looks correct to me, varying height, varying factorizations.  In the middle, we see a bunch of probably fake powers of two (e.g. a(44) _should_ be 24656 = 2^4 * 23 * 67).  At the right, we see reported failures to factor, but they seem strangely large (bars too high and uniform), seems wrong (e.g. we _should_ get a(71) = 2 and a(72) = 149440928 = 2^5 * 7 * 13 * 19 * 37 * 73 ).

Sage can be used to verify values, e.g. code below put into https://sagecell.sagemath.org/
```
n = 32
g = gcd(5^n-1,3^n-1)
print(g, factor(g))
```

![Screenshot from 2024-10-26 19-51-11](https://github.com/user-attachments/assets/3427e9c8-8536-4dcc-b919-7330a3154311)

gwhitney:
[...]

PR #485, opened 2024-10-31
fix: Don't attempt to factor 'unsafe' integers

gwhitney:

<hr/>

Resolves those portions of #475 that can be dealt with before upgrading to mathjs 13 and bigint within Formulas.
That issue #475 will need to be revised and adjusted to beta milestone once this PR is merged.

Also adds at least rudimentary tests for simpleFactor, including one that won't pass until we make it less "simple".


katestange:
New tests pass, factor fence behaving as expected, at least for the first two screenshots in #475.  Note that the gcd screenshot in that issue is still not good, but I think you aren't claiming this fixes that.

PR #497, opened 2024-11-15
ui: OEIS search fixes and improvements.

### LINKED ISSUE #493: Wrong icon for OEIS search bar help.
The main frontscope app page clearly establishes the convention that the black-circle-with-white-question-mark icon means a tooltip-on-hover (it's in the specimen bar in the name field and on any parameter with hideDescription: true and works this way. That's exactly the behavior we have in the OEIS search bar in the sequence switcher, so it should use the same icon, not a new one (magnifying-glass-in-rounded-square).

### LINKED ISSUE #494: Further enhancements to OEIS search and IDs displayed
These are several enhancements related to the handling and display of OEIS sequences and IDs that came up in the discussions for #486 but which are not properly part of the documentation site overhaul. We do want to implement them for alpha, however:

- [x] In the OEIS search results, clicking on either the sequence name or number should add the sequence; only clicking on the external link icon should take you to the external OEIS page.
- [x] All occurrences of OEIS IDs in the user interface (e.g., in the name field of the Sequence tab) should have external link icons that go to the OEIS page for that sequence.
- [x] When a new sequence is added to the Sequence Switcher by virtue of having selected it in the search results, the specimens gallery should scroll to that new sequence, and it should be temporarily highlighted with the primary numberscope color.

gwhitney:

<hr/>

* Uses the ?-in-circle help icon for the help popup on the OEIS search bar, not the magnifying-glass-in-square.
* Entire row of OEIS search results adds the sequence (not just the description); only the "external link" icon is the external link to the OEIS.

Resolves #493.
Resolves #494.


gwhitney:
This is marked draft because I plan for it to fully resolve #494, so it is a work in progress. However, I am sharing it now because I am unsure about the design of the OEIS search results table. I have made each entire row of the table add the sequence, so that it won't matter where you click; if your urge is to click on the ID, so be it, that will still add (as discussed). Correspondingly, the entire  row turns pale blue when you hover it.

But there still needs to be a way to get to the OEIS web page for that sequence. So I have made just the "external link icon" (itself and only the icon itself) be the external link. Like other active controls, I have rendered it in dark grey that turns black when you hover. But I found, to my eye, that decoration still too subtle to realize that it's a link or at least some kind of control that does something. So now I have also made it enlarge slightly when you mouse over it. Is that now tacky/overemphasized? 

What about the location of the external link icon? Is there some other better way to render the link to the OEIS? I think the semantics of the icon are pretty clear, just I'm worried about it being noticed, especially as I think it's not standard for only the icon to be a link, I think usually it's _part_ of a link (as it was before this PR) and just indicates that link leaves the current website.

If the link and how it looks/responds to mouseover is actually OK, should that look/feel just be replicated to all occurrences of OEIS IDs throughout the UI? In other words, should we insert the external link icon after each one, make just the icon a link, make it change darkness and get bigger on hover?  If so, should we do it just for the sequence ID in an OEIS sequence name in the sequence parameter tab and in specimen cards in the Sequence Switcher, or should we literally search all displayed text (e.g. if the description for one sequence happens to mention another sequence) and insert links for every one of them?

I can work on adding a sequence scrolling to that specimen card and temporarily highlighting it in the meantime, but then to make further progress here these questions need to be settled, at least to their "alpha version answers". Thanks!

gwhitney:
OK, there's the effect of scrolling to a new sequence, highlighting it, and then letting it fade. How's that look?
Now I don't think I can go any further with this PR until we settle on how exactly external links to the OEIS are going to look/work. Let me know what you think of the mockup that is currently in this PR in the OEIS search results table. Thanks!

katestange (approved):
Ok I can't break it by playing with it and I love the fade effect and the way the links are now is clearly an improvement.  Merging.

katestange:
Oh except this is a draft.  maybe you wanted to do more?

katestange:
> Correct. The issue says to make all oeis ids link in the same way, wherever they may appear. The question is how far to take that: just the ones we deliberately display (like in the name of a sequence, relatively straightforward) or any that might come up in any text (like happen to be mentioned in the description for another sequence, say -- this sort of scan would be more intricate). Let me know your thoughts on that.

Hmm, I don't think the more intricate route is necessary.  I'd mainly like to have the links on the sequence cards and from the sequence panel.

gwhitney:
All right, all of the OEIS links are similar now: in search results, sequence switcher, and sequence param editor. Marking as ready for review.

katestange:
Ok this seems to do everything you describe and it seems to run on different window sizes, scrolling appropriately, and the links work reliably and look nice in several browsers.  I have no suggestions!  I'm going to merge!

PR #498, opened 2024-11-16
fix[Formula]: reject unknown function at parse to avoid throwing error

### LINKED ISSUE #427: Some formula entry impractical in ui2
Right now when entering a formula, if you want to change from `permutations(n,3)` to `combinations(n,3)` (say), and you do this by deleting `permut` character-by-character and then typing `combin`, you get an error dialog that e.g. `permations` is an undefined function for every single keystroke that you have to respond to by clicking on an 'OK' button. It is frustrating and makes formulas essentially unusable. We need to make the validation of formulas "softer" in some way for the instant update to be usable.

_Originally posted by @gwhitney in https://github.com/numberscope/admin/discussions/6#discussioncomment-9899148_

gwhitney:

<hr/>

Resolves #427.

### DIFF
--- a/src/sequences/Formula.ts
+++ b/src/sequences/Formula.ts
@@ -101,18 +101,46 @@ Plus the standard parameters for all formulas:
 !}
 **/
 
+const maxWarns = 3
 class Formula extends Cached(paramDesc) {
     static category = 'Formula'
     name = `${Formula.category}: ${paramDesc.formula.default}`
     static description = 'A sequence defined by a formula in n'
+    nErrors = 0
 
     initialize(): void {
         super.initialize()
         this.name = 'Formula: ' + this.formula.source
+        this.nErrors = 0
     }
 
     calculate(n: bigint) {
-        const result = this.formula.compute(Number(n))
+        let result = 0
+        try {
+            result = this.formula.compute(Number(n))
+        } catch (err: unknown) {
+            this.nErrors++
+            if (this.nErrors < maxWarns) {
+                const message =
+                    err instanceof Error ? err.message : 'of unkown error.'
+                this.statusOf.formula.addWarning(
+                    `value for n=${n} set to ${result} because ${message}`
+                )
+            } else if (this.nErrors === maxWarns) {
+                this.statusOf.formula.addWarning(
+                    `[1 additional warning discarded]`
+                )
+            } else {
+                // replace discarded message
+                const ensure = this.statusOf.formula.warnings.pop()
+                console.assert(ensure && ensure.endsWith('discarded]'))
+                this.statusOf.formula.addWarning(
+                    `[${this.nErrors - maxWarns + 1} additional `
+                        + 'warnings discarded]'
+                )
+            }
+            return BigInt(result)
+        }
         if (result === Infinity) return BigInt(Number.MAX_SAFE_INTEGER)
         else if (result === -Infinity) return BigInt(Number.MIN_SAFE_INTEGER)
         else if (Number.isNaN(result)) return BigInt(0)
--- a/src/sequences/Formula.ts
+++ b/src/sequences/Formula.ts
@@ -2,7 +2,7 @@ import {Cached} from './Cached'
 import {SequenceExportModule} from './SequenceInterface'
 
 import {math, MathFormula} from '@/shared/math'
-import type {GenericParamDescription} from '@/shared/Paramable'
+import type {ParamValues, GenericParamDescription} from '@/shared/Paramable'
 import {ParamType} from '@/shared/ParamType'
 
 /** md
@@ -112,6 +112,11 @@ class Formula extends Cached(paramDesc) {
         super.initialize()
         this.name = 'Formula: ' + this.formula.source
         this.nErrors = 0
+    }
+
+    assignParameters(realized?: ParamValues<GenericParamDescription>) {
+        this.nErrors = 0
+        return super.assignParameters(realized)
     }
 
     calculate(n: bigint) {
--- a/src/sequences/Formula.ts
+++ b/src/sequences/Formula.ts
@@ -116,8 +116,10 @@ class Formula extends Cached(paramDesc) {
 
     calculate(n: bigint) {
         let result = 0
+        let resultType = ''
         try {
             const rawResult = this.formula.compute(Number(n))
+            resultType = typeof rawResult
             result = math.floor(rawResult)
             if (isNaN(result)) {
                 result = 0
@@ -129,8 +131,14 @@ class Formula extends Cached(paramDesc) {
         } catch (err: unknown) {
             this.nErrors++
             if (this.nErrors < maxWarns) {
-                const message =
+                let message =
                     err instanceof Error ? err.message : 'of unkown error.'
+                if (resultType && message.includes('convert')) {
+                    message = message.replace(
+                        'convert',
+                        `convert ${resultType} value`
+                    )
+                }
                 this.statusOf.formula.addWarning(
                     `value for n=${n} set to ${result} because ${message}`
                 )
--- a/src/sequences/Formula.ts
+++ b/src/sequences/Formula.ts
@@ -1,7 +1,7 @@
 import {Cached} from './Cached'
 import {SequenceExportModule} from './SequenceInterface'
 
-import {MathFormula} from '@/shared/math'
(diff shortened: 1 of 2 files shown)


gwhitney:
Sorry about the stacked PRs, but this one is straightforward and orthogonal to #497. Feel free to respond to/review either in whatever order is convenient. Whichever gets merged first, rebasing the other will be trivial.

katestange:
This behaves as described and fixes that problem. But I got a popup that permutations requires second argument to be less than first argument. There are still going to be tons of popups as we use this. Is it possible to route all popups that derive from mathjs to a red error message like this PR does for unknown function?

gwhitney:
Well the halting problem says we can't detect all error-throwing at parse time. So the best I can do is put a catch on the computation and post the error message; it may be murky though. I will implement something and you can give it a try.

katestange:
> Well the halting problem says we can't detect all error-throwing at parse time. So the best I can do is put a catch on the computation and post the error message; it may be murky though. I will implement something and you can give it a try.

Yeah, I guess I was hoping just that pop-ups (that require a click) would show up as red messages instead (no click), if we can tell they are coming from the user typing in the sequence box.  I don't know how much can be done in this direction, but if there's a way to tell they are because of typing in the box, we could just re-route from pop-up to red error.  We might occasionally get weird error messages there I suppose.... I look forward to seeing what you try.

gwhitney:
OK, give that a try. Let me know what you think.

katestange:
So far I really like this.  I haven't looked at the code yet, because I wanted to just mess around and discover what it does.  I haven't been able to prompt a pop-up even doing weird stuff like making matrices and whatnot.  And the limit (1022 additional warnings discarded) is really great! :)

katestange:
Ok I did manage to provoke some pop-ups.  The main thing is the following, which happens when I use functions that don't ouput the right type to interpret as a Bigint for a sequence.  A simple way to provoke this is to type `typeOf(n)` in the formula box.
Screenshot from 2024-11-16 16-55-22

gwhitney:
OK, now it checks for NaN inside the catch block, so it is even more robust. What do you think now?

katestange:
> OK, now it checks for NaN inside the catch block, so it is even more robust. What do you think now?

Why does it give this seemingly nonsensical error, though?
Screenshot from 2024-11-16 17-32-16

gwhitney:
Umm because the value returned by the formula `typeOf(n)` is the _string_ `"number"`, which indeed cannot be converted to a number. Would you like me to include the name of the data type in the error message?

katestange:
> Umm because the value returned by the formula `typeOf(n)` is the _string_ `"number"`, which indeed cannot be converted to a number. Would you like me to include the name of the data type in the error message?

Oh my gosh that totally confused me! :)  Yeah, if you are able to easily include "of type ..." that might help someone clueless like me. :)

gwhitney:
How's that read?

src/sequences/Formula.ts
```diff
@@ -101,22 +101,65 @@ Plus the standard parameters for all formulas:
...
+                    `[1 additional warning discarded]`
+                )
+            } else {
+                // replace discarded message
+                const ensure = this.statusOf.formula.warnings.pop()
+                console.assert(ensure && ensure.endsWith('discarded]'))
+                this.statusOf.formula.addWarning(
+                    `[${this.nErrors - maxWarns + 1} additional `
+                        + 'warnings discarded]'
+                )
```
katestange, line 156: I'm not sure why, but when I tested this by beginning with `typeOf(n)` it gave the expected error, then I expanded this to `typeOf(matrix([n,n]))` and the following happened:  the warnings were all hidden (but apparently present...).  Bug?
Screenshot from 2024-11-16 21-19-14

src/sequences/Formula.ts
```diff
@@ -101,22 +101,57 @@ Plus the standard parameters for all formulas:
 !}
 **/
 
+const maxWarns = 3
```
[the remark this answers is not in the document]
gwhitney: Well, I am counting 2 full warnings plus the warning that additional warnings were discarded makes 3 warnings, exactly as maxWarns prescribed. Does that not seem right? In other words, I can only put up maxWarns-1 "real" warnings to leave room for the "discarded warnings" warning...

src/sequences/Formula.ts
gwhitney: Hmm yes seems like nErrors is not getting reset even though the warnings list was apparently cleared. Will investigate.

gwhitney: Hmmm can't reproduce. I used formula `typeOf(n)` and got the expected warnings. I tried inserting my cursor after the opening paren and typing `matrix([`, then skipping over the `n`, then typing the `])`, but that left the warnings looking totally fine. I also tried having `matrix([n,n])` in my clipboard and pasting it over the `n`, but again all looked well. So I think I need the exact URL you started with, and the exact sequence of keystrokes/pastes/whatever to get it into the shown state to debug.

katestange: Go to http://localhost:5173/?name=Formula&viz=ModFill&modDimension=1000&seq=Formula&formula=typeOf%28matrix%28%5Bn%2Cn%5D%29%29 and then backspace the [ and replace it.

src/sequences/Formula.ts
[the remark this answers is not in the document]
katestange: Oh I see, just confused me.

src/sequences/Formula.ts
gwhitney: Indeed, I can now reproduce it. Will investigate.

gwhitney: OK, that was a subtle one. How about now?

katestange: Now the URL I sent you shows too many warnings.
Screenshot from 2024-11-17 00-03-02

gwhitney: Hunh. It only happens on page load/reload. If I wipe out some totally unrelated formula and then type in this one, it's totally fine. Ugh. Will have to dive deeper tomorrow.

katestange: Something to do with Vue?

gwhitney: OK, gave up on trying to sync a property of the Formula class and the warnings on the list. Now I just use the warnings themselves to keep track and all seems well. What do you think?

katestange: Looks like it's working.  Question -- what controls how many n's it tries (and therefore how many warnings are suppressed).  For example when it says "n=1280 set to 0 because Cannot convert string value "DenseMatrix" to a number" why 1280?

katestange: In fact, weird behaviour, this n=whatever increments when I do the erase + replace left square bracket (as I described for the original bug).  Probably doesn't matter...

katestange: well, I'm impressed, it let me type this: det(matrix([[n,n],[n^2,n-1]]))*n^n^n^n^n^n^n^n^n^n^n^n^n^n^n^n*2*3*4*5*6+combinations(n,2)

katestange: sorry, git ate some of the symbols

gwhitney: > weird behaviour, this n=whatever increments when I do the erase + replace left square bracket (as I described for the original bug). Probably doesn't matter...

Well, actually it is not so weird. You're mucked around with the formula, but gone back to what the formula was before. So frontscope is clever enough to realize that this is the _same_ Sequence that it was before, but that mucking around it interprets as a request for more values, so it caches some more values of that same Sequence. That's why the _n_ starts much larger this time, it's adding more values to the end of what's already cached. Make sense?

gwhitney: > it let me type this

Did it actually compute any values successfully? I guess n=1 is ok but it seems like that expression should overflow at n=2...

gwhitney: Let me know if there's anything else I need to do on this.

katestange: > Did it actually compute any values successfully? I guess n=1 is ok but it seems like that expression should overflow at n=2...

Well that's hard to know, it did output something that modfill was willing to plot.  Maybe we should file an issue to investigate whether overflows are silently giving garbage.

Issue #431, opened 2024-08-28
Inability to dock one tab when other is large.


gwhitney:
Regarding docking.  If you expand one pane at the side to be big, then when you drag the other pane over to the same side, there is no "blue drop zone" and you can't drop to dock.  Instead if you drop, you end up floating over the docked pane.  Do we want to allow dragging to dock in this situation, forcing resize of the large pane?

_Originally posted by @katestange in https://github.com/numberscope/admin/discussions/6#discussioncomment-9970590_

gwhitney:
Planned per Oct 29 meeting: 
* If you drag a tab so that your mouse pointer is over the sidebar, then it (always!) docks. If one position is occupied, it goes into the other position, even if it has to "warp" there. If neither position is occupied,it goes in top or bottom depending on where your mouse pointer ends up. (Make sure the mouse pointer glyph does not change while the tab is grabbed.) And in this case, it does not resize. 
* If it is going into the "cavity" from one already there, then:
    * If the sum of the two sizes is less than the space available, leave the previously present one alone in size, and make the newly inserted one taller to fill up the space.
    * If the sum of the two sizes is greater than the space available, divvy up the whole space proportionally to the sizes before the drag.

gwhitney:
Implemented in ui2 via #499, closing.

PR #499, opened 2024-11-17
fix: Drag into sidebar always docks into an empty zone

gwhitney:

<hr/>

Prior to this PR, dragging a tab on top of a docked one would leave
  it floating over the docked tab. Moreover, if the empty zone in the
  sidebar was less than half the size of the tab being dragged, it would
  be impossible to dock the dragged tab at all.

  Now any drag into a sidebar with at least one empty zone results in
  a docking to the targeted zone if it is empty, or the first empty zone
  if not.

  Resolves https://github.com/numberscope/frontscope/issues/431.

  Note that this PR does not at the moment implement any resizing.
  There is resizing described in issue https://github.com/numberscope/frontscope/issues/431. However, now that this is
  working smoothly, it is making perfect visual sense to me that the
  dragged tab fits itself into the remaining empty space, leaving the
  already-docked tab alone, and in fact, this behavior seems to me
  preferable to what's described in https://github.com/numberscope/frontscope/issues/431. But if there remains a clear
  desire for the resizing as described in the issue, that can be
  implemented. Just let me know.

OK, this has been rebased and is ready for its own review. There are two spurious commits with changes to Formula.ts, a637a7f and 46d1ac9. These are an artifact of how I originally submitted this on top of #498. But the latter one precisely undoes the other, and so they will "squash commit" to nothing: the only commit that matters at this point is a53d7ea, and as you can see at the top, GitHub realizes that only Scope.vue has changed. Sorry about the confusion, I should have just done this as an independent PR like I did with the last two, but I didn't set Git up quite right this time. But now all should be well for this to be reviewed normally and once it is ready, squash-merged like normal.

### DIFF
--- a/src/components/Tab.vue
+++ b/src/components/Tab.vue
@@ -25,6 +25,7 @@
     import interact from 'interactjs'
     import type {InteractEvent} from '@interactjs/types'
     import {
+        maybeClearDropzoneContainer,
         positionAndSizeTab,
         positionAndSizeAllTabs,
         selectTab,
@@ -258,8 +259,7 @@
             )
             const dropzoneContainer = dropzone?.parentElement?.parentElement
             if (
-                tab instanceof HTMLElement
-                && dropzone instanceof HTMLElement
+                dropzone instanceof HTMLElement
                 && dropzoneContainer instanceof HTMLElement
                 && tab.classList.contains('docked')
             ) {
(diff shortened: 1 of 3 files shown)


gwhitney:
Aaron found a bug: pull out bottom tab, drag top tab down, try to drag that tab back up, it won't go, and then try to drag the old bottom tab back into the gap at the top, it just lands over the sidebar, it doesn't dock.
(Maybe the top zone did not get updated to empty when the tab dragged down.)

gwhitney:
Aaron is fine with only the dragged tab resizing, so unless Kate disagrees we will just go with the current behavior (once the bug is fixed).

gwhitney:
Another bug: Drag tab into left, resize it in left, drag it out so nothing is in left. When you put something in left, it now uses previous divider position.

gwhitney:
OK, I believe that commit fixes all of the issues we observed at today's meeting. Ready for further review.

katestange:
Ok, I can't get it to do anything that seems like a bug.  I'm going to merge.

Issue #386, opened 2024-07-10
ModFill should have mouseover


katestange:
Modfill should have mouseover so you can find out what each square is.

Issue #309, opened 2024-04-23
Mod Fill should have legend


gwhitney:
Since the diagram is square, at least in the sense of having the same number of rows as columns, there should be an indication on screen of which direction is the modulus and which direction is the residue. (Currently, each column is a modulus and each row a residue.)

gwhitney:
This should be considered but be optional for the overhaul of ModFill for alpha, so not setting a milestone.

gwhitney:
Did not get done in the overhaul for alpha, so setting milestone to beta.

gwhitney:
You could also possibly display the number of "hits" on the pixel.

Issue #143, opened 2022-07-20
Add color schemes to Mod Fill visualizer


gwhitney:
Currently, Mod Fill simply colors a cell black if anywhere in the series there is a representative of the corresponding congruence class. You can watch the diagram fill in to get some sense of when in the sequence a given residue occurs, but there is no indication in the final diagram.

This binary scheme is reasonable and should be kept, but I don't think it would be overcomplicating the visualizer to offer a couple more color schemes. The ones that come to mind (a subset or superset of these could be implemented) involve the user choosing a gradient (from some standard options like greyscale or typical heatmap gradients) and then using it to indicate one of
1) The position of the first term where that residue occurs
2) The position of the latest term visited where that residue occurs
3) The absolute number of times that residue occurs (not sure if this one makes sense)
4) The fraction of the visited entries in which that residue occurs

In (4) there is the question of whether the gradient should have the same dynamic range for different moduli -- for higher moduli, we might expect that all individual residues have low(er) frequencies, since there are so many residues, so we might want to (automatically, or under user control?) "turn the gain up" on the larger moduli.

This proposal is after an idea floated in #49.

gwhitney:
Cf #210.

gwhitney:
These things should be considered but optional for the ModFill overhaul for alpha.

PR #505, opened 2024-11-24
feat: add colours to ModFill

katestange:

<hr/>

This is still marked draft but feedback is welcome. It adds several colouring abilities to ModFill:

1.  options to choose colour and alpha (transparency) for drawing, transparency possibly as function of modulus
2.  option to highlight certain terms by using a second colour; a formula box allows you to pick which terms to highlight by a formula on the index, e.g. isPrime(n), and relevant controls
3. removes boring ModFill gallery items and adds nice ones (in flux again now)
4. updates documentation (still needs images)
5. adds background colour option as well as aspect ratio option

Things to do:

- [x] update snapshots
- [x] update images in docs
- [x] there's a background colour option that should work once the sequence cache bug is fixed
- [x] update gallery
- [x] consider other colour options
- [x] consider making some parameters appear when highlighting formula is not default


katestange:
Ok sorry for the rebase mess, this one should be clean but I see now something is wrong -- presumably something to do with my new formula parameter and your formula changes.  I'll take a look at that when I get a chance.

gwhitney:
OK, just let me know when you're ready for me to look.

katestange:
Ok, the merge issue is fixed, it should run and be ready except for the snapshots.

gwhitney:
First some general high-level feedback:

1. Given that you are diving into color schemes, have you reviewed #143? This PR more or less implements scheme 3 there, and then it has the highlight color not mentioned there. Are any of the other color schemes worth implementing now? This scheme allows one to use a saturation gradient for the number of times that a residue has occurred; would you want to enable the visualizer to use a hue gradient instead?
2. I'm thinking you are trying to keep this focused, and not considering it the complete overhaul, so you would want to leave consideration of possibly doing #309 and/or #386 to another PR, correct?
3. In a similar vein, you want to leave the question of whether to always have ModFill be square, or to have a toggle that lets you flip between "square" and "fill the rectangle", to another PR? (I definitely think that square should at least be an option.)
4. I've never loved the description "A triangular grid". To me that calls to mind the tessellation of equilateral triangles. Can you think of better/clearer wording? Possibly just remove the word "triangular"?
5. Should the default for "Highlight Formula" be `false`, corresponding to the good ol' monochrome behavior? If we changed it to default to 0, the description could suggest `isPrime(n)` as an example.
6. When "Highlight Formula" is `false` or `0`, should the Highlight color be hidden, since it won't be used?
7. On Featured specimens, does ModFill warrant three? Is there any total number we're more or less going for, or average per visualizer? If you wanted to prune one, the 'Prime Residues' one seems easiest to discover on one's own, so maybe has the least need to be in featured specimens.  (And I know I am undermining my own point here but I sort of miss Residue Rise -- it made a really interesting 3D-looking texture after a while, which i am not sure the reason for. So maybe I am actually barking up the wrong tree, and we should be tending toward a whole lot of featured specimens, and I should just be campaigning for Residue Rise to be left in... What's your reaction on this point?)
8. Not sure why transparency is an integer from 1 to 255; wouldn't it be more typical to make it a decimal between 0 and 1?  In particular, in the Chaos visualizer, we call this same feature "Alpha" and have it be between 0 and 1. I think the "Transparency" label is clearer, and the 0 - 1 dynamic range is clearer, but in any case, I think most strongly of all that the two should be completely consistent with each other.

gwhitney:
One more general philosophical question here: When the transparency level is low, the bottom-left corner is always darker than the right-hand side, simply because there are fewer residues so each on average occursmore times. In many cases, this is visual noise, there's not really "more hits" in the smaller residues than the larger residues, in that each term does in fact have one residue per modulus; they're just spread out over more different residues in the higher moduli. Should there be an option, perhaps even an  option turned on by default, to allow the transparency to depend on n so that you can adjust things so that "all else being equal", the diagram would be uniformly saturated left to right?

gwhitney:
More small-bore comments (as with most of my feedback, more questions than directives):

1. Would 'Highest modulus to display' be clearer as the label instead of just 'Highest modulus'?
2. I think the term "hit" is unclear in "Transparency of each hit". Is there better wording? Is "Transparency of each residue plotted" any better? "Transparency of each occurrence of a residue"?
3. The capitalization in "Highlight Formula" is inconsistent with other fields; "Highlight formula" would be more consistent. But would a label like "Indices to highlight" be clearer/more informative? You can explain it is a formula in the description (which you basically do already).
4. There's a space missing after 'of' in 'residue ofa(n)' in the description for this parameter.
5. Logically, the highlight transparency could be different from the usual one. Do you want to add a parameter for that? If there were an RGBA color picker, it would be nice to just use that for both...

gwhitney:
I searched around. There is no browser-native rgba color picker, but it appears that it would be pretty easy to create a RGBA param type and roll our own input setup  with a combination of a color picker and 0-1 slider. If you think that would be useful, I am happy to help. We could use the rgba color in Turtle, too, I suppose -- Wait For It would definitely look different with some transparency for the turtle path, for example.

src/visualizers/ModFill.ts
```diff
@@ -28,29 +39,99 @@ modulus to consider.
...
-        displayName: 'Mod dimension',
+        displayName: 'Highest modulus',
         required: true,
         validate: function (n: number, status: ValidationStatus) {
             if (n <= 0) status.addError('Must be positive.')
         },
     },
+    /** md
+- Alpha: The rate at which cells darken with repeated hits.  This
+should be set between 1 (very transparent) and 255 (solid).
```
gwhitney, line 52: This is called Transparency in the ui. The names should be the same in the ui and the docs.

gwhitney:
Code is all nice and straightforward. Once the design/wording stuff above is decided/resolved, we should be good to go.

katestange:
> I searched around. There is no browser-native rgba color picker, but it appears that it would be pretty easy to create a RGBA param type and roll our own input setup with a combination of a color picker and 0-1 slider. If you think that would be useful, I am happy to help. We could use the rgba color in Turtle, too, I suppose -- Wait For It would definitely look different with some transparency for the turtle path, for example.

I think this would be lovely!  Yes, Turtle would benefit too, and it would simplify parameter panels (although it hides powerful functionality in some cases, so we might want to add a hint or something).  Would the roll our own basically use the browser RGB plus a slider?  Do you imagine doing that as part of this PR or another one?

katestange:
I need help with a weird (race condition?) bug.  Try this URL:
http://localhost:5173/?name=draw+problems&viz=ModFill&modDimension=100&alpha=20&fillColor=a51d2d&highlightFormula=0&highColor=ff7800&seq=Formula&formula=floor%28n%2F1%29
You'll see (pic attached from firefox), that the bottom row(s) isn't/aren't always drawn at first.  What's happening, as far as I can tell, is that code is running and thinks it is running (console logs confirm this), but drawing is not happening on canvas, until some sort of delay time has passed.  For example, if you put a command to change background colour, it won't happen unless you program it to happen after some number of terms.  It seems (??) worse with floor(n/1) than with n (which should be equivalent) and worse (??) on firefox than chrome, but it is intermittent how many rectangles are missed.  I'm stumped.
draw problems
Screenshot from 2024-11-25 08-43-13

gwhitney:
> I need help with a weird (race condition?) bug.

For me it is only happening on page load. If I click your URL and then enter a "2" in the number of terms to use, I get both horizontal stripes. Similarly if I change any other parameter of the the visualizer i seem to be getting all of the horizontal stripes. Does that mesh with what is happening for you?

gwhitney:
Wow, and I agree, I can't get it to happen with formula `n` but it reliably does on page load with formula `n/1`. And it only seems to be happening on ModFill, in that with formula `n/1` for two terms I get two strokes in Turtle, two dots in Chaos, etc. Does that also match with your experience? So weird...

gwhitney:
OK, it's an async initialization mess with sequences. Roughly the problem is that Specimen.makeSequence returns right away, but it has initiated an asynchronous validate, which may eventually conclude that the parameters have changed, and that it needs to initialize the sequence, but that might end up happening after the sequence has filled its cache, and then the cache gets destroyed... Will try to untangle, but the sequence of events is really muddled. This is not a proud part of Numberscope's design :-(

gwhitney:
Not sure why it isn't happening to other visualizers but I think it is because of a quasi-bug in modfill that we should fix: it increments its index before draw has succeeded. Since eventually there will be a cache miss, that's not actually kosher. But it has been a useful quasibug, in that there shouldn't be a cache miss on the first call to compute

katestange:
>     1. Given that you are diving into color schemes, have you reviewed [Add color schemes to Mod Fill visualizer #143](https://github.com/numberscope/frontscope/issues/143)? This PR more or less implements scheme 3 there, and then it has the highlight color not mentioned there. Are any of the other color schemes worth implementing now? This scheme allows one to use a saturation gradient for the number of times that a residue has occurred; would you want to enable the visualizer to optionally use a hue gradient instead?

So far haven't tackled more of this.

>     2. I'm thinking you are trying to keep this focused, and not considering it the complete overhaul, so you would want to leave consideration of possibly doing [Mod Fill should have legend #309](https://github.com/numberscope/frontscope/issues/309) and/or [ModFill should have mouseover #386](https://github.com/numberscope/frontscope/issues/386) to another PR, correct?

Correct.

>     3. In a similar vein, you want to leave the question of whether to always have ModFill be square, or to have a toggle that lets you flip between "square" and "fill the rectangle", to another PR? (I definitely think that square should at least be an option.)

Done.

>     4. I've never loved the description "A triangular grid". To me that calls to mind the tessellation of equilateral triangles. Can you think of better/clearer wording? Possibly just remove the word "triangular"?

Tried to improve.

>     5. Should the default for "Highlight Formula" be `false`, corresponding to the good ol' monochrome behavior? If we changed it to default to 0, the description could suggest `isPrime(n)` as an example.

Done.

>     6. When "Highlight Formula" is `false` or `0`, should the Highlight color be hidden, since it won't be used?

Not sure how to do this -- currently the params system doesn't have something like that.  Should I consider adding it?  Seems useful.

>     7. On Featured specimens, does ModFill warrant three? Is there any total number we're more or less going for, or average per visualizer? If you wanted to prune one, the 'Prime Residues' one seems easiest to discover on one's own, so maybe has the least need to be in featured specimens.  (And I know I am undermining my own point here but I sort of miss Residue Rise -- it made a really interesting 3D-looking texture after a while, which i am not sure the reason for. So maybe I am actually barking up the wrong tree, and we should be tending toward a whole lot of featured specimens, and I should just be campaigning for Residue Rise to be left in... What's your reaction on this point?)

Not addressed right now (and this most recent commit doesn't change the gallery which is a bit broken by this commit.

>     8. Not sure why transparency is an integer from 1 to 255; wouldn't it be more typical to make it a decimal between 0 and 1?  In particular, in the Chaos visualizer, we call this same feature "Alpha" and have it be between 0 and 1. I think the "Transparency" label is clearer, and the 0 - 1 dynamic range is clearer, but in any case, I think most strongly of all that the two should be completely consistent with each other.

Done.

katestange:
> One more general philosophical question here: When the transparency level is low, the bottom-left corner is always darker than the right-hand side, simply because there are fewer residues so each on average occursmore times. In many cases, this is visual noise, there's not really "more hits" in the smaller residues than the larger residues, in that each term does in fact have one residue per modulus; they're just spread out over more different residues in the higher moduli. Should there be an option, perhaps even an option turned on by default, to allow the transparency to depend on n so that you can adjust things so that "all else being equal", the diagram would be uniformly saturated left to right?

I made transparency a value that could also be a function of m.  That way, someone can play with this.  But in my experiments, the way it is set by default is the most visually pleasing.  You can try a few combos, and if there are good ones we can add them to the docs.

src/visualizers/ModFill.ts
katestange: I fixed this throughout (I think)

katestange:
> Not sure why it isn't happening to other visualizers but I think it is because of a quasi-bug in modfill that we should fix: it increments its index before draw has succeeded. Since eventually there will be a cache miss, that's not actually kosher. But it has been a useful quasibug, in that there shouldn't be a cache miss on the first call to compute

I believe I fixed this quasibug, but the underlying cache bug still remains.

katestange:
Here's an example where a formula is still popping up some errors while trying to type:  backspacing the variable inside "isPrime(a)" produces a pop-up, but only for the formula field in ModFill, not for the formula field in Sequence.  Do you know how to easily fix this (seeing as you were just fixing this in another PR)?
http://localhost:5173/?name=Formula%3A+n&viz=ModFill&modDimension=13&backgroundColor=000000&fillColor=1a5fb4&alpha=0.01&aspectRatio=1&highlightFormula=isPrime%28a%29&highColor=f66151&alphaHigh=0.4&seq=Formula
Screenshot from 2024-11-25 21-20-37

katestange:
My newest commit allows for variables n (index) and a (value) in the highlight formula.  I also added checking whether they were actually in the parsetree and avoiding even trying math.safeNumber() on things that aren't in the parsetree, so as to avoid pointless failures.  This is necessary, for example, if you want to use highlight formula involving n but not a on VF numbers, where a is not a safe number after just 79 terms, but n is fine.

katestange:
Question:  I just realized I did the aspect ratio bare hands, but we have a capability for that, right?  I should have requested a canvas of the correct aspect ratio?  Don't know why I forgot that...

gwhitney:
OK, I have licked the async initialization bug, and am getting a PR ready but it takes a lot of updating tests to get a commit. In the meantime, I have some questions/comments on your latest commit:

A) On aspect ratio, I certainly didn't mean that you should write code to implement different aspect ratios within the ModFill visualizer! I was just thinking you would implement requestedAspectRatio() and then you would get a canvas of the appropriate aspect ratio. Then exactly the old computation would work. I was thinking you would either always request 1 if you think square is best, or have a checkbox that determines whether 1 is requested or the "full" aspect ratio is used. And sadly, I think the code should be reverted to something like that -- I don't think we want to be redoing aspect ratio work in each visualizer. Do you want to do that reversion, or would you like me to?

B) On the description of the visualizer, "array" is fine with me.

C) On hiding the highlight color and alpha when the highlight formula is `false` or `0`, there is already a mechanism: visiblePredicate. Do you want to use it to implement this hiding, or shall I?

D) On m-dependency in the transparency: that's nice =) `0.07*log(m)` makes "Dance no. 163" (btw, I _love_ your titles) look more balanced to me -- see what you think.

E) I like all your wording changes! I just have one additional suggestion: I think that "alpha" should be renamed "Opacity" both here and in Chaos, because the larger the value, the more opaque, not the more transparent. Then the range phrase should be "Between 0 (transparent) and 1 (completely opaque)." What do you think?

F) If no formula at all is specified for Highlight opacity, should it default to be the same thing as Opacity?

G) Some minor questions/nits on the gallery specimens, disregarding the fact that they need some updating since the parameter changes:
    (i) In Dance no. 163, should we use formula `163n` rather than `163*n` to be more math-comfortable?
    (ii) In same specimen, similarly the highlight indices formula `n%163 > 81` seems easier to read/comprehend and produces essentially the same picture. Should we switch? Note you can also write this `n mod 163 > 81` if you prefer.
    (iii) On Baffling Beatty Bars, should we similarly remove the unneeded `*` in the formula and in the Highlight indices formula? And since `floor()` is implicit in the Formula sequence, should we remove the `floor()` from the `floor(sqrt(2)n)`? (The `floor()` is needed in the highlight indices because that's not a sequence, it's just a direct use of MathFormula.)

H) Just to mention -- I don't think this necessarily illuminates anything for the typical user, but maybe should be available as an option for the power user -- you can do most everything that you can with the existing parameters, and much more besides, with MathFormulas that let you compute color-space coordinates (rgba or maybe more usefully hsva) from the index, modulus, and residue. In fact, I think those things cover most of the options in the issue proposing the various color schemes, with the exception of the fact that overlaying transparent colors gives you a weighted average of the input colors to a cell, with recent additions upweighted. If one also added straight-up averaging of color coordinates of _all of_ the "hits" on a cell as well as transparent overlaying, then I think that would cover all the cases listed in that issue. But it would also make using the visualizer much more complicated and hard to get into for the casual user. So we can revisit these ideas in the future, perhaps -- maybe have a "power user mode" that gives you direct access to such formulas.

gwhitney:
OK, finally got the fix to the cache filling async messup pushed to your branch. Don't forget to pull before you do anything else.

gwhitney:
> Here's an example where a formula is still popping up some errors while trying to type: backspacing the variable inside "isPrime(a)" produces a pop-up, but only for the formula field in ModFill, not for the formula field in Sequence. Do you know how to easily fix this (seeing as you were just fixing this in another PR)?

I somehow seem to have killed the popups in my latest commit, even though I don't understand why. I will have to look into it tomorrow. But indeed, this code is throwing errors when you delete the variable `a` in `isPrime(a)` -- just now they are only showing up in the console. To avoid this, you have to do the computation in a try-catch block (see Formula.ts). Maybe the try-catch block should be moved into the MathFormula code in math.ts -- but the error handling still has to be able to happen, that has to be different in Formula.ts and in your visualizer.

gwhitney:
OK, I don't actually understand why the error popups stopped (i.e., why errors were only going to the console, where a visitor won't see anything, rather than to an alert message on screen). It seems to have something to do with my restructuring the async calls to get rid of the "filling the cache before the parameter changes were finished" bug. But I am not really sure why that rearranging of the asyncs had any effect on this, so it took me a long time to manage to restore them, and in the end I only managed to do it by replicating the error-handling code from main.ts also in Specimen.ts, which is annoying.

But the latest commit (be sure to pull!) should get the popups back. 

Then what you want to do about the "throw-happiness" of MathFormulas in ModFill is up to you: you can wrap your calls to `.compute` in your own try-catch blocks, or you can put a try-catch block in MathFormula.compute() with some mechanism that lets the caller tailor what happens when a computation goes awry (I think that could maybe be as simple as passing in a status object where messages should go when there's a problem), or something else that seems helpful to you. If you want any help with this, let me know. I was tempted to just say that we should defer until we have multiple Sequences for a single Visualizer, but (a) the opacity as a function of the modulus isn't an integer sequence, it's a real-number sequence, and (b) the "whether-to-highlight" parameter is a function of (at least as of now) two quantities, not a sequence. So it looks like they are going to stay just MathFormula parameters, and so there is a need for some more robust error handling around such parameters.

gwhitney:
OK, sorry to keep pushing to your PR. I was able to understand a little bit more about the error handling -- basically a throw inside of an async function results in some Promise somewhere entering the 'rejected' state rather than the 'resolved' state, rather than that throw propagating farther up the call chain. That understanding allowed me to move the replicated error handling a bit farther up the call chain, which is good because presumably now it can catch errors from a bit wider scope. But I still wasn't able to avoid replicating the error handling altogether, and I don't see any way to move it any further up the call chain. I'm not going to stress about this any further unless it's troubling you.

gwhitney:
At today's meeting, we resolved that when Kate asks Glen to, he will move as much of the error-handling code from the Formula Sequence into MathFormula (basically everything except the part that assumes the result is going to be converted to a bigint, since in general the results of MathFormula are not). The communication to the visitor of any resulting errors will be via an (optional) ValidationStatus object that can be passed into `.compute()` -- if it's there, messages will be put on that status rather than thrown as errors. That way, a MathFormula client can arrange for the messages to appear on screen by passing a ValidationStatus that's tied to something already displayed on screen, like the parameter in which the formula is specified.

katestange:
The most recent commit deals with some of this
 
> A) On aspect ratio, I certainly didn't mean that you should write code to implement different aspect ratios within the ModFill visualizer! I was just thinking you would implement requestedAspectRatio() and then you would get a canvas of the appropriate aspect ratio. Then exactly the old computation would work. I was thinking you would either always request 1 if you think square is best, or have a checkbox that determines whether 1 is requested or the "full" aspect ratio is used. And sadly, I think the code should be reverted to something like that -- I don't think we want to be redoing aspect ratio work in each visualizer. Do you want to do that reversion, or would you like me to?

Fixed.
> 
> B) On the description of the visualizer, "array" is fine with me.

Thumbs up.
> 
> C) On hiding the highlight color and alpha when the highlight formula is `false` or `0`, there is already a mechanism: visiblePredicate. Do you want to use it to implement this hiding, or shall I?

Done.
> 
> D) On m-dependency in the transparency: that's nice =) `0.07*log(m)` makes "Dance no. 163" (btw, I _love_ your titles) look more balanced to me -- see what you think.

Will return to this on next commit.
> 
> E) I like all your wording changes! I just have one additional suggestion: I think that "alpha" should be renamed "Opacity" both here and in Chaos, because the larger the value, the more opaque, not the more transparent. Then the range phrase should be "Between 0 (transparent) and 1 (completely opaque)." What do you think?

Done.
> 
> F) If no formula at all is specified for Highlight opacity, should it default to be the same thing as Opacity?

I don't know how to do this?
> 
> G) Some minor questions/nits on the gallery specimens, disregarding the fact that they need some updating since the parameter changes: (i) In Dance no. 163, should we use formula `163n` rather than `163*n` to be more math-comfortable? (ii) In same specimen, similarly the highlight indices formula `n%163 > 81` seems easier to read/comprehend and produces essentially the same picture. Should we switch? Note you can also write this `n mod 163 > 81` if you prefer. (iii) On Baffling Beatty Bars, should we similarly remove the unneeded `*` in the formula and in the Highlight indices formula? And since `floor()` is implicit in the Formula sequence, should we remove the `floor()` from the `floor(sqrt(2)n)`? (The `floor()` is needed in the highlight indices because that's not a sequence, it's just a direct use of MathFormula.)

Will return to these.
> 
> H) Just to mention -- I don't think this necessarily illuminates anything for the typical user, but maybe should be available as an option for the power user -- you can do most everything that you can with the existing parameters, and much more besides, with MathFormulas that let you compute color-space coordinates (rgba or maybe more usefully hsva) from the index, modulus, and residue. In fact, I think those things cover most of the options in the issue proposing the various color schemes, with the exception of the fact that overlaying transparent colors gives you a weighted average of the input colors to a cell, with recent additions upweighted. If one also added straight-up averaging of color coordinates of _all of_ the "hits" on a cell as well as transparent overlaying, then I think that would cover all the cases listed in that issue. But it would also make using the visualizer much more complicated and hard to get into for the casual user. So we can revisit these ideas in the future, perhaps -- maybe have a "power user mode" that gives you direct access to such formulas.

Yes, that's interesting -- controlling the colours directly.  In this commit I added a pile of variables to all the available formulas, because -- why not?  I can imagine using all of them in some circumstance or other.

Other things I did in this commit:

1. made background colour work but it's a kludge until your other PR about the cache issue
2. added "Sunzi mode" which blanks the canvas between terms -- I put this down at the end, and I'm on the fence about it's existence, it's kind of interesting but not visually pleasing, what do you think?  If it's worth it, I could add "fade trails" so it isn't quite so stroboscopic... that might be more interesting and pleasing.  This would let you see if the residue distribution changes over time

Finally, about the aspect ratio, I'm not really a fan of defaulting to square.  It's kind of ok with a white background, but with any other background colour it just looks awkward on the white (this will be an issue with numberglyph at some point too).  I think it looks much nicer to use the full available space by default.  This raises the issue about the whole request aspect ratio thing in general, that there's leftover space and it's always white no matter what.  Thoughts?

katestange:
The most recent commit improves/makes some nice featured items.  I figure we can cull later (and will have to), so I'm just including the ones people seemed to like, possibly too many.  And since I can't get most OEIS sequences right now, I'm pretty limited in making nice examples.  The only advice of yours I didn't take was to remove "floor" from the beatty main sequence, because I think displaying the floor bars in the name is more understandable to the naive user.

katestange:
Ok, @gwhitney I'm passing this over to you to do the error handling stuff.  Plus, I'm hoping you might be able to do the following item I wasn't sure how to do:

> > F) If no formula at all is specified for Highlight opacity, should it default to be the same thing as Opacity?
> 
> I don't know how to do this?

To be more specific, I interpreted this as that when you turn on highlighting, it will read the parameter for alpha for the main colour, and auto-fill that for the highlighting alpha.  But maybe that isn't what you meant.  Maybe you meant that if you leave the highlighting alpha blank, it should behave as the other one... but then what does it show in grey?  It would have to auto-fill to avoid being confusing... anyway, I wasn't sure what to do there.

And finally, I'm curious on your feedback on the new changes.

gwhitney:
Just fyi you mentioned above that I needed to get the async stuff in to make background work. The funny thing is that I did get that in and when I put it in, the first value of the sequence was working and displaying fine. Now, the first value of the sequence is _computing_ properly, and it is drawing, but it's somehow getting blanked by the background, so it looks as though it wasn't drawn. You can prove this by using a sequence with number of terms set to 1 -- it looks like it draws nothing, but if you resize quickly you see as you do that it constantly draws that one term and then blanks it and after you finish, it ends up blanked. This is totally mysterious to me because the only place the background color is drawn is in setup(), and naive console.logs are showing setup is indeed called _before_ draw. So there is some other weird timing bug going on that I will have to track down as the next order of business. Then I can look at all the other stuff. Not sure when this will happen.

gwhitney:
OK so I eventually realized there was drawing code that was doing more background calls, and since it was happening before the stopping check it was blanking the prior draw. So I took all of the workaround stuff for the former async bug out and now it seems to run fine to me. But I am going to hold off on checking in any further changes until you have a chance to pull this and check whether the not drawing things at the beginning problem still seems to be happening for you, or if all seems to be well now. Let me know when you have a chance.

Glad the aspect ratio code in the infrastructure worked well. I would still personally  just have a checkbox to toggle between "1" and "fill page" myself, since those are the only settings I can imagine using, but it's up to you of course.

The sunzi mode doesn't do much for me (and I don't get the Sunzi reference?) but I do see that optionally supplying an rgba color to paint all the non-hit cells each frame could be worthwhile, in that it could let you fade out the history, and at the extreme of alpha=1 that recovers sunzi mode. So feel free to take out sunzi mode, leave it in, or replace it with the non-hit paint (better phrased) as you see fit.

On the leftover screen space when aspect ratio differs from natural: I agree that perhaps it should one of our greys, rather than white. I don't quite see how to let the visualizer control the color of that space, though -- it's precisely the space outside the visualizer's canvas. What should we go with on this?

I can take care of doing (F) defaulting highlight opacity to opacity when you give the all-clear on the background/async bug and I work on mathformula error handling.

Finally on the featured specimens, I see Residue Rise is back as Woven Residues. What do you think about making the minimum of the random sequence be equal to the max modulus to get rid of the horizontal lines?

On Picasso's Periods, I was a little surprised you didn't use the classic $n^2+n+41$. If you do that and increase the opacity of the blue a bit, you of course start with a nice rose period that turns blue with occasional pink revivals... But there may be solid number-theoretic motivations for $n^3+2n+1$ that I am unaware of.

katestange:
Sorry for the workaround confusing you!  Is the background colour working for you without it?  It isn't for me, at least in Firefox.  I thought the async/cache bug was getting fixed in a separate PR, so I put the workaround code there so the background colour would work.  For me, nothing is getting drawn (including background) until the second or third draw().  I see now that it _is_ working in Chromium.  Strange.

katestange:
The Sunzi reference is Sunzi's Theorem (Chinese Remainder Theorem) that allows you to recover a number from enough of its residues modulo different moduli.  The idea is that we are seeing one number per frame, but in the form of a collection of residues.  I like your idea about how to incorporate it as a spectrum from strobe to gradual fade -- this will let you see changing characteristics on any time scale.

gwhitney:
> For me, nothing is getting drawn (including background) until the second or third draw(). I see now that it _is_ working in Chromium. Strange.

Indeed, that is very strange. ModFill works fine for me in, and identically at least to first approximation, in Firefox, Edge, and Chromium. I take it that consistently there's no plotting going on for you until the 2nd or 3rd call to draw() in Firefox. Is it only ModFill that seems to have a problem? Run Prime Residues on Firefox but with the number of sequence entries set to 1 then 2 (and then 3 if it's qualitatively different) and post the screenshots. Thanks!

gwhitney:
P.S. For the screenshots, run Prime Residues with like highest modulus 25 and a background color, too, to make it clear what has or hasn't been drawn. Thanks.

katestange:
Will post screenshots per request momentarily.  For this commit, just added a more streamlined sunzi effect and framerate, and adjusted featured Picasso again.

katestange:
Requested Screenshots from firefox (setting should be visible; in particular, background is set to black)
Screenshot from 2024-11-27 10-29-58
Screenshot from 2024-11-27 10-30-06
Screenshot from 2024-11-27 10-30-12
Screenshot from 2024-11-27 10-30-26
Screenshot from 2024-11-27 10-30-39

gwhitney:
OK, I'm going to take a quick look at this bug insofar as I can given that it's not happening on my machine. When you get a chance, let me know what exact version of Firefox you are running (and/or try updating to as recent a version as you can, but check the version number before you update). Also please check: when you have the specimen as the screenshots above, with the number of terms set to 2 (i.e., highest number where you see nothing), if you quickly resize the browser, do you see any flashes of black, or of blue stripes near the bottom of the canvas, i.e. glimpses that it is drawing only to get wiped out by new white. In the meantime, I will try to follow up on the one weirdness I can observe here, which is that it seems to be initializing twice on page reload, when it seems like once would suffice. I speculate that maybe if I could get rid of the double initializing here, then it might start displaying properly on your end.

gwhitney:
Oh wait, I take it back. With the latest code, I only seem to be initializing once. So please do post that further information when you get a chance, and then if that doesn't ring any bells, I don't see any option but a joint debugging session when we can find the time, since nothing tangible awry is happening here :-/

katestange:
The bug has apparently disappeared (?!??!!?) so we forge onwards!

gwhitney:
OK, passing the baton back. Let me know how that works for you, in terms of error checking in all formula entry boxes and using the Opacity formula if there is no Highlight Opacity formula.

gwhitney:
Incidentally, `min(0.4, 0.01*m)` might be an even better saturation equalizer than `k*log(m)`, see what you think...

katestange:
@gwhitney I don't know why, but the "changing a sequence" e2e test is timing out.  Are you seeing this?

gwhitney:
Can't run tests right now since internet on plane is too slow. But there is one test -- I can't remember which -- that erratically times out for me (roughly one time in 3 or 4). It always passes on a re-run, in my experience. If you are getting one consistently timing out, try running just that one and then looking at the web report to see if you can see what's going on. If you're still stuck next time I have some time to look at this (likely Friday), I will poke around.

katestange:
Interestingly your bug fix resulted in a 2 pixel difference in the Chaos snapshots.  So there was something (undetectable) wrong happening from it. :)

katestange:
Ok, I think this is ready for review.  I think it's a big enough PR, that I've not implemented any new colour schemes or features in this latest round.  If you'd like to keep something for future, let me know and we can file an issue or you could feel free to add something or request something.  But nothing seems like a burning need for me right now, so I'm marking it ready.

katestange:
Also, I must add, it never ceases to amaze me how a PR I think will be small -- some colour changes in ModFill -- will end up touching 60 files!  I guess I see why you try not to overlap PRs and rebase too much.

gwhitney:
> Also, I must add, it never ceases to amaze me how a PR I think will be small -- some colour changes in ModFill -- will end up touching 60 files! I guess I see why you try not to overlap PRs and rebase too much.

Well, "only" a dozen source files, the rest are all tests. And it's happening because we are still in process of flushing out fundamental bugs, and not being very disciplined about pausing the PR, doing a separate bug fix PR, and then rebasing the feature PR, since we are a small pre-alpha project.

I take it the test timeout stopped bothering you?

gwhitney:
> Interestingly your bug fix resulted in a 2 pixel difference in the Chaos snapshots. So there was something (undetectable) wrong happening from it. :)

Just in the CI snapshots, right?

katestange:
> Just in the CI snapshots, right?

Correct!

gwhitney:
> Ok, I think this is ready for review. I think it's a big enough PR, that I've not implemented any new colour schemes or features in this latest round. If you'd like to keep something for future, let me know and we can file an issue or you could feel free to add something or request something. But nothing seems like a burning need for me right now, so I'm marking it ready.

Right, that seems like a perfectly fine decision. Do any of the other colour's scheme suggestions seem like they are worth leaving as issues (meaning we intend to do them at some point)? Or should one of us move them to Discussions (meaning we will think about them someday but have not made any specific decision to necessarily ever implement them)? I am fine either way, just let me know your inclination.

I will review what's here so far when I have the chance, my guess is Fri or Sat.

gwhitney:
Suggestion: I find the current incarnation of "Picasso's Periods" very strobe-y, to the point of being a bit unpleasant, and the blue/rose "periods" to go by  bit too quick. I'd suggest dialing the frame rate down to 30? What do you think? I understand you may not be able to respond for a while; I'll just hold on the review until you do have a chance to get back on this. If you decide you like the decision, I can push the change to defineFeatured if you want (super tiny change); if you decide you'd like to leave it as it is, I will just go ahead with the rest of the review (I already fixed one little bug, where a parameter was not disappearing if the default value of the parameter its visibility depended on was a value on which it should not appear).

katestange:
> Suggestion: I find the current incarnation of "Picasso's Periods" very strobe-y, to the point of being a bit unpleasant, and the blue/rose "periods" to go by bit too quick. I'd suggest dialing the frame rate down to 30? What do you think? I understand you may not be able to respond for a while; I'll just hold on the review until you do have a chance to get back on this. If you decide you like the decision, I can push the change to defineFeatured if you want (super tiny change); if you decide you'd like to leave it as it is, I will just go ahead with the rest of the review (I already fixed one little bug, where a parameter was not disappearing if the default value of the parameter its visibility depended on was a value on which it should not appear).

I think that's a good call.  Could even go to 20 if you prefer.  Also, you might decrease the opacity on the highlight from 0.4 to 0.3 -- if you think that looks good, you could fiddle with that too.  I'd be happy with any of these variations.  If you don't mind pushing your favourite variation in this general vicinity, that's great!

gwhitney:
Wow, I wasn't expecting a response any time soon. I ended up at 24fps, a common speed between 20 and 30; and I didn't mind the "prime bursts" being so bright, but I did ramp up the saturation of the blue periods slightly. Proceeding with review.

gwhitney:
I have just gone through the PR checklist and all seems good; merging.

PR #516, opened 2024-12-11
ui: More "finger pointer" cursors to indicate interactions.

### LINKED ISSUE #513: Need more "finger pointer" cursors
In list view in switchers, the mouse should be a "finger pointer" over the list, so it is clear it is clickable.
The sequence and visualizer names should also have "finger pointer" upon mouseover.

These were bugs (4) and (5) from #510.

gwhitney:
* Lines of the list form of SpecimensGallery components (sequence switcher, gallery, etc.)
  * The sequence name and visualizer name in a parameter tab (since they bring up the switcher modals).

Resolves #513.
<hr/>

### DIFF
--- a/src/components/ParamEditor.vue
+++ b/src/components/ParamEditor.vue
@@ -171,6 +171,7 @@
         font-size: var(--ns-size-heading);
         padding: 6px 8px 6px 8px;
         width: 100%;
+        cursor: pointer;
 
         a {
             color: var(--ns-color-grey);
--- a/src/components/SpecimensGallery.vue
+++ b/src/components/SpecimensGallery.vue
@@ -137,10 +137,13 @@
     thead {
         text-decoration: underline;
     }
+    tbody tr {
+        cursor: pointer;
+    }
     tr:nth-child(even) {
         background-color: var(--ns-color-pale);
     }
-    tr:hover {
+    tbody tr:hover {
         background-color: color-mix(
             in srgb,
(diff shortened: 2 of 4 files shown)


katestange:
Ok, I have a hard time believing these couple lines of code actually broke something, but here you have it:  this doesn't happen on ui2.  Go to http://localhost:5173/?name=Divisor+Square&viz=Chaos&corners=8&walkers=8&alpha=0.7&pixelsPerFrame=2000&seq=OEIS+A000005  then click on the word "Chaos" to open the visualizer chooser.  If it is in list mode, go to thumbnail mode.  Then go back to list mode.  I get the attached (verified in Chrome and Firefox).
Screenshot from 2024-12-10 20-18-17

gwhitney:
OK, when you get a chance try it with the latest commit. The tiny changes of course did not _cause_ this bug, they merely _uncovered_ a pre-existing bug. Recall what I said that we are still shaking out some inherent bugginess. I think the bugs mostly stem from how we are using Vue together with our need to do a lot of stuff asynchronously. So the state of the world may not be as "clean" as we'd like to think: for example, actions that we think of as happening "when a window pops up" may still not have completed by the time the visitor clicks to close the window (in particular that was what was happening in this bug). I suspect we will continue to shake out stuff like this well after alpha goes to main, sadly.

katestange:
Well I guess it's good we are shaking out bugs!  Looks good now!

PR #517, opened 2024-12-12
feat: Limit WebGL contexts in use by replacing thumbnails with snapshots when done

### LINKED ISSUE #506: Preserve WebGL contexts by replacing thumbnails with snapshots when done
There are two parts to this planned change:

(A) In Thumbnails, run the animation until it gets to its (current) stopping point, then replace the Specimen with just a snapshot of the state it has gotten to, as a png displayed on the canvas. Make sure this frees up the WebGL context of the Specimen, if it had one.

(B) When a Thumbnail is first displayed, if it is a WebGL visualizer, and there are more than N (probably 8) active WebGL contexts, show an hourglass (or loading animation) until a WebGL context frees up, at which point start the usual animation and proceed as in part (A).

gwhitney:

<hr/>

Resolves #506.

Also adds a method to the VisualizerInterface that indicates whether the visualizer uses WebGL. This could/should probably be used in testing to determine which tests use WebGL rather than checking the name of the visualizer, but that would be a different PR.

Also makes a baby step toward trying to understand the sporadic failures of the "minimizing a tab" e2e test in scope.test.ts.

### DIFF
--- a/e2e/tests/scope.spec.ts
+++ b/e2e/tests/scope.spec.ts
@@ -83,6 +83,9 @@ test.describe('Scope: on some featured visualization', () => {
     })
 
     test('minimizing a tab', async ({page}) => {
+        await expect(page.locator('#visualiserTab')).not.toHaveClass(
+            /minimized/
+        )
         await page.locator('#visualiserTab .minimize').click()
         await expect(page.locator('#visualiserTab')).toHaveClass(/minimized/)
         await expect(
--- a/src/components/Thumbnail.vue
+++ b/src/components/Thumbnail.vue
@@ -14,19 +14,32 @@
     let savedContainer: HTMLDivElement | null = null
     let specimen: Specimen | undefined = undefined
     let usingGC = false
+    const thumbnailGCmax = 8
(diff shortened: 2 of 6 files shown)


katestange:
Query:  how is the order of thumbnail ordering chosen?  On my computer it seems to choose rather randomly (not the first 8 thumbnails).

gwhitney:
Well, when each thumbnail starts up, it looks at the count of how many have already started up, and if it's less than 8, it displays, otherwise it waits. So the browser's async call scheduler is determining the order, maybe together with how many times each specimen card happens to yield control via an async or await call. In other words, the current scheme doesn't allow much control over the ordering. If it's important that they appear to fill in with the same ordering as they appear in the switcher (left to right, top to bottom), I'll have to come up with another scheme... Let me know.

katestange:
No, I don't think fussing over the ordering is necessary for alpha, and will be naturally revisited later when we get to the later stages of this webgl thumbnail management after alpha.  Ok, well, I can't break it by messing around and it passes all tests on my machine, etc.  I can't think of anything to improve, so I'll merge!

PR #522, opened 2024-12-18
fix: Restore Grid visualizer to operation

gwhitney:

<hr/>

This PR is a work in progress, aimed at restoring the Grid visualizer. Currently it compiles and runs without crash, but still does not operate.


gwhitney:
@katestange @Vectornaut this PR so far fails e2e tests. That's because unfortunately I had to re-do some of the Delft scheme for rendering the ParamEditor. So I don't quite have the layout matching exactly, and in fact I doubt I will get it to match _exactly_ because the arrangement of divs and such is slightly different. It looks broadly similar, though. The main current difference, as you can see for example by checking this out and then executing `npm run test:e2e:report` and using perhaps the "slider" to compare the actual and expected for the failing tests, is that the sections in the ParamEditor for each parameter have less vertical whitespace between them than they used to. Before marking this PR as ready for review, I would appreciate some feedback from one or both of you as to whether the "ideal" amount of whitespace between parameters should be basically similar to what it was before this PR, basically similar to what this PR is currently laying out, or perhaps right in between these options.  Thanks so much!

katestange:
I wasn't at all sure what I would like best before looking at the comparison.  But I find I prefer the tighter spacing.  I think it is more consistent with the current visual style.

gwhitney:
Super, thanks for looking. Should I try just a trifle looser than now, or leave it just how it is in the version I checked in for examination?

katestange:
> Super, thanks for looking. Should I try just a trifle looser than now, or leave it just how it is in the version I checked in for examination?

Do what feels good to you.

gwhitney:
> > Super, thanks for looking. Should I try just a trifle looser than now, or leave it just how it is in the version I checked in for examination?
> 
> Do what feels good to you.

Life advice ;-)  OK, I went through Grid again and tried to optimize the layout within the new structure. This was a combination of loosening it just slightly from where I had ended up, and centering the parameters within their space better. I am pretty happy with where things have ended up, so I updated the snapshots (and I think the new layout settings look pretty good across the other visualizers, too.

I do have a new question based on other changes I made to try to make the intricate parameters of Grid clearer, so when you have a chance, please pull the latest commit and let me know what you think about the following:

I moved all of the parameters related to the properties shown to the end of the (Grid) visualizer tab, and made it so that the Property 1, Property 2, etc. parameters only show up if the enumeration that lists pre-configured collections of properties is set to its "Custom" option. This change made it pretty natural to call that parameter "Highlight properties." The idea is that you can pick one of the collections of properties to highlight off of the list, or if you want, select "Custom" and then get a sequence of slots in the visualizer tab to select individual properties (and the parameters for each) to highlight. And I made the "Primes" option of the "Highlight properties" be the default. The question is, suppose you switch to the Grid visualizer, and then you choose one of the preconfigured options of "Highlight properties" (maybe "Polygonal numbers"), and then select "Custom". What state should the Property 1, Property 2, ... parameters be in? I can think of at least four ways to go with this:

A) The current behavior is that the default setting for the custom Property 1, Property 2, ... matches the "Primes" setting from the preconfigured ones. We could just leave it that way. It seems potentially a little confusing in that if you happen to be on "Primes" nothing will appear to happen when you switch to Custom, and if you are on something else like Polygonal, it will look like it is switching back to Primes when you select Custom.

B) Choose a different initial/default setting for the custom Property 1, Property 2, ... that doesn't correspond to any of the preconfigured options in "Highlight properties" -- for example, set it up to highlight numbers divisible by 3 in green, say.

C) Just set all of the custom properties to "None" so you get a blank slate entirely the background color on which you can start "painting" properties. Conceptually attractive, but could this look to some visitors like it just "stopped working" when you selected "Custom"?

D) Set the custom properties to match the most recently selected pre-configured collection of properties, so that whenever you select "Custom", nothing will change in the visualization, but in the Visualizer parameter tab, you will then see all of the parameters of all of the properties being displayed by whatever that pre-configured collection was, and you can start customizing from there. As with (A) there is the possibility of some confusion that selecting Custom doesn't do anything, but on the other hand, the behavior will now be consistent no matter what pre-configured collection you happened to be looking at, and it will allow you to see the details of each pre-configured option. It also has the advantage that right now, if you liked say the "Color by Last Digit 1" option but wanted to change just the color that multiples of 10 produce, it's really tough, but with this alternative, that would become really easy: just select "Color by Last Digit 1", then select "Custom", then modify the highlight color of the "Last Digit is" "Digit 0" property.

Any thoughts on which way to go with this bit of the UI design? There may of course be other good options besides any of (A)-(D) above.

[Note that this PR is still far from ready for review in general -- there are definitely still bugs with the Grid Visualizer. For example, for me, the first time I load it the display is garbled, and it only seems to right itself once I change one of the parameters.]

gwhitney:
OK, now Grid actually seems to be doing what it's intended to do. Still need guidance on options (A-D) from last comment, and there's still plenty of other things that need improvement, like placing the spiral when the aspect ratio is not 1.

gwhitney:
OK, Grid now deals with aspect ratio better (I just insisted on square when using the Spiral path, rather than trying to implement rectangular spirals). More to come, happy to get guidance on properties whenever.

gwhitney:
OK, now we have translucent property colors. I also made the Turtle path allow opacity setting, since we now have an rgba color ParamType (`ParamType.ACOLOR`). I made the "Beatty DNA" trace a bit transparent; it seemed to add some richness. I left all of the other feature specimens with opaque traces, although I think "Wait For It" also looks good at very low opacity, like hex 20. You may want to try it, and feel free to change the featured specimen if you like.

katestange:
> I think "Wait For It" also looks good at very low opacity, like hex 20. You may want to try it, a

Oh yes it looks very interesting with hex 20!  Please go ahead and change that if you are back in the featured specimens.

katestange:
> Any thoughts on which way to go with this bit of the UI design? There may of course be other good options besides any of (A)-(D) above.

The advantage of D in utility seems significant to me; it allows you to understand the presets.  Maybe we could try that, and see if it causes too much confusion?  All of them seem to have their points of confusion.

gwhitney:
OK, I have done (D), and changed the Wait For It stroke. Let me know what you think. Next up on my end is to allow arbitrary formulas for properties, reducing the number of ad-hoc named properties on the list, while allowing many more possible properties.

katestange:
It looks good!  I like the solution (D).  It's nice because it lets you find out the settings for any preset.  I wonder if we should even show the custom options all the time and simply have the preset populate them?  Then when you change one of the options, you automatically move into "custom" mode?  Pros:  even easier and more intuitive to edit a preset.  Cons: is it cluttered?

gwhitney:
> wonder if we should even show the custom options all the time and simply have the preset populate them? Then when you change one of the options, you automatically move into "custom" mode? Pros: even easier and more intuitive to edit a preset. Cons: is it cluttered?

I am open to it, but yes I am worried it will be overwhelming at first. We can defer to the next time we are all meeting together to decide maybe, or if you want me to check in a commit that shows the properties all the time, let me know.

gwhitney:
Just a note so I don't forget: Grid in ui2 has a mathematics error in that it claims every number with an even number of 3 mod 4 prime factors to odd powers is a sum of two squares (e.g. 77), when in fact only those numbers with zero 3 mod 4 prime factors to odd powers is a sum of two squares. This needs to be fixed in this PR.

However, the math needs of the Grid overhaul to allow arbitrary formulas as properties are sufficiently strong that my plan is actually to put this on hold and do a PR that updates to mathjs 14, now with 100% more bigint support, and then rebase this once that PR gets through.

katestange:
> Just a note so I don't forget: Grid in ui2 has a mathematics error in that it claims every number with an even number of 3 mod 4 prime factors to odd powers is a sum of two squares (e.g. 77), when in fact only those numbers with zero 3 mod 4 prime factors to odd powers is a sum of two squares. This needs to be fixed in this PR.

Actually the correct criterion is the following:  every prime divisor that is 3 mod 4 must appear to an even power.  Hence 7^2 is a sum of squares.

katestange:
> However, the math needs of the Grid overhaul to allow arbitrary formulas as properties are sufficiently strong that my plan is actually to put this on hold and do a PR that updates to mathjs 14, now with 100% more bigint support, and then rebase this once that PR gets through.

Exciting!!

gwhitney:
> > However, the math needs of the Grid overhaul to allow arbitrary formulas as properties are sufficiently strong that my plan is actually to put this on hold and do a PR that updates to mathjs 14, now with 100% more bigint support, and then rebase this once that PR gets through.
> 
> Exciting!!

@katestange @Vectornaut New Year's puzzle: I am fixing bigint bugs in mathjs, and one of them is that it will not choose a random bigint uniformly from an interval. If the interval is small, this choice is easy to make (at least to the quality of mathjs's existing pseudorandom number generator). However, if the the interval is very large, certainly once it's over 2^53 in length, this seems tricky to do in guaranteed finite time. My plan is to choose the smallest _N_ so that $2^N$ is larger than (or equal to) the length of the interval, and then do _N_ coin flips, and if I get unlucky and land outside the interval, redo the process until I land in the desired interval. But this method, of course, although it almost surely terminates, does not surely terminate. Any thoughts on a method that surely terminates? My concern is that if the interval length is a very large prime, there is no way to accurately split the choice into subchoices, since we only have 53 bits of uniform distribution in the base PRNG we are using.

I think this is mainly of academic interest, since I think in practice the "repeat N coin flips til you get lucky" method will work just fine. But I was curious if I was missing something, or if this was a known problem maybe with a known solution -- a quick web search didn't turn up anything obvious.

Vectornaut:
@katestange

> Any thoughts on a method that surely terminates?

Every method I've thought of so far would sacrifice a little uniformity, so you end up with a tradeoff between uniformity, speed, and memory use. You're already limited by the amount of randomness available to the base pseudo-random number generator, so maybe you could estimate how much uniformity you're losing from that; you could use the estimate to help decide how much additional loss of uniformity is acceptable.

Vectornaut:
@katestange Here's another thought about deciding how much loss of uniformity is acceptable. The [`randomInt`](https://mathjs.org/docs/reference/functions/randomInt.html) function must already be implemented for the JavaScript `Number` type. I'm guessing that [this](https://github.com/josdejong/mathjs/blob/164c6b0403be2b39bd31733fe03ce42dc0423b24/src/function/probability/randomInt.js#L64-L66) is the implementation (although I don't know enough TypeScript to understand how calling `randomInt` would lead to `_randomInt` being called). The uniformity of that implementation is limited by the precision of the `Number` type. You could choose a measure of uniformity, estimate its value for `randomInt` over `Number`, and then try to make sure that `randomInt` for `BigInt` is at least that uniform.

katestange:
> My plan is to choose the smallest _N_ so that 2 N is larger than (or equal to) the length of the interval, and then do _N_ coin flips, and if I get unlucky and land outside the interval, redo the process until I land in the desired interval.

This is called rejection sampling and it is what is typically used.  For example, in [numpy](https://github.com/numpy/numpy/blob/dfb00057a5bba94284dfbebffe37c37b08a16d42/numpy/random/_generator.pyx#L671), they use something called [Lemire's method](https://shufflesharding.com/posts/dissecting-lemire), which is a fancy version of rejection sampling.  The underlying pseudo-random number generator is typically a bit-stream generator (one random bit per call).

katestange:
> Here's another thought about deciding how much loss of uniformity is acceptable.

Thanks for the link.  I feel fairly strongly that we should aim for perfect uniformity.  Some of the visualizers (like Chaos) are very sensitive to the statistical properties of the PRNG.  I don't see a problem with Glen's rejection sampling method, except that it could be modified a bit to avoid it taking an annoyingly long time in corner cases (e.g. random between (1,2^n+1).  I'll try to find some helpful advice on that that isn't as complicated at Lemire's.

katestange:
Ok I learned a bit more.  I think plain rejection sampling as Glen described is fine, since the desired window is always at least 1/2 the multiple-of-2 window, so you have at least a 50% chance of win.  Lemire's is written up very nicely [here](https://github.com/colmmacc/s2n/blob/7ad9240c8b9ade0cc3a403a732ba9f1289934abd/utils/s2n_random.c#L187).  It is just a more efficient way of doing rejection sampling in the situation where your underlying PRNG is giving you randomness in fixed bitlengths, like bytes, and you might want to divide.  So I don't think it's important for us; if that matters, it should be implemented in javascript's random integer in range functionality already.

katestange:
@gwhitney I want to clarify what you said.  Here's what I was imagining you said:  say you want random in range [0,k].  Then take the smallest n so that 2^n > k and get random from [0,2^n).  Reject if outside [0,k].  Is that what you meant?  Then I think it's ok.  You average one rejection since k is at least half of 2^n.  But I don't see the relevance of the primeness of k in this proposal.

gwhitney:
> @gwhitney I want to clarify what you said. Here's what I was imagining you said: say you want random in range [0,k]. Then take the smallest n so that 2^n > k and get random from [0,2^n). Reject if outside [0,k]. Is that what you meant? Then I think it's ok. You average one rejection since k is at least half of 2^n. But I don't see the relevance of the primeness of k in this proposal. (Note: the point is that getting random in [0,2^n) is just a bunch of calls to random in [0,2^s) concatenated, where s is small enough to invoke the javascript PRNG.)

Yes that's my proposal and I agree primeness is not relevant. It seems relevant to thr existence of a certain-to- terminate algorithm; it's hard to see how to break a choice among p things into two choices in a mathematically exact way that doesn't need just as much precision in each choice.

katestange:
> Yes that's my proposal and I agree primeness is not relevant. It seems relevant to thr existence of a certain-to- terminate algorithm; it's hard to see how to break a choice among p things into two choices in a mathematically exact way that doesn't need just as much precision in each choice.

Got it.  I say go with your original rejection sampling proposal.

gwhitney:
> Got it. I say go with your original rejection sampling proposal.

Great: https://github.com/josdejong/mathjs/pull/3345 submitted. This PR is basically on hold until Jos acts on that PR, so if anyone wants to submit a PR in the meantime, it would be fine.

katestange:
> This PR is basically on hold until Jos acts on that PR, so if anyone wants to submit a PR in the meantime, it would be fine.

Is it a good moment for the self-similarity PR?  It's not done but it's close.

gwhitney:
> Is it a good moment for the self-similarity PR? It's not done but it's close.

Yes, perfectly fine. I would say this won't get back into gear until the second half of January, most likely.

gwhitney:
NOTE this PR is out of sync since the update to main to correct a contributor name, and will likely need to be closed and resubmitted. But since we are likely to rebuild Grid as Spiral based on a generalized FormulaGrid visualizer anyway, that shouldn't really be a burden. Leaving this open as a placeholder until the pathway is worked out.

gwhitney:
Note further that its first three commits had to be grabbed into the color math PR in order to get the new Turtle parameter layout to work, so those changes do not need to be made in any future reconstruction of this PR.

gwhitney:
The plan is now to revisit this _after_ alpha, when we figure out how to make one visualizer an easier interface "skin" over an advanced formula-based visualizer. So closing; this effort will need to be taken up from scratch after alpha.

PR #526, opened 2025-01-03
feat: self-similarity visualizer

katestange:

<hr/>

This PR adds a self-similarity visualizer to replace ShiftCompare.

In addition, it adds biggcd, valuation and divides to math.ts.  These may not be needed once math.js does all its bigint stuff.  It also adds a few more unit tests for modulo.

Things still to be done:

- [ ] Finalize parameters interaction
- [ ] Add documentation
- [ ] Update snapshots

### DIFF
--- a/src/shared/__tests__/math.spec.ts
+++ b/src/shared/__tests__/math.spec.ts
@@ -40,6 +40,7 @@ describe('modulo', () => {
     it('gives the bigint remainder upon division', () => {
         expect(math.modulo(7, 5)).toBe(2n)
         expect(math.modulo(large, 10)).toBe(3n)
+        expect(math.modulo(large, 1)).toBe(0n)
         expect(math.modulo(12, 5n)).toBe(2n)
         expect(math.modulo(99999999999999999999999999n, 100n)).toBe(99n)
     })
@@ -47,8 +48,14 @@ describe('modulo', () => {
         expect(math.modulo(-7, 5)).toBe(3n)
         expect(math.modulo(-large, 100n)).toBe(7n)
         expect(math.modulo(25, 5n)).toBe(0n)
+        expect(math.modulo(0, 1n)).toBe(0n)
+        expect(math.modulo(1, 1n)).toBe(0n)
+        expect(math.modulo(1n, 1)).toBe(0n)
+        expect(math.modulo(-1, 1n)).toBe(0n)
     })
     it('requires a positive modulus', () => {
+        expect(() => math.modulo(77, 0)).toThrowError('Attempt')
+        expect(() => math.modulo(large, 0n)).toThrowError('Attempt')
         expect(() => math.modulo(77, -7)).toThrowError('Attempt')
         expect(() => math.modulo(-109n, -6n)).toThrowError('Attempt')
     })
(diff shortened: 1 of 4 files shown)


katestange:
Ok, before moving on to documentation and snapshots, I think this needs a little feedback on how to handle the parameters.  I think I have a tendency toward "poweruser" parameter setups, so I'm happy for some feedback how to handle things.

katestange:
Another thing I could use advice on is how to manage the lines drawn per frame.  The reason shift-compare tended to hang is that it tried to draw the whole screen at once on difficult computations.  I've made this one do some number of rows per frame so the picture comes in gradually.  But I'm not sure how best to manage the number of lines per frame so that the sequence switcher in thumbnail mode (for example) is not laggy.

katestange:
There's another weird bug I'm encountering, maybe you can confirm/deny for me.  Go to this url: http://localhost:5173/?name=IntegerStellar&viz=SelfSimilarity&width=400&distance=2&height=400&seq=Formula  Do you see a nice picture, or a plain yellow screen?  I sometimes have to change the Transformation Formula to m*1 from m (should be equivalent!) to get the nice picture to show up!

gwhitney:
Hmm, `divides` was already in math.ts, so now it is doubly defined. Please remove the excess one, thanks. And the reason for `b >= 0 && b <= 0` is that inequalities work with disparate numeric types (bigint mixed with number, for example), but equality testing throws an error.

gwhitney:
Indeed, modulo and gcd will not be needed after the mathjs update, so no need for more modulo tests.

gwhitney:
> There's another weird bug I'm encountering, maybe you can confirm/deny for me. Go to this url: http://localhost:5173/?name=IntegerStellar&viz=SelfSimilarity&width=400&distance=2&height=400&seq=Formula Do you see a nice picture, or a plain yellow screen? I sometimes have to change the Transformation Formula to m*1 from m (should be equivalent!) to get the nice picture to show up!

I can confirm that from a cold load, the above URL is an almost-all-yellow screen, but that changing some of the parameters, e.g. just overtyping the default formula 'm' with an explicitly entered 'm', fixes it to the nice diagonal rows of different intensity dots.

gwhitney:
A comment that is likely only of academic, not practical, interest: conceptually, this visualizer could be unified with Grid by (a) allowing a formula for the value to be entered in cell (x,y), e.g. `gcd(a(x), a(y))` and (b) adding a color data type to mathjs and allowing color formulas based on the value _c_ entered in a cell and the max entry _M_ and min entry _m_, e.g, `rgba(255, 255, 0, (c-m)/(M-m))`.

gwhitney:
I will take a shot at debugging the "weird bug"; I'll let you know if I succeed or stop working on it for now.

gwhitney:
OK, I fixed the weird bug in the last commit (you didn't have the correct inputs in the default value for shiftFormula). But there is another bug: you are not handling cache misses correctly, in that you loop over 500 cells, and if you miss a cell because the try-catch in drawNew fails, you never go back to that cell. That leaves a blank gap in the diagram the first time you use a given sequence. It gets filled in if you change some other parameter than the sequence, because by then the cache is full. I will leave this for you to fix, but let me know if you need help.

katestange:
> Hmm, `divides` was already in math.ts, so now it is doubly defined. Please remove the excess one, thanks. And the reason for `b >= 0 && b <= 0` is that inequalities work with disparate numeric types (bigint mixed with number, for example), but equality testing throws an error.

Thanks, that was a manual merge error, fixed.

katestange:
> I will leave this for you to fix, but let me know if you need help.

I believe I have fixed it.  It simply retries the same location until the cache successfully returns.

katestange:
@gwhitney question:  I'm trying to update snapshots.  But this PR is failing not only on the new featured items, but also on a bunch of transversal items.  Is it the case that the existence of a new visualizer has resulted in a different subset of transversal items, and therefore it's resurfacing out of date snapshots (for example Modfill)?  That's what it looks like (e.g. some ModFill look out of date, some snapshots are just missing).  If so, is it ok to go ahead with re-snapshotting those?

gwhitney:
Yes and yes.

gwhitney:
> A comment that is likely only of academic, not practical, interest: conceptually, this visualizer could be unified with Grid by (a) allowing a formula for the value to be entered in cell (x,y), e.g. `gcd(a(x), a(y))` and (b) adding a color data type to mathjs and allowing color formulas based on the value _c_ entered in a cell and the max entry _M_ and min entry _m_, e.g, `rgba(255, 255, 0, (c-m)/(M-m))`.

Actually, after diving a bit more into the review, I am now thinking this comment is of more than academic interest, and that we should actually think carefully about the design of Grid vs SelfSimilarity before proceeding with both/either. I came to this perspective after seeing that four of the five Gallery examples of SelfSimilarity are simply calculating an opacity based on a function of the (x,y) coordinates of a pixel, and two of them are just binary properties, and two of them don't even use the "distance" defined in the visualizer at all.  So at least a few of them seem much more like "Grid" visualizations rather than "SelfSimilarity" -- the latter name anyway speaks conceptually much more of things like autocorrelation in a sequence, which doesn't really seem to be going on much in the examples so far.

So there is an underlying complicated visualizer here, that can do everything that either grid or self-similarity can. Call it maybe FormulaGrid: You get pixel coordinates _x, y,_ and _s_ (the spiral distance from the center), and function symbol _a_ that computes the Sequence selected on an input, and then you can supply one (or more) formulas that compute an rgba color to draw in that cell. Perhaps there should be a shape (rectangle, small square, ellipse, small circle) -- associated with each formula if there is more than one.

Should we implement that visualizer? In its raw form, it will not be neophyte-friendly. But it could be very fruitful for "power users".

If we do implement that visualizer, what of Grid and SelfSimilarity? Do they become simply simplifying options/presets for FormulaGrid with a clever UI layout in an effort to make FormulaGrid more accessible at first? Or do they become separate visualizers in their own right, with their own specific controls focused on their particular perspectives on visualizing the sequence, perhaps with autogenerated links at the bottom of their parameter panes that take you to the FormulaGrid visualizer configured to reproduce that particular visualization?

I think I kind of like this latter option. In this scenario, we would have FormulaGrid as described above. And then maybe the current Grid would become the Spiral visualizer, focusing on settings/parameters that are interesting from the point of view of exploring number spirals, but the curious advancing user could transition to the FormulaGrid visualizer using the link provided. And analogously, the SelfSimilarity visualizer could focus on autocorrelation (broadly interpreted) of a sequence. And those examples like Modular Multiplication Table that really are best described just as a color function of (x,y) coordinates could be Gallery examples directly of the FormulaGrid visualizer -- simple cases of it that produce striking images, that would serve as an on-ramp to a rather high-level, abstract visualizer.

So @katestange @Vectornaut I'm going to hold off on further review until we have a chance to discuss these points and settle on a strategic or at least tactical direction here. Maybe we could even manage it in person at JMM... Or if you have a clear vision of how to proceed here, Kate, feel free to just share it.

katestange:
Hi @gwhitney @Vectornaut -- this is an important question, and it's been bouncing around my head a little too.  I think Glen's proposal is an excellent one, and aligns with our goals of making a powerful but accessible tool.  I don't see any reason to avoid duplication of functionality, when there's a different use-case for different access points to that functionality.  Meaning, I don't think it's a problem that two visualizers can create the same picture, if they serve different communities to get to that picture.  I think Glen's design -- having "on ramp" visualizers that are accessible and do "part" of the whole thing, is a lovely solution!  Vectornaut, thoughts?

If we go this way, I'm assuming the code design is to implement the power visualizer first, and then have the others be derived from it or layered on top of it in some way (so they are shells that just call the power visualizer under the hood based on their parameters)?

gwhitney:
Yes, I think the most practical implementation route would be to implement the putative FormulaGrid, and then see if we can derive visualizer classes from it with streamlined interfaces focusing on certain sorts of visualizations, but calling FormulaGrid under the hood. Would be ideal to have just one actual implementation of the drawing part -- that's the most DRY and maintainable.

katestange:
Ok, well, unless @Vectornaut objects, I think we should go ahead.  What's the order of operations, do you think?  Does it need to wait until bigint is in?

gwhitney:
It's certainly most convenient to wait until bigint is in. Along with/on top of that, we can add factorization, valuation, elementary number theoretic function like sum of divisors, number of divisors, euler's phi, etc. (some of those we can also submit to mathjs itself, but unlike the fixes to current bigint support we don't need to wait for thos submissions to go in). We can add a color type to mathjs if we decide to go that way (I think we may as well). Then implementing FormulaGrid will be very straightforward and not a lot of code. The main design question on my mind for FormulaGrid is whether we want to allow the order in which the cells are filled in to be customizable: i.e., always left to right, top to bottom, or whether we want to allow them to be drawn along a spiral path for the spiral cases, etc?

It also occurs to me that by allowing hex and/or triangle shapes as well, we can lay out the number triangles (like the arithmetic triangle) in its usual way to get those nice modular colorings with this same visualizer. We could allow hex/triangle spirals as well,  I suppose.

gwhitney:
NOTE that this PR is out of sync since the update to main to correct a contributor name. It will likely need to be rebuilt on top of the new ui2. But in fact, I think we are planning to rebuild it on top of a new unified underlying FormulaGrid visualizer, so that shouldn't really be any extra burden. Leaving this open as a placeholder until all of the details are worked out.

gwhitney:
And in fact, based on #541, we are likely to just redo this after alpha, for beta, as (one of of two) skins on an underlying formula-based grid "pre-visualizer". So this PR will likely be closed at some point, but keeping it around for now as a placeholder for lots of ideas and prototypes.

gwhitney:
Should we close this, in the same way that I just closed #522 ?

PR #528, opened 2025-01-15
feat: Color math

gwhitney:

<hr/>

This PR adds the color class of [chroma.js](https://github.com/gka/chroma.js) to Numberscope and incorporates it into mathjs, with `c + d` interpreted as overlaying color `d` on color `c` (and hence associative and non-commutative, with `c+d = d` whenever `d` is opaque), and `s*c` for `s` a number and `c` a color to mean "multiply the alpha of `c` by `s`". It also adds named constants to mathjs for all of the css colors, so that one can now write such formulas as `blue + 0.3*yellow` to get the standard blue overlaid with 30% opaque yellow.


gwhitney:
I have converted this to draft because I am reluctant to merge a PR with "disconnected" functionality: As initially submitted, there is no use of the Chroma class or color formulas anywhere within Numberscope. So there are a few possible ways to go:

1. If there is a visualizer in ui2 as it stands now that would clearly benefit from color formulas (rather than just color constants), please just point me to it and I will incorporate color formulas;
2. I could combine this PR with one establishing an initial `FormulaGrid` visualizer, which would definitely use color formulas (that is after all the point of doing this work); or
3. Given that it is self-contained and passes all its own tests, we could just review and merge this as-is, making an exception for adding unused code, for expediency in getting ui2 -> main.

If we go with option (1) or (2), I have one main worry: right now when you have a color parameter, you get a nice color picker to help you choose a color you like. If one switches such a thing to a MathFormula parameter, even if it returns a color, you will just be faced with a text entry box, and if you want to find a particular color constant to use in your expression, you will either have to use a css name for a color (like `blanchedalmond` -- really! -- and so on), or you will have to guess/use some other tool to find the parameters to write something like `chroma(0.35, 0.22, 0.86, 0.5)`, and there will be no immediate color feedback on what that color looks like. So should we have a ColorFormula parameter type and it gets both a text box and a color picker, and whenever you select a color in the picker the mathjs expression for that color automagically inserts in the text box? And perhaps conversely whenever you highlight a `chroma(...)` expression in the text box and open the color picker, it starts up showing the corresponding color? And just to put the pie way up in the sky, should we try to go so far as to display any `chroma(...)` subexpressions in the formula with the corresponding color as the background for that part of the expression (say)?

I look forward to your thoughts/guidance on how we should proceed with this PR, and on what facilities we need in the near term for generating/entering/editing/dealing with color formulas. Thanks!

gwhitney:
Just to try to show what I am imagining, the pipe dream is for formula entry boxes (or at least ones designated for colors) to end up looking something like
$$\texttt \colorbox{blue}{blue} + 0.5*\colorbox{red}{chroma(1, 0, 0, 1)}$$
or that sort of thing. Probably with the text displayed in a better contrasting color.

gwhitney:
We discussed this at meeting and decided that if I add a "formula mode" to the Turtle visualizer in which the step length, turn angle, and color of each segment are determined by formula, that will use the color formulas and connect this code to the rest of Numberscope, as well as provide real-world test that it is working.

We also decided not to require any additional features to facilitate color formula entry: color pickers etc can be added in the future.

gwhitney:
OK, as a stepping stone to color formulas I have allowed the width and color of segments to depend on the sequence value as well. (You will see a "formula mode" toggle at the bottom, but it is not implemented yet.) To get the colors to stay when you drag etc. I had to draw the segments with `rect` instead of `line` (and have not yet updated the snapshots).

Dig this enhanced "Wait for it" http://localhost:5173/?name=Wait+For+It&viz=Turtle&domain=-1+1&turns=30+120&steps=30+30&widths=4+1&strokeColor=%237a9f6f+orange&bgColor=5d509f&seq=Formula&formula=sign%28sin%28n%2B1%29%29

Wait_For_It

My plan is to update the formulas as the tables are changed so that when you change to formula mode, the behavior does not change, but not vice versa (since the formulas are way more general than the tables). I will get that working, fix the snapshots, and then mark this as ready for review. But I am happy for any feedback in the meantime.

gwhitney:
P.S. This already gives another way to get semitransparent colors: you can just add two more hex digits at the end of a #hhhhhh color string and they will be interpreted as alpha. Inserting `60` after `7a9f6f` in the above Wait For It example looks pretty good.

katestange:
This is gorgeous!!

gwhitney:
This PR is obsolete following the ui2 rebase. Closing.

PR #530, opened 2025-01-20
feat: Allow color computations in MathFormula and use to allow computed Turtle colors

gwhitney:

<hr/>

This is a reconstituted version of #528. It incorporates the color computations of chroma.js into mathjs and allows them to be used in MathFormulas to compute colors. It then uses this facility to allow the stroke color (and all other segment parameters of Turtle) to be computed by formula.

With the more complicated options for the Turtle parameters, it also adopts the changes to ParamField and ParamEditor layout from the first three commits of #522.


gwhitney:
OK, this now picks up where #528 left off. You can now switch back and forth to the formula display of Turtle parameters, but the formula parameters do not yet have any effect. I will continue the implementation and mark as ready for review when I think all is well, but welcome feedback now or at any point.

gwhitney:
Now the list rules convert themselves into formulas whenever they are entered, but Turtle still only draws using the list rules.  Also, all of the checks should now pass (at least when run locally). This way, I can switch over to always using the formulas for drawing and make sure that I get exactly the same images.

gwhitney:
OK, Turtle is computing with formulas for all the attributes of each segment: turn, step, width, and color. When you are in List mode and you change the list parameters, it updates all of the formulas to be ones that implement the list rules. I suspect I need to update the Github snapshots, but this can be reviewed in the meantime. And it could really use a Turtle example that actually uses one or more of the formulas in a meaningful way, so Kate if you have or can cook up one or more such example(s), please push a commit that adds it/them

Note that the translated formulas for the list rules are pretty klunky; they are done in a way to hopefully work in all circumstances. If you just want the turn angle to be 30 degrees times the sequence value mod 5, for example, you can write `30*(a%5)`. If you want the step lengths to be some table lookup covering all of the residues of the sequence value mod 3, you can write for example `[2.5, 1.5, 3][1 + a%3]` (sadly mathjs expression array indexing is one-based not zero-based).

gwhitney:
OK, snapshots updated. All differences were just font/text layout. So the new Turtle is reproducing the old one, while extending it to allow formulas. As far as I am concerned, this just needs a commit with a Featured specimen (or more) that uses formulas, with at least one color formula involved, and the usual review of the PR itself.

gwhitney:
We had a sort of "group review" in meeting today, and it generated the following TODO items; I will mark this PR as draft until they have been implemented:

- [ ] The color chooser in List mode should insert a space before the new color if the current value is nonempty.
- [ ] Allow descriptions to have links.
- [ ] The color input should have a label "A color or list of colors, in order corresponding to the sequence values listed in Domain". The word "color" could be a link to our documentation page/section on color specifications.
- [ ] Why is `chroma(255)` blue? That does not seem right/intended.
- [ ] Allow `chroma()` expressions in the color list in List mode to be consistent with documentation
- [ ] The Color chooser should fire on select action, not change action.
- [ ] Formulas also need to change when the Domain changes!
- [ ] Add a `rainbow()` constructor (using the oklch color space) for colors.
- [ ] Conform the variable names to n - sequence index; a - sequence value; s - step serial number starting from 1; f,l - first and last sequence index; x, y -- coordinates; b - bearing.

I am not planning to make any other comments as/when I simply push a commit that implements one or more of the above, but I will check the items off as I do, and mark this again as ready for review when they are all done.

katestange:
This produces a disconnected path.  This is perhaps a malformed URL in some sense, as it includes both List and Formula mode parameters.  To reproduce from first principles, open Wait For It, then go to Formula mode.  Change the formula entries to a, a, 1, white; then change the sequence formula to xgcd(3,n)[2].  The resulting path is disconnected, but as far as I can tell from the choices "a,a,1,white" it shouldn't be possible?

http://localhost:5173/?name=Formula%3A+n&viz=Turtle&domain=-1+1&turns=30+120&steps=30+30&widths=2&strokeColor=%237a9f6f&bgColor=5d509f&ruleMode=1&turnFormula=a&stepFormula=a&colorFormula=white&seq=Formula&formula=xgcd%283%2Cn%29%5B2%5D

Screenshot from 2025-01-24 18-56-37

katestange:
I'm trying to do things like 'chroma.scale()(0.5)' in the colour formula but it doesn't work.

katestange:
Here's a nice possible specimen.  http://localhost:5173/?name=The+vertigo+of+divergence&viz=Turtle&speed=10&ruleMode=1&turnFormula=a%5E1.0012&stepFormula=10&colorFormula=chroma%28%27skyblue%27%29.set%28%27hsl.h%27%2C+a%25360%29&seq=Formula&last=9999&length=10000

katestange:
And this is convergence of the first Riemann zero (approximately....)  http://localhost:5173/?name=Convergence+of+the+first+Riemann+zero&viz=Turtle&speed=10&ruleMode=1&turnFormula=log%28n%29*14.134725*360%2F2%2Fpi&stepFormula=200%2Fsqrt%28n%29&colorFormula=chroma%28%27skyblue%27%29.set%28%27hsl.h%27%2C+a%25360%29&seq=Formula&first=10

katestange:
See also: https://community.wolfram.com/groups/-/m/t/271204

katestange:
Also fun: http://localhost:5173/?name=Countdown&viz=Turtle&bgColor=22274e&speed=10&ruleMode=1&turnFormula=30%2B15*a&stepFormula=50&widthFormula=2&colorFormula=chroma%28%27skyblue%27%29.set%28%27hsl.h%27%2C+n%25360%29&seq=Formula&formula=%28n%5E2-n%2B2%29%2543

gwhitney:
> I'm trying to do things like 'chroma.scale()(0.5)' in the colour formula but it doesn't work.

If you look at the warning

`mathjs: value for n=0, a=1, f=0, b=0, x=0, y=0 set to 0 because of Unexpected type of argument in function multiplyScalar (expected: number or Complex or BigNumber or Fraction or Unit or string or boolean, actual: function, index: 0)`

you can get some clue as to what is going on: mathjs is parsing this formula, not entirely unreasonably, as multiplying the result of calling `chroma.scale()` by the number 0.5. (It does not know the types of the expressions at parse time, so it can't use the fact that chroma.scale() is a function to direct the parse; and in fact, until it executes the parsed expression, it has no way even of knowing that chroma.scale() returns a function because `chroma` is an object imported from the independent third-party package chroma.js; it is not part of mathjs.

you can work around this with various ridiculous exotic expressions that enforce the function call interpretation; one that works right now is `[to(arg,x)=arg(x), to(chroma.scale(),0.5)][2]` but this is clearly no kind of solution. It seems mathjs currently lacks a notation to enforce a function call interpretation of the expression you started with. I will open a design discussion in mathjs about this, but with Jos less able to respond, it will likely be quite some while before that resolves. In the meantime if you have a suggested notation, we could implement it just in numberscope. Any thoughts? I mean, we can certainly do something like `call(chroma.scale(), 0.5)` but that's maybe a bit cumbersome? What about `let(cs = chroma.scale(), cs(0.5))` -- is that any better? Or something like `chroma.scale()((0.5))` with the double parentheses to emphasize that it should be a function call not multiplication -- or is that too weird?

Incidentally, Jos has recovered enough to merge the bigint bugfixes, so I am returning for a bit to the effort to update mathjs  in numberscope to a version that supports bigints. So that will give some time to mull over this scale() operation.

katestange:
> you can work around this with various ridiculous exotic expressions that enforce the function call interpretation; one that works right now is `[to(arg,x)=arg(x), to(chroma.scale(),0.5)][2]` but this is clearly no kind of solution. It seems mathjs currently lacks a notation to enforce a function call interpretation of the expression you started with. I will open a design discussion in mathjs about this, but with Jos less able to respond, it will likely be quite some while before that resolves. In the meantime if you have a suggested notation, we could implement it just in numberscope. Any thoughts? I mean, we can certainly do something like `call(chroma.scale(), 0.5)` but that's maybe a bit cumbersome? What about `let(cs = chroma.scale(), cs(0.5))` -- is that any better? Or something like `chroma.scale()((0.5))` with the double parentheses to emphasize that it should be a function call not multiplication -- or is that too weird?

Thanks for the explanation!  Among the options here, I prefer the double parens, because space is at a premium in these little dialogs.  There's another thing that doesn't work but should, I believe:  `chroma.brewer.OrRd[1]`.  This is probably similar, so perhaps we need double square parens too?

gwhitney:
Actually, there's an even better alternative, I think, that doesn't yet work but that I am proposing for mathjs and that we could pre-implement here: `chroma.scale()(0.5,)` -- the presence of the comma making it clear (and unambiguously parseable) that that is an argument list.

I will check out the `chroma.brewer` situation.

katestange:
> Actually, there's an even better alternative, I think, that doesn't yet work but that I am proposing for mathjs and that we could pre-implement here: `chroma.scale()(0.5,)` -- the presence of the comma making it clear (and unambiguously parseable) that that is an argument list.

Oh that seems like a very nice solution.

katestange:
> I am fairly agnostic as to which way to go among these three, so if you have a preference, let me know.

I'm inclined to shorten things, if they have to change from the chroma docs anyway, so (2) is appealing in that sense.

But first, is it the same issue with `chroma.scale(['#fafa6e', '#2A4858']).mode('lch').colors(6)[2]` (also not working)?  The following two seem to work: `['#ddd', 'yellow', 'red', 'teal'][2]` and `chroma.scale('RdBu').colors(5)[2]`.  I'm just trying out various things from the docs.  Should this PR have unit tests for the various ways to use chroma through the mathjs formulas?  (The colours could be read as hex and compared to expected values?)  Especially if you can write up one or two example unit tests to show me how this works, I could do some thorough testing in this regard and add some such tests.

gwhitney:
> I'm inclined to shorten things, if they have to change from the chroma docs anyway, so (2) is appealing in that sense.

OK, I will add that to the TODOs.
 
> But first, is it the same issue with `chroma.scale(['#fafa6e', '#2A4858']).mode('lch').colors(6)[2]` (also not working)? 

Nope. That was because mathjs by default uses a fancy Matrix type to interpret the expression `[foo, bar]`. I've reconfigured it to use ordinary arrays and now this works fine.

> Should this PR have unit tests for the various ways to use chroma through the mathjs formulas?

It already does, just look at src/shared/__tests__/math.spec.ts. There are tests of direct use of chroma, but mostly of using it through the expression language. I did examples of lots of things from the docs, but I was not as thorough as you. Feel free to add as many tests as you like, including skipped tests for things you'd like to work but which don't yet. Thanks!

gwhitney:
P.S. I am again mostly preoccupied with mathjs bigint shortcomings so we definitely won't clash if you happen to add tests any time in the next couple days.

katestange:
> It already does, just look at `src/shared/__tests__/math.spec.ts`. There are tests of direct use of chroma, but mostly of using it through the expression language. I did examples of lots of things from the docs, but I was not as thorough as you. Feel free to add as many tests as you like, including skipped tests for things you'd like to work but which don't yet. Thanks!

Done!  I've added pretty much all the constructor types by working through the docs.  The thing you said was fixed doesn't pass, btw, not sure why.  But anyway there are a number of skipped tests there.

katestange:
Another thing I just came across.  Go to the following URL with no saved sequences:

http://localhost:5173/?name=Formula%3A+xgcd%283%2C+n%29%5B2%5D&viz=Turtle&speed=10&ruleMode=1&turnFormula=a%5E1.0012&stepFormula=10&colorFormula=green&seq=Formula&formula=xgcd%283%2Cn%29%5B2%5D

and then click to change the sequence.  Popups occur about complex numbers.  I'm guessing it's the turn formula that is a problem, combined with the xgcd and array call in math.js.  But these errors aren't getting caught, they are breaking through to pop-up.  If you change the xgcd formula to random (both positive and negative), no problem.  If you change the turn formula to reference n instead of a, no problem.

gwhitney:
OK, I've added these anomalies to the to-do list.

gwhitney:
> I'm guessing it's the turn formula that is a problem, combined with the xgcd and array call in math.js. But these errors aren't getting caught, they are breaking through to pop-up.

Indeed, it appears that the issue is that the sequence has a negative value, and then the turn formula is raising that to a a fractional power, and so getting a complex number. Then Turtle is trying to interpret that complex number as a turn angle, which it is unable to do. It is producing a popup rather than a formula warning on the parameter because the error is not occurring in the formula computation -- that is working perfectly well to produce a complex number -- but rather in the implementation code itself when it tries to convert the complex number to the JavaScript number type, which is impossible.

It's not clear what the best solution here is. The name of the function (that we added to mathjs) "safeNumber" seems to imply that it is always safe to call it. But what is it supposed to do when it is called with something like a matrix or a complex number that mathjs might return, for which there is simply no sensible way to convert the result to a number? I don't see what else it could do but throw an error.  So then we would still have to catch errors any time we call safeNumber, which is a pain.

So one possibility would be that for formula parameters we could supply a list of allowed return types (as a list of the names that mathjs uses for types) so that the parameter would go into an error state when it returned a disallowed type (and the code would get some value that would forestall further throws). This list could default to `[bigint]` or possibly `[number, bigint]`, so that you would have to explicitly bless any other type of return value that you might want to allow (like `Chroma`). What do you think of that option? Other thoughts on how to proceed here?

gwhitney:
OK all of the new tests now pass, except for the one that needs code to realize that `foo()(x,)` is a function call of the result of `foo()` on `x`. Note this has been done by just making `OrRd` etc valid pre-defined constants.

gwhitney:
> So one possibility would be that for formula parameters we could supply a list of allowed return types (as a list of the names that mathjs uses for types) so that the parameter would go into an error state when it returned a disallowed type (and the code would get some value that would forestall further throws). This list could default to `[bigint]` or possibly `[number, bigint]`, so that you would have to explicitly bless any other type of return value that you might want to allow (like `Chroma`). What do you think of that option? Other thoughts on how to proceed here?

OK, this has percolated and I now have a concrete proposal. MathFormula objects should have a preferred return type, defaulting to bigint (because that's what a Formula Sequence wants), but it can be any type supported by mathjs.  The compute infrastructure will attempt to convert to your preferred type for you, so that if the conversion fails, it can return a dummy value and post an error to the status for that formula parameter.

How's that sound? I think it provides ease for the visualizer writer while being comprehensible for the sequence explorer -- if the latter puts in a formula that produces an unusable value like a complex number, they will get a message to that effect in the parameter box and can adjust their formula.

gwhitney:
> This produces a disconnected path.

Right, there was a bug that prevented steps of negative length, which are perfectly sensible, from actually being drawn. Fixed now.

gwhitney:
OK, it turned out there was literally no way to get an event to reliably fire across browsers when you close the built-in color picker, if the color picked hasn't changed, which mean that with the built-in picker, there was no way to fulfill our goal of re-selecting the same color in the chooser inserting it into the stroke color parameter.

So in the latest commit I switched to a 3rd party color picker. That one, it turns out, has a built-in method for selecting a sequence of colors, so I enabled that for the Stroke color(s) parameter. So now there is no need to mess with text representations of colors in the Turtle List mode at all. And therefore the color chooser only appears, right below the Color formula, in formula mode. (The background color button is always active.) Moreover, the chosen color replaces whatever part of the formula you had selected prior to making the choice, making it easy to replace one color in your formula with another.

Hopefully this is pleasant and easy to use. I will continue to plow through the many points here.

gwhitney:
> * Allow descriptions to have links and make the word "color" in the above  description be a link to our documentation page/section on color specifications.

Now that the color list is not using text color notations, I recommend we postpone this and not do it as part of this PR. Links in descriptions basically means switching to supporting Markdown descriptions (don't clearly see any other way, since arbitrary HTML is not allowed in the text of fields for security reasons), and that is a larger effort, seems beyond the scope of this PR. If you're comfortable turning on Formula mode, I think we can assume that you are a Numberscope guru or at least know how to find the doc pages...

gwhitney:
> * Conform the variable names to n - sequence index; a - sequence value; s - step serial number starting from 1; f,l - first and last sequence index; x, y -- coordinates; b - bearing.

Oops, unfortunately this list overlooked that `f` needs to be used for the frame number. So I am going to implement:

- n: sequence index
- a: sequence value
- s: step serial number
- m: minimum sequence index (so n = m + s - 1)
- M: Maximum sequence index
- f: frame number
- x,y: coordinates of current position
- b: bearing

Any objections/better suggestions, let me know.

gwhitney:
> * "Preimplement" `chroma.scale()(0.5,)` to unambiguously mean that the function `chroma.scale()` should be called on `0.5` to avoid the implicit multiplication trap for trying to use this color formula.

> * Get skipped tests to pass

I have filed https://github.com/josdejong/mathjs/pull/3378 to support this notation in mathjs. The "preimplementation" idea looks pretty tricky: it's a small change inside of mathjs's parser, but the parser is open to customization from outside the package only in very specific ways. So it seems as though we would need to basically have a copy of the mathjs parsing code within frontscope in order to "preimplement" this, and it would be temporary until mathjs adopts this or some other notation for allowing function-returning expressions to be called. So my recommendation is to wait on these items: not make them part of this PR, and once a mathjs update is merged that has this function-calling feature, revisit the skipped tests (which now only deal with this point).

gwhitney:
I like the Vertigo of Divergence and am in the process of adding it as a featured specimen.  

On the other hand, I don't see how your second example reflects the convergence of the first nontrivial zeta zero. If we let $s_1 = \frac12 + \rho_1i \approx \frac12 + 14.135i$, and ignoring the radian -> degree conversion (for which in the future I recommend `360/tau` as the clearest factor), you have the turn formula $\rho_1 \log n$ and the step formula $200/\sqrt n$. As these expressions are the argument and constant-scaled radius of $n^{-s_1}$, I think you are attempting to do direct summation of the original defining sum of the zeta function by having each turtle move represent the addition of the next complex number. But to do that summation, we want the _bearing_ of each turtle move to be the argument of the corresponding term in the zeta-summation, not the _turn angle_ (= bearing difference) to be that argument.

Fortunately, that's easy to do: subtracting `b`, the current bearing, from any expression `EXPR` to create a turn angle formula will make `EXPR` become the new bearing. But now changing the turn formula to `log(n)*14.135*360/tau - b` produces a not-very-interesting picture: it jumps around for the first half-dozen steps haphazardly, and then settles into a steady outward spiral. But that shouldn't be surprising, in two ways:
(a) the original defining summation _diverges_ for Re z ≤ 1.
(b) this turn angle could just as well be written (ignoring unit conversion) as  `log(n)*14.135 - log(n-1)*14.135` , i.e. $\rho_1\log(1 - 1/n)$, which is getting smaller and smaller. So the turtle is turning less and less, and so of course it is spiraling outwards.

So what I think _is_ interesting -- definitely something I didn't know before -- is that for larger imaginary parts, at least near roots of zeta, the direct summation of the terms in the series, before it starts its final "death spiral to infinity", comes _very_ near to crossing right over the actual value of its analytic continuation. Moreover, one can easily compute the exact pair of terms such that the midpoint of the line segment between them comes closest to the value of zeta. And you get a pretty picture of what's going on. These statements are the gist of https://math.stackexchange.com/questions/805396/simpler-zeta-zeros that you referenced.

So for example http://localhost:5173/?name=Convergence+of+Riemann+Zero+%2310143&viz=Turtle&speed=10&ruleMode=1&turnFormula=log%28n%29*10000.065345417*360%2Ftau+-b&stepFormula=200%2Fsqrt%28n%29&colorFormula=rainbow%282b%29&seq=Formula&first=1&last=3183&length=3183 reproduces exactly the diagram of that math stackexchange article (prettier, though, I think ;-). So we could include that as a featured visualization.

There are lots of things to notice here. A few are:
* The path beautifully comes back to visually precisely where it starts, precisely because this $\rho_{10143}$ is the imaginary part of a root of zeta. In fact, if you increase the scale factor to 200,000 from 200, the very last stroke still goes through exactly the starting pixel! (That's how good this cutoff of the summation is at estimating the true zeta function.) So we could instead have a featured specimen in which we animate that scale factor, zooming in on the beautiful structure at the origin. A step formula of `200*1.03^f/sqrt(n)` works well for this.
* Alternatively, if you swap in the previous root $\rho_{10142}\approx 9998.850397090$ in the turn formula, you get a trajectory which shares the qualitative features that it consists of a bunch of whorls connected by s-curves and lands back at 0, but which takes a completely different path in detail to get there. So we could animate the imaginary part of s starting from this root, and you get a compelling picture of the enormous numerical coincidences that must occur for zeta to have a root (in fact it makes it seem unfathomable that zeta in fact has any nontrivial roots). And _that_ could be the featured specimen. A turn formula of `log(n)*(9998.850397090+0.007f)*360/tau - b` works well for this.
* One could instead animate the real part of s with something like `200/n^(0.35+0.001f)` but that's less visually interesting because it mostly looks like scaling. While it makes it seem clear that there will only be one root as the real part changes, because the center of the final whorl is moving close to linearly, I wouldn't recommend this animation as a featured specimen.

Looking forward to your thoughts/recommendations. One other thing to note is that the _alternating_ sum of inverse s-th powers of n, with the proper scaling factor, does converge in the usual sense to $\zeta(s)$. However, this produces a less interesting picture because it is a sum that in fact converges very well. So you just get a sort of star-shaped spiral that homes in very quickly on the origin, even for the first root. I haven't tried animating the imaginary part of this summation, though -- that might be interesting in terms of seeing the trajectory of zeta, starting from a smallish imaginary part and reaching the first few roots. Let me know if you think that's worth looking at.

Oh and as far as the third example it was indeed cute but I don't think it's quite "featured specimen" caliber. But I did find the following related specimen that I think could be a candidate:
http://localhost:5173/?name=PEMDASymmetry&viz=Turtle&bgColor=22274e&speed=10&ruleMode=1&turnFormula=30%2B15a&stepFormula=4&colorFormula=rainbow(abs(y-160))&seq=Formula&formula=(n%2F10)^2+-+n%2F10^2&last=8191&length=8192

Vectornaut:
> Should Turtle have a "Bearing/Turn" toggle to tell you whether the angle formula gives you the absolute or relative bearing?

Yeah, I think that would be useful, because the OEIS includes at least one sequence that lists the turtle bearings for a curve. Sequence [A229214](https://oeis.org/A229214) gives the bearings for the flowsnake.

gwhitney:
> Yeah, I think that would be useful, because the OEIS includes at least one sequence that lists the turtle bearings for a curve. Sequence [A229214](https://oeis.org/A229214) gives the bearings for the flowsnake. It would also be great for drawing cutting sequences on the torus (also known as Sturmian sequences)!

OK, in progress. When a commit for it is pushed here, I will also post a link for the flowsnake specimen in the [interesting specimens](https://github.com/numberscope/frontscope/discussions/477) discussion.

gwhitney:
@katestange @Vectornaut I am stumped by "TouchEvent is not defined". I cannot reproduce on any browser. So I definitely need more info. I am using Firefox 134.0.2, what version do you have? There are definitely reports of this error online, always associated with either Firefox or Safari, but mostly years old. There is one software package I found (https://github.com/grafana/grafana/issues/78431) that had an issue with this fixed only a year ago, so it could be a thing still. But the funny thing ist neither vue-pick-colors (the component I added) or its sole dependency `@popperjs/core` even contain the strings 'Touch' or 'touch' in their entire package contents. So I guess we should start with  your Firefox version(s), an exact reproducing sequence of actions, and the backtrace it is presumably producing in the console either just before or just after popping up that error window, and then take it from there. Thanks so much for lending a hand with this.

Vectornaut:
> So I guess we should start with your Firefox version(s), an exact reproducing sequence of actions, and the backtrace it is presumably producing in the console either just before or just after popping up that error window, and then take it from there.

#### Browser

Firefox 135.0

#### Reproduction steps

1. Launch Numberscope (commit 23c1c4a) with `npm run dev`
2. Go to [`http://localhost:5173/?name=OEIS+A000045&viz=ModFill&seq=OEIS+A000045`](http://localhost:5173/?name=OEIS+A000045&viz=ModFill&seq=OEIS+A000045)
5. Click the black swatch next to the Mod Fill visualizer's *Fill color* control
   * *A color picker appears*
6. Click any point on the constant-hue slice in the color picker
   * *The error dialog appears*

#### Console messages

When I trigger the error, the following log messages appear.

```
Numberscope encountered error: ReferenceError: TouchEvent is not defined
    onSelect index.esm.js:1
    G.render/t[0]< index.esm.js:1
    cacheKey runtime-dom.esm-bundler.js:1705
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    invoker runtime-dom.esm-bundler.js:729
    addEventListener runtime-dom.esm-bundler.js:680
    patchEvent runtime-dom.esm-bundler.js:698
    patchProp runtime-dom.esm-bundler.js:775
    mountElement runtime-core.esm-bundler.js:4873
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    componentUpdateFn runtime-core.esm-bundler.js:5326
    run reactivity.esm-bundler.js:225
    setupRenderEffect runtime-core.esm-bundler.js:5454
    mountComponent runtime-core.esm-bundler.js:5229
    processComponent runtime-core.esm-bundler.js:5182
    patch runtime-core.esm-bundler.js:4700
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    componentUpdateFn runtime-core.esm-bundler.js:5326
    run reactivity.esm-bundler.js:225
    setupRenderEffect runtime-core.esm-bundler.js:5454
    mountComponent runtime-core.esm-bundler.js:5229
    processComponent runtime-core.esm-bundler.js:5182
    patch runtime-core.esm-bundler.js:4700
    componentUpdateFn runtime-core.esm-bundler.js:5406
    run reactivity.esm-bundler.js:225
    runIfDirty reactivity.esm-bundler.js:263
    callWithErrorHandling runtime-core.esm-bundler.js:199
    flushJobs runtime-core.esm-bundler.js:408
    promise callback*queueFlush runtime-core.esm-bundler.js:322
    queueJob runtime-core.esm-bundler.js:317
    scheduler runtime-core.esm-bundler.js:5448
    trigger reactivity.esm-bundler.js:253
    endBatch reactivity.esm-bundler.js:311
    notify reactivity.esm-bundler.js:597
    trigger reactivity.esm-bundler.js:571
    set value reactivity.esm-bundler.js:1448
    togglePicker ParamField.vue:177
    maybeTogglePicker ParamField.vue:172
    patchStopImmediatePropagation runtime-dom.esm-bundler.js:758
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:217
    invoker runtime-dom.esm-bundler.js:729
main.ts:11:12
    errorHandler main.ts:11
    callWithErrorHandling runtime-core.esm-bundler.js:199
    handleError runtime-core.esm-bundler.js:246
    callWithErrorHandling runtime-core.esm-bundler.js:201
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    invoker runtime-dom.esm-bundler.js:729
    (Async: EventListener.handleEvent)
    addEventListener runtime-dom.esm-bundler.js:680
    patchEvent runtime-dom.esm-bundler.js:698
    patchProp runtime-dom.esm-bundler.js:775
    mountElement runtime-core.esm-bundler.js:4873
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    componentUpdateFn runtime-core.esm-bundler.js:5326
    run reactivity.esm-bundler.js:225
    setupRenderEffect runtime-core.esm-bundler.js:5454
    mountComponent runtime-core.esm-bundler.js:5229
    processComponent runtime-core.esm-bundler.js:5182
    patch runtime-core.esm-bundler.js:4700
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    mountChildren runtime-core.esm-bundler.js:4932
    mountElement runtime-core.esm-bundler.js:4855
    processElement runtime-core.esm-bundler.js:4820
    patch runtime-core.esm-bundler.js:4688
    componentUpdateFn runtime-core.esm-bundler.js:5326
    run reactivity.esm-bundler.js:225
    setupRenderEffect runtime-core.esm-bundler.js:5454
    mountComponent runtime-core.esm-bundler.js:5229
    processComponent runtime-core.esm-bundler.js:5182
    patch runtime-core.esm-bundler.js:4700
    componentUpdateFn runtime-core.esm-bundler.js:5406
    run reactivity.esm-bundler.js:225
    runIfDirty reactivity.esm-bundler.js:263
    callWithErrorHandling runtime-core.esm-bundler.js:199
    flushJobs runtime-core.esm-bundler.js:408
    (Async: promise callback)
    queueFlush runtime-core.esm-bundler.js:322
    queueJob runtime-core.esm-bundler.js:317
    scheduler runtime-core.esm-bundler.js:5448
    trigger reactivity.esm-bundler.js:253
    endBatch reactivity.esm-bundler.js:311
    notify reactivity.esm-bundler.js:597
    trigger reactivity.esm-bundler.js:571
    set value reactivity.esm-bundler.js:1448
    togglePicker ParamField.vue:177
    maybeTogglePicker ParamField.vue:172
    patchStopImmediatePropagation runtime-dom.esm-bundler.js:758
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:217
    invoker runtime-dom.esm-bundler.js:729
```
```
In component: 
Proxy { <target>: {…}, <handler>: {…} }
main.ts:12:12
```
```
Additional information: native event handler main.ts:13:12
```

gwhitney:
OK, @katestange or @Vectornaut, could one of you try the following in this branch:

1) Do `npm run preview` and visit the URL that produces and verify the problem is present.
2) Replace `node_modules/vue-pick-colors/dist/index.esm.js` with the version attached to this comment. (After all this, you can `rm -r node_modules/vue-pick-colors` and then `npm install` again to revert to the standard vue-pick-colors, but the differences are tiny so it shouldn't really make any difference whether you do or not.)
3) Run `npm run build:clean`
4) Run `npm run preview` and verify that the problem is no longer occurring.

Sorry about the slightly circuitous route via preview, but I am not clear on how to be certain that `npm run dev` will pick up a hot-patched dependency, whereas a clean build is obliged to go and get whatever is there in your node_modules to package up a self-contained distro.

Thanks! If this works I will submit a PR with the tiny fixes (that just check if TouchEvent support is enabled before trying to use it; turns out many desktop Firefox distros don't turn on the support and that's what's biting us).

gwhitney:
> * Implement a "preferred type" scheme for Formula parameters, where the infrastructure does the conversion for you and warns in the UI if not possible.

OK, this isn't really feasible before updating to the new mathjs. It also isn't _really_ germane to this PR; it's just that Kate happened to run into some complex numbers being generated from a negative input to a fractional power while messing around with this PR. So I am going to file this as a separate issue for alpha, so it can be done within or after the mathjs update PR, and take it off the list for this one.

gwhitney:
> * Degree/radian

Oh! It occurs to me that mathjs already has angle units: deg and rad and even the obscure grad (400 grad = τ radians). So I recommend that in formula mode, angles are angles and they should either have a unit, or if they are left unitless, are interpreted as radians. So I think there is no need for a unit toggle (and I prefer not to include one to keep the interface simpler, but I will put a mention in the docs of the units). On the other hand, in List mode, to keep things accessible to newcomers, I suggest we continue to interpret the typed-in numbers as degrees.

gwhitney:
> > * Degree/radian
> 
> Oh! It occurs to me that mathjs already has angle units: deg and rad and even the obscure grad (400 grad = τ radians). So I recommend that in formula mode, angles are angles and they should either have a unit, or if they are left unitless, are interpreted as radians. So I think there is no need for a unit toggle (and I prefer not to include one to keep the interface simpler, but I will put a mention in the docs of the units). On the other hand, in List mode, to keep things accessible to newcomers, I suggest we continue to interpret the typed-in numbers as degrees.

Oops. Using mathjs' built-in units _would_ be a nice solution, except that its Unit class currently eagerly converts everything to radians internally, meaning that 180 deg + 180 deg is not necessarily precisely 360 deg. (Since internally it is trying to add pi to itself to get exactly tau; that one operation will be off by no more than a double ulp, but such errors can compound, throwing off any attempt to do turtle paths in integer degrees that crucially take advantage of arithmetic mod 360.) So I have filed a [mathjs issue](https://github.com/josdejong/mathjs/issues/3381), and in the meantime we should have a degrees/radians selector. We can remove the selector in favor of specifying units in the formula if/when the mathjs issue is satisfactorily resolved.

gwhitney:
OK, all that remains is:
- [ ] Kate finalizing the new featured specimens for this PR
- [ ] Kate or Vectornaut verifying the vue-pick-colors fix I posted above and Glen filing that upstream,
- [ ] Glen updating all of the snapshots/making sure all tests pass
- [ ] Final review.

(This checklist supersedes the one posted much nearer the top of the conversation.) Looking forward to wrapping this up.

gwhitney:
In the [Interesting Specimens] discussion, Kate writes:

> Folding rates, under bearing setting, has a bit different effect; maybe the description should be updated slightly to indicate it is changing the angle or bearing, as appropriate.

Well, it says that the folding rates change the Angles above, which they do. If the Angles happen to be Bearings, then it changes those Bearings... So I'm not really sure what else I would say. But you should feel free to edit any of the documentation in ways that you feel would make it clearer!

gwhitney:
Kate also writes:

> The Angle field is a little flaky. Delete out everything and it populates NaN and you have to delete those three characters to get an empty field, but then that stays empty.

Have reproduced. Will debug/fix, and then look at other comments.

gwhitney:
OK, that one should be taken care of.

gwhitney:
Kate also writes:
> More annoying and odd, if you type 0 30 -90 10 and then arrow key over to sit immediately after the -90 and backspace twice, the cursor pops to the end and the - disappears. You might have wanted to backspace the 90 to put an 80... it won't behave the way you expect. This behaviour is, however, intermittent! It did this for me (repeatedly) in chrome and then when I tabbed over here to firefox and back, it stopped doing it quite like that -- now it would only pop to the end when deleting the - also.

OK, I cannot reproduce jumping after two backspaces (so just the 90 disappears, leaving a bare - sign). In multiple browsers and tries, I consistently  just get a red error message below the entry field: `Input must be a list of numbers separated by whitespace or commas`, which seems fine.

I am, however, reproducing the behavior that after one more backspace, the list becomes `0 30 10` as expected, _but_ the cursor jumps to the end of the field. I am worried that will be tricky to track down, but I agree it is irritating so I will give it a shot.

gwhitney:
Oh! It is because changing the angle list (when it actually changes the value as a list) means the Turtle has to update the angleFormula (since that's what really drives the turtle) and then it has to refreshParams and so it writes the canonicalized version of the angle list back into the text field, disrupting the insertion point... I will have to make a mechanism for controlling _which_ parameters get refreshed, and not refresh the ones that were just changed by the person exploring...

gwhitney:
OK, that one should be fixed now, too.

gwhitney:
Kate also writes:
> If you get NaN in turn angles and don't notice, but then go to "Formula" mode (maybe that's all you wanted anyway), you then have an error (from the NaN) sitting up top of the visualizer panel and nothing works right (colouring doesn't work, canvas turns back to default colour), but the offending field is hidden. Not sure what to suggest here, but it did cause me confusion for a bit (particularly the canvas turning colour). Should the error checking on hidden "List" variables be turned off when they are hidden?

OK, it's now harder to create this error, because literally typing in a NaN gets caught by the validation and doesn't update the parameter (no up-top "error in angles" shows up even). And if while you are in this error state, you switch to Formula mode, the Turtle works normally. However, if you switch back to List, the erroneous angles parameter value now does actually get transferred into the Turtle, so to speak, and the up-top red box shows up. And if you _now_ just ignore or don't notice the error and switch yet again to Formula, you are in Kate's bad state.

Like Kate, I am not 100% sure how to handle this. I do not think that it would work to just turn error checking off for hidden parameters, because there is no guarantee that hidden parameters don't control the behavior of the visualization, and in fact, in some cases they do -- they are just "advanced" or "detailed" parameters that you want to hide by default to avoid clutter. And the visualizer code relies on never getting values that fail validation. Hence, can't turn off validation for hidden parameters.

I can think of three possibilities:
(A) Don't sweat it, it's now pretty intricate to invoke this bad state and you have to switch modes ignoring errors twice for it to bite you...
(B) Don't allow parameters in-error to be hidden (i.e., if the status of an individual parameter is invalid, ignore its hiding parameter and always show it).
(C) Specifically for Turtle, when rule mode switches to list, any in-error parameters are set back to their last-known good values or the defaults if we can't figure out a prior good value. This switching parameters is OK because we already have a discontinuity in behavior when you switch back to list, since not all Formulas can be captured by List mode (and so we don't even try to update List as we change Formula, as opposed to vice versa, which we do).

What do you think? Other ideas? I think I like (B) best because it seems more helpful than (A) and more general and actually simpler to implement than (C) and should take some burden off of future visualizer writers. Let me know.

katestange:
This demonstrates a way of drawing a graph.  http://localhost:5173/?name=Mobius+Heartbeat&viz=Turtle&domain=0+1+-1&angles=0+89+-89&angleMeaning=1&steps=0+10+10&strokeColor=%23E9F4F2+%23DC967E+%237FB7ED&bgColor=080908&speed=20&angleFormula=%7B%220%22%3A+0%2C+%221%22%3A+89%2C+%22-1%22%3A+-89%7D%5Bstring%28a%29%5D&stepFormula=%7B%220%22%3A+0%2C+%221%22%3A+10%2C+%22-1%22%3A+10%7D%5Bstring%28a%29%5D&widthFormula=%7B%220%22%3A+1%2C+%221%22%3A+1%2C+%22-1%22%3A+1%7D%5Bstring%28a%29%5D&colorFormula=%7B%220%22%3A+%23E9F4F2%2C+%221%22%3A+%23DC967E%2C+%22-1%22%3A+%237FB7ED%7D%5Bstring%28a%29%5D&colorChooser=813d9c&seq=OEIS+A008683

katestange:
Sorry I seem to have been adding all my comments on this PR into the discussion instead, by going to wrong window, sorry!

katestange:
> What do you think? Other ideas? I think I like (B) best because it seems more helpful than (A) and more general and actually simpler to implement than (C) and should take some burden off of future visualizer writers. Let me know.

Yeah, (B) seems best, but I'm ok with (A) if you want.

katestange:
I'll just re-add this here since I think it's a good option for the specimens and it might get lost in that other discussion.  A possible contender for gallery: http://localhost:5173/?name=The+Raucous+Recaman&viz=Turtle&domain=0+1&steps=2+3&strokeColor=%235e8d85&bgColor=E5E0F6&speed=20&ruleMode=1&angleFormula=89.5&stepFormula=a%2F30&colorFormula=rainbow%28n%29&colorChooser=813d9c&seq=OEIS+A005132

gwhitney:
> I'll just re-add this here since I think it's a good option for the specimens and it might get lost in that other discussion. A possible contender for gallery: http://localhost:5173/?name=The+Raucous+Recaman&viz=Turtle&domain=0+1&steps=2+3&strokeColor=%235e8d85&bgColor=E5E0F6&speed=20&ruleMode=1&angleFormula=89.5&stepFormula=a%2F30&colorFormula=rainbow%28n%29&colorChooser=813d9c&seq=OEIS+A005132

A few points:

(a) Aaron and I both think you're the Featured Specimens curator, so feel free to add/edit the featured ones or let me know what changes/additions you want :-)

(b) We should definitely hang on to this whether or not it goes into Featured Specimens because I think it will be a good one for making our case for integrating more tightly with OEIS, as Neil _loves_ the Recaman  sequence.

(c) Have you shown Edmund? I think he devised the "loop de loop above and below the number line with semicircles" visualization of the Recaman sequence.

(d) If you want to see all 100000 terms (even by zooming out a long way -- eventually the strokes disappear), you either need to use a much larger denominator in `a/100` or use a sublinear function; I like how `a^(2/3)/2` looks.

gwhitney:
> This demonstrates a way of drawing a graph. http://localhost:5173/?name=Mobius+Heartbeat&viz=Turtle&domain=0+1+-1&angles=0+89+-89&angleMeaning=1&steps=0+10+10&strokeColor=%23E9F4F2+%23DC967E+%237FB7ED&bgColor=080908&speed=20&angleFormula=%7B%220%22%3A+0%2C+%221%22%3A+89%2C+%22-1%22%3A+-89%7D%5Bstring%28a%29%5D&stepFormula=%7B%220%22%3A+0%2C+%221%22%3A+10%2C+%22-1%22%3A+10%7D%5Bstring%28a%29%5D&widthFormula=%7B%220%22%3A+1%2C+%221%22%3A+1%2C+%22-1%22%3A+1%7D%5Bstring%28a%29%5D&colorFormula=%7B%220%22%3A+%23E9F4F2%2C+%221%22%3A+%23DC967E%2C+%22-1%22%3A+%237FB7ED%7D%5Bstring%28a%29%5D&colorChooser=813d9c&seq=OEIS+A008683

Can you post a version of this that uses a sequence that's on the server? I can't test sequences in Github until it's available on the standard server?

gwhitney:
> Yeah, (B) seems best, but I'm ok with (A) if you want.

OK, I will implement (B) when I can.

gwhitney:
Kate writes:
> This doesn't produce anything... but it fails silently: http://localhost:5173/?name=OEIS+A000521&viz=Turtle&domain=0+1&steps=2+3&strokeColor=%235e8d85&bgColor=000000&speed=100&ruleMode=1&angleFormula=a+%25+30&stepFormula=3&widthFormula=6&colorFormula=chroma%28%7Bh%3A+%28n%2F1.5%29+%25+360%2C+s%3A0.65%2C+l%3A+0.75%7D%29&seq=OEIS+A000521

For me, it produces a popup that says `Error Message:
RangeError: Attempt to use number -1152921504606846800 as a number`, which seems like acceptable behavior. Try clicking on this link again, does it really not produce a popup?

gwhitney:
Kate also says:
> By contrast, this produces something... http://localhost:5173/?name=OEIS+A000521&viz=Turtle&domain=0+1&steps=2+3&strokeColor=%235e8d85&bgColor=000000&speed=100&ruleMode=1&angleFormula=a+%25+30+%3D%3D+0+%3A+30+%3A+0&stepFormula=3&widthFormula=6&colorFormula=chroma%28%7Bh%3A+%28n%2F1.5%29+%25+360%2C+s%3A0.65%2C+l%3A+0.75%7D%29&seq=OEIS+A000521
The only difference is in the angle formula. Bigint issue? I think once the coefficients become bigints, the comparison always fails (we get a straight line). The picture doesn't seem right.

I get the straight line shooting off. I will investigate when I can.

gwhitney:
> Kate also says:
> 
> > By contrast, this produces something... http://localhost:5173/?name=OEIS+A000521&viz=Turtle&domain=0+1&steps=2+3&strokeColor=%235e8d85&bgColor=000000&speed=100&ruleMode=1&angleFormula=a+%25+30+%3D%3D+0+%3A+30+%3A+0&stepFormula=3&widthFormula=6&colorFormula=chroma%28%7Bh%3A+%28n%2F1.5%29+%25+360%2C+s%3A0.65%2C+l%3A+0.75%7D%29&seq=OEIS+A000521
> > The only difference is in the angle formula. Bigint issue? I think once the coefficients become bigints, the comparison always fails (we get a straight line). The picture doesn't seem right.
> 
> I get the straight line shooting off. I will investigate when I can.

Right. This is running on a version of mathjs that does not yet have bigint support. (That switchover is still held up by outstanding PRs to fix bugs in the bigint version.) So there's no way it can accurately compute `a % 30` once a exceeds MAX_SAFE_INTEGER. So this example will have to be deferred until after bigint support is in.

gwhitney:
> > Yeah, (B) seems best, but I'm ok with (A) if you want.
> 
> OK, I will implement (B) when I can.

Alright, this is in. But I am slightly uneasy: if a parameter is being displayed only because it is invalid, and then you go and fix it (because of course you need to), as soon as it is no longer invalid it - boom - disappears. It's kind of jarring. So please try it and see what you think and if you have any suggestions.

gwhitney:
> This demonstrates a way of drawing a graph

Never mind, I have it running on the local backscope, and I will shortly be installing that on the server. Very clever on the graph! Were you thinking of that as a featured specimen? It's also fun to make the step length for 0 be some tiny value like 0.3. (And then if you make the down steps just slightly shorter than the up ones, like 9.8, it looks almost exactly like a stock chart!)

katestange:
BTW, regarding The Raucous Recaman, it turns out those settings are lovely for lots of sequences.  It's pretty nice to see the thumbnail gallery.  That's a good property to have in gallery specimens, because then they are good starting points.

gwhitney:
I mean, I totally agree with your judgment that the Raucous Recaman should be a featured specimen, but what exact parameters should be put in? As it stands now, the step length needs to be tuned on a sequence-to-sequence basis for the optimal pictures (and in fact as you may have seen, I recommended sublinear growth to look at all of Recaman.  Should we try to make something adaptive so that the gallery will look as good as possible? E.g., constant, plus a linear term up to some cutoff, patched together with a scaled log that takes over at that point? Or is that too crazy? 

Feel free to push a commit adding it or let me know what you think I should do as far as Raucous Recaman in the featured sequences. Thanks!

katestange:
> Try clicking on this link again, does it really not produce a popup?

You're right, I get a popup.  No worries then.

katestange:
> So this example will have to be deferred until after bigint support is in.

Yup no worries.

katestange:
> 2\. Replace `node_modules/vue-pick-colors/dist/index.esm.js` with the version attached to this comment.

I don't see the attached file?

katestange:
> Alright, this is in. But I am slightly uneasy: if a parameter is being displayed only because it is invalid, and then you go and fix it (because of course you need to), as soon as it is no longer invalid it - boom - disappears. It's kind of jarring. So please try it and see what you think and if you have any suggestions.

Hmm, yes, jarring, maybe that's not a good solution after all.  I'm ok with (A) or (C), or (C) with just default values.  As you say in (A), it's a fairly corner-case behaviour, so as long as it doesn't get in some kind of inconsistent state, the user can probably figure it out...

gwhitney:
> > Alright, this is in. But I am slightly uneasy: if a parameter is being displayed only because it is invalid, and then you go and fix it (because of course you need to), as soon as it is no longer invalid it - boom - disappears. It's kind of jarring. So please try it and see what you think and if you have any suggestions.
> 
> Hmm, yes, jarring, maybe that's not a good solution after all. I'm ok with (A) or (C), or (C) with just default values. As you say in (A), it's a fairly corner-case behaviour, so as long as it doesn't get in some kind of inconsistent state, the user can probably figure it out...

OK I am going to try setting invalid lists to their default values when rule mode is selected and see how that works.

gwhitney:
> > 2. Replace `node_modules/vue-pick-colors/dist/index.esm.js` with the version attached to this comment.
> 
> I don't see the attached file?

Hmm I don't see it either. And I tried again to attach it to this post, and GitHub it seems rejects such attachments. So I will try emailing it to both of you.

gwhitney:
> > > 2. Replace `node_modules/vue-pick-colors/dist/index.esm.js` with the version attached to this comment.
> > 
> > 
> > I don't see the attached file?
> 
> Hmm I don't see it either. And I tried again to attach it to this post, and GitHub it seems rejects such attachments. So I will try emailing it to both of you.

Nope, it bounced from both of your email accounts. Hmm. I will put it on a server and email you both a link

katestange:
> 1. Do `npm run preview` and visit the URL that produces and verify the problem is present.
> 
>     2. Replace `node_modules/vue-pick-colors/dist/index.esm.js` with the version attached to this comment. (After all this, you can `rm -r node_modules/vue-pick-colors` and then `npm install` again to revert to the standard vue-pick-colors, but the differences are tiny so it shouldn't really make any difference whether you do or not.)
> 
>     3. Run `npm run build:clean`
> 
>     4. Run `npm run preview` and verify that the problem is no longer occurring.
> 
> 
> Sorry about the slightly circuitous route via preview, but I am not clear on how to be certain that `npm run dev` will pick up a hot-patched dependency, whereas a clean build is obliged to go and get whatever is there in your node_modules to package up a self-contained distro.

Ok, I replaced the file, but `npm` doesn't have a `build:clean` script?  Is there an argument I should pass?

gwhitney:
Huh, you're right. I guess I've been switching between too many different projects. Just execute `rm -r dist` in the top-level directory of your frontscope clone instead of `npm run build:clean`. Thanks.

katestange:
> Huh, you're right. I guess I've been switching between too many different projects. Just execute `rm -r dist` in the top-level directory of your frontscope clone instead of `npm run build:clean`. Thanks.

Ok, it works!!  The new file fixes it.

gwhitney:
Great. Have filed https://github.com/qiuzongyuan/vue-pick-colors/pull/38

gwhitney:
> > > Alright, this is in. But I am slightly uneasy: if a parameter is being displayed only because it is invalid, and then you go and fix it (because of course you need to), as soon as it is no longer invalid it - boom - disappears. It's kind of jarring. So please try it and see what you think and if you have any suggestions.
> > 
> > 
> > Hmm, yes, jarring, maybe that's not a good solution after all. I'm ok with (A) or (C), or (C) with just default values. As you say in (A), it's a fairly corner-case behaviour, so as long as it doesn't get in some kind of inconsistent state, the user can probably figure it out...
> 
> OK I am going to try setting invalid lists to their default values when rule mode is selected and see how that works.

I have just pushed a commit in which Turtle resets invalid list rule params whenever you switch the rule mode to formula. Please give that a try in the scenario in which you mess up one of the lists and then switch to rule mode and switch back and let me know if it seems like OK behavior now.

gwhitney:
If it is OK, then all that remains before this PR is ready for final review is for @katestange to make the final call on new featured specimens for this PR (it's totally OK if you ditch 'PEMDASymmetry,' Kate) and then for Glen to update the snapshots and make sure all tests pass.

gwhitney:
> Sorry to bring up

No, that's exactly what review is for! Reproduced, at least roughly (for me, the first click on Background color works, and thereafter only every other one). Looking into it.

gwhitney:
OK, there is definitely some weirdo interaction going on between the color picker click events and the changing of the state of whether it is showing or not. (The whole problem arises because the default behavior of vue-pick-colors is that when it is open, to close it and select the color you are obliged to click elsewhere, whereas it seems clear to me that if it is open and you click again on the original color swatch, it should close and select the color. So to achieve that, we also have to track the clicks on the color picker, and track its value changes, and try to keep track of whether it is currently popped up or not... complicated.)

Anyhow, I think I have it ironed out in the latest commit, even though it's much more complicated than it seems like it ought to be. Let me know if it works for you. Then when you have the time to settle the featured specimens, great. Just let me know what else I can do.

katestange:
So I've curated the featured specimens, but I notice the last commit is failing unit tests, so I won't commit something on top unless you want me to?

gwhitney:
`npm run test:unit` is working for me in the latest commit. Does it really fail for you? In any case, if you are only changing defineFeatured, that won't affect any unit tests, only end-to-end tests, so I am already planning on working on those after you check it in anyway. So go ahead and commit and push, as far as I am concerned.

katestange:
> `npm run test:unit` is working for me in the latest commit. Does it really fail for you? In any case, if you are only changing defineFeatured, that won't affect any unit tests, only end-to-end tests, so I am already planning on working on those after you check it in anyway. So go ahead and commit and push, as far as I am concerned.

It's also failing on github.  And lint is failing also.  In any case, I'll go ahead and commit and push in the meanwhile.

gwhitney:
Oh, I bet it is the Node version. What does `node --version` say for you? I am at v22.13.1. I will also have to look at specifying the Node version for GitHub actions.

gwhitney:
The numberscope.colorado.edu server is also at Node v22.13.1, so there won't be any problem there. I will document that frontscope requires v22+ and assuming you can update with no problem and that I can just set the node version in GitHub actions, I think all will be well.

katestange:
> The numberscope.colorado.edu server is also at Node v22.13.1, so there won't be any problem there. I will document that frontscope requires v22+ and assuming you can update with no problem and that I can just set the node version in GitHub actions, I think all will be well.

I updated node and unit tests now pass.  But lint still complains about `ParamField.vue` on my computer and github.

gwhitney:
Yes, I will fix the lint error when I get the e2e tests to pass. Thanks!

katestange:
I did update the snapshots with my commit, and I okay'd a lot of changes because the transversal changes when we mess with the gallery.  But reviewing some of the diffs on github here, I'm worried about some of them.  Should I mark these in a code review?

katestange:
> I did update the e2e snapshots/screenshots with my commit, and I okay'd a lot of changes because the transversal changes when we mess with the gallery. But reviewing some of the diffs on github here, I'm worried about some of them. Should I mark these in a code review?

I think some of it might be because I'm running my own backend here and even though it isn't my first time through the tests, maybe the factoring is too slow or something?  FactorFence for example is showing up blank.

gwhitney:
> I think some of it might be because I'm running my own backend here and even though it isn't my first time through the tests, maybe the factoring is too slow or something? FactorFence for example is showing up blank.

Since we now have the numberscope.colorado.edu updated with its local copy of OEIS metadata, maybe just try with your frontscope set back to getting its sequences from there?

gwhitney:
Ah, I see. For example, your commit set the expected for Divisor Square to an empty canvas. I agree that can't be correct. I will revert that change to the expected in my next commit (still please don't make any more commits on your end for the moment).

gwhitney:
Given the two-fold symmetry inherent in `chroma.scale([<HEX1>, <HEX2>, <HEX1>])` in the Zeta Zero #10143 you checked in , are you sure you still want to double b when indexing (  `floor(number(2b rad, deg) % 360)` )? Try just `floor(number(b rad, deg) % 360)` and see if you like it better and let me know. Also, in the alpha computation, you are intentionally dividing n by 100 so that it saturates after just 80 terms out of 3183 rather than by 1000 so it saturates after 800? Just want to make sure.

gwhitney:
Must have hit the wrong button, sorry! (temporary closure) Anyhow, I've now pushed a commit that has everything pass for me. I expect some of the GitHub snapshots will need updating, and the GitHub action is so far recalcitrant about updating to Node 22. Once I get those things cleared up, I will mark this as ready for review. Even in the meantime, if you would like any of the color changes I suggested for `Zeta Zero #10143` just let me know.

gwhitney:
OK, finally all tests pass and I don't think we are aware of any operational issues. So this is ready for final review. The only points I am aware of are the aspects of the color choice in `Zeta Zero #10143` I raised above, but those are just stylistic aspects for Kate to judge (I am fine either way on them). But I also welcome anyone looking over the now-copious code changes and/or playing with it further and providing any feedback.

Oh and as a bonus in this last commit, once you do npm install, if you have ImageMagick installed on your system, when you do `git diff` it will visually show the differences to `.png` files. Hopefully that will help forestall the kind of change to e2e snapshots that need to be reverted (e.g., you will be presented with the fact that your new snapshot is blank, if that's what's happened).

src/visualizers/Turtle.ts
```diff
@@ -204,53 +307,240 @@ to prevent lag: this speed cannot exceed 1000 steps per frame.
...
     },
     /** md
-- Background color: The color of the visualizer canvas.
-     **/
-    bgColor: {
-        default: '#6b1a1a',
-        type: ParamType.COLOR,
-        displayName: 'Background color',
+- Angle formula: an expression to compute the angle for a given step of the
+  turtle's path. Just as in list mode, the Angle is... parameter
```
katestange, line 406: what's the ...?

katestange: oh, I get it.  Maybe say the `Angle is....` parameter or something (my grammar parser just broke)

gwhitney: How's this? I figured that any typography we applied to offset `Angle is...` should be applied uniformly to all parameter names. But then using all code-quotes was not very easy on the eyes, in part because in mkdocs the code bits come out red. So I did boldface... but it is now a lot of boldface. Would you prefer italics? Let me know if so and I will switch from bold to italics for the Turtle parameter names.

katestange: I like it.

katestange:
Ok, I did some tweaking of the gallery items.  Back to you.  If you like them, then we can update the github snapshots and I think we are good to go!

gwhitney:
> Ok, I did some tweaking of the gallery items. Back to you. If you like them, then we can update the github snapshots and I think we are good to go!

I do like the changes. I made just one tiny tweak of my own: I appreciate your attention to the detail of the accent in Recamán's name. Just using a Unicode string rather than an HTML entity works better, though, since it then shows up the same in the Gallery and in the Specimen Bar when you visit the specimen.

And  I have only one suggestion for you to consider: make the width formula on `Zeta Zero #10143` be `3 - n/1500` or even `4 - n/1000`. Then the trace gets both more intense and sharper, making a nice visual analogy for the function value "coming into focus" as the series converges (well, quasi-converges before it starts to diverge, really, but who's counting?). It also preserves the beauty of the picture that you get by typing 3 extra zeros into the length formula to make it `200000/sqrt(n)` -- with width set at a constant 3, that closeup loses a lot of charm, and its demonstration of the excellent accuracy here, in the blurriness of the strokes. If you like it and want me to handle the work of the update and new snapshots, just let me know your preferred formula.

And we don't need to update any github snapshots, since the WebGL visualizers, including all Turtle ones, simply don't run on GitHub actions -- last time we tried, we couldn't get the WebGL contexts to be set up properly on Github CI. As you can see, all GitHub tests are passing.

We're converging rapidly now ;-)

katestange:
> We're converging rapidly now ;-)

Lovely suggestion on the zeta zero!  I love the blurry coming into focus idea.  How about `4-n/1000` for width and `300/sqrt(n)` for the step/length, which zooms us in a little closer so we are using more of the scren.

gwhitney:
OK, done. There is one hitch though. You definitely need to pull and run the e2e tests locally before merging this PR. That's because the screenshot matching parameters in Playwright were set loose enough so that one of the Zeta Zero tests passed _without_ updating the snapshot, even though the outcome looked obviously different to the naked eye. So I sharpened it by a factor of two, and with that setting, on _my_ machine, the only additional test that failed was the other Zeta Zero test, as desired, and then I updated the snapshot. I checked in the sharper threshold. But we had obviously at some point put in that fuzzier threshold because we needed to because of some legitimate difference in outcomes on different machines/environments. So we need to make sure that you can now still _pass_ all the tests with the sharper threshold.

If you can, I am (personally) fine with your merging this PR. If some of your tests fail, we need to look at which ones and see if we can use a different image comparison method that will do a better job of detecting when things legitimately match and when they don't.

gwhitney:
(This comment can be safely ignore if in fact all e2e tests still pass for you locally.

The reason a different image comparison could help is that right now, the comparator is confused by the vast amount of background in all of our snapshots -- it sees lots and lots of pixels the same and so is prone to thinking they "match". If in fact it only cared about pixels that were not the background color in one image or the other, the comparison would have a much larger effective "dynamic range," and so we'd have a better chance of setting a threshold that would exclude what we feel are non-matches while allowing whatever the variation is between machines that we needed to allow for in the prior fuzz setting.

And it appears we can set up such a custom image matcher, see https://stackoverflow.com/questions/74748149/how-to-implement-comparing-two-screenshots-in-one-test-with-playwright. The application in that post is a bit different, but I think the technique is adaptable to our use case.)

katestange:
All tests (e2e and unit) pass for me on the newest commit!

Issue #539, opened 2025-02-18
Hex with alpha in URL causes Numberscope to complain


katestange:
1. Go to http://localhost:5173/?name=Turtle&viz=Turtle&bgColor=7C4545&seq=Formula
2. Change background colour.
3. The URL now has the new bgColor in it, which is 8 characters long (including alpha)
4. URL reload causes errors in the parameters pane

gwhitney:
I believe the issue is that the function isColor in ParamType.ts is too restrictive -- I think it insists on the # which I think the new picker does not always include (and which I am fine having be optional) and it definitely only allows 4 hex digits, when I think any of 3,4,6, or 8 are fine.

This is a small enough point that I think it can be folded into pretty much any PR for convenience's sake. Anyone who's about to do so should assign this issue to themselves.

gwhitney:
Resolved with merge of #542 into ui2.

Issue #447, opened 2024-08-28
Chaos visualizer overhaul for alpha


gwhitney:
This is one of a series of issues, one for each visualizer, to make sure they are "alpha-ready". The main aspects of this overhaul are:

1. Does the visualizer handle/request canvas size/aspect ratio gracefully and appropriately for the new ui?
2. Is the visualizer pushing as much validation down to the per-parameter checks as possible?
3. Is it doing the appropriate level of resetting/recomputation in presketch vs setup?
4. Is it using descriptions, warnings, and errors appropriately, and not changing the user's input out from under their keystrokes? (Note that if a human interaction entails a change in _another_ parameter -- for example, if you were binning terms and there were 10K terms then setting bin size to 10 might change number of bins to 1000, while changing the number of bins to 100 would change the bin size to100 -- that's OK; we just don't want to seem to be ignoring or contradicting their immediate keystrokes; instead, we inform the person what's at issue with what they entered.
5. Is it sufficiently and appropriately responsive to events and input? Does it handle resizes as gracefully as can be in the current scheme?

In some cases, there are accumulated small previously-existing issues that should be addressed as well.

gwhitney:
For Chaos in particular, the answer to (1) above is no -- for rectangles outside a certain aspect ratio interval around 1, some of the diagram is off-screen.

This should include resolution of #384 and #317.

gwhitney:
Also, it seems to me that under "color dots by index", the "color cycling length" parameter is not having any effect -- unless I am mistaken in what sort of effect it should have.

katestange:
You can cause this to hang by inputting a very large number of corners.

gwhitney:
Likely "alpha factor" should be renamed "Opacity" to match with other visualizers

gwhitney:
And if whoever is doing this really wants to go for it, that person could allow an option for (a) using one of the brewer palettes and/or (b) allowing a user-selected palette with the new COLOR_ARRAY parameter type.

gwhitney:
Completed in ui2 with #561, so closing.

Issue #384, opened 2024-07-08
chaos vis should allow for fading dots


katestange:
Chaos visualizer could let the user configure a gradual fade on the dots, so that the picture has a dynamic quality (not just filling in more and more).

gwhitney:
Just adding on to avoid issue proliferation:

- [ ] There is a bug in the parameter interface in that if you choose "highlight by walker" and then choose an invalid walker in the resulting dropdown, but then choose a different coloring scheme, the error message that you have chosen an invalid walker persists and the visualizer won't draw, even though now that setting is both invisible and irrelevant. When the parameter box is scrolled down so you don't see the error message, it's really hard to figure out this is going on!
- [ ] There is currently no way to make a monochrome image if you have multiple walkers; none of color by walker, index, destination, or highlight walker does it. Suggest adding a monochrome option with a color picker.

gwhitney:
I would say these items should be considered, but optional, for the Chaos overhaul as part of the alpha milestone. So not setting a milestone on this issue.

katestange:
Ok, I spent a bit more time thinking about the fade capability.

Right now there is a variable 'f' for frame, and it is being used as the frame in which the dot is computed.  So you can change colours over time by using 'f' in your colour formula, but it doesn't accomplish a fade, because there's no variable for 'frame you are viewing right now'.  

In order to do a fade, we would need to compute all colours in all frames, and right now we are struggling with lag, so I don't think I'll implement this in the current alpha overhaul PR (which means this issue will stay open).  

But if at some point we want to do this, there are a few ways to go.  Right now we compute and store the dot info (location, colour, etc), adding new dots to storage arrays with each frame.  Then we draw as needed from the stored info.   All fade options besides (0) would require redrawing the whole picture at each frame, but they may require more/less recomputing the stored datta.  Options:

(0) do something artificial like overlaying a partially transparent colour on top of what's there to fade it

(1) allow only lerping from one colour to a "faded" colour, and simply pull up the stored colour and change it according to the number of frames elapsed since it was stored (this would require storing the initial frame value).

(2) allow a new variable named "viewFrame" or something, so you could fade by mixing according to viewFrame - initializationFrame.  This would allow you to do all types of other things too.  In this scenario colours need to be computed afresh every frame using all variables, so you'd have to store everything you can't recreate, which is a fair bit of stuff.

[...]

Issue #317, opened 2024-04-24
Document Chaos visualizer


gwhitney:
It's currently just got a stub.

gwhitney:
Should be part of the Chaos overhaul for alpha.

gwhitney:
Completed in ui2 via #561, closing.

PR #540, opened 2025-02-18
Chaos overhaul

### LINKED ISSUE #518: Index coloring not working in Chaos Visualizer?
It seems to me that in ui2, no matter what "Color cycling length" I enter for the "Color dots by" "Index" option, all of the dots just come out blue... Am I missing something?

katestange:

<hr/>

This should eventually resolve #518 #447 #384 #317 #539

Still to do:

- [ ] update snapshots
- [ ] put images in the documentation
- [x] there's a lag issue (?help?)
- [ ] remove console.logs
- [ ] go through the issues above
- [ ] featured specimens
- [ ] more


gwhitney:
A to-do item: the "label corners of polygon" checkbox does nothing at the moment.

gwhitney:
> 2\. Wait for it to finish drawing -- it will log "Stopping"

I attempted to check this issue out for you, but I never observed "Stopping" in my console, so is perhaps something else awry?

gwhitney:
> > 2. Wait for it to finish drawing -- it will log "Stopping"
> 
> I attempted to check this issue out for you, but I never observed "Stopping" in my console, so is perhaps something else awry?

Oh, I see you commented out most of the code in that stopping section, including the console.log of "Stopping". I am observing the lag on the initial drag. But there's a lot of chatter in the console, and the console doesn't seem to be telling me much about _where_ in the code it's freezing up. In particular, sometimes I see a last line of "Starting" while I am waiting for the lockup to end, but other times the last line is "end drawing loop 10001". And if I add a console.log of "Starting1.7" after "1.5", it seems to change the distribution of what I see. So I expect we're seeing variations in async scheduling and execution. So I am really not sure what we're waiting on before that first drag starts... But I don't think it's particularly happening _at_ that spot in the code between Starting and Starting1.5.

gwhitney:
I am not _at all_ sure I have gotten to the bottom of this, but it seems to me that the first time that Chaos needs to execute a `model()` command on a given chunk, it incurs some processing time penalty -- maybe this is loading geometry information to the GPU?? -- and that penalty depends on the complexity of what is in that geometry object.  I am going to push a commit, that we may want to revert mostly or completely later, that (a) reduces the chunk size to 256; (b) draws squares instead of round dots, because I am assuming they have significantly fewer vertices; and (c) every time a chunk is added, voluntarily redraws the whole thing so it has now used that chunk at least once -- the idea there is to spread out the instants at which a chunk is first used, instead of, on that first drag at the end, drawing every chunk for the first time. With these changes, I think the lag is greatly reduced -- I won't say it's gone. You can give it a whirl and see what you think, and whether you want to investigate more/more deeply, or whether you want to just go ahead with something along these lines.

I don't think the above observations really explain why Chaos seems much harder on my poor computer than Turtle, though -- and I don't think we do anything in Turtle to avoid all of the chunks being used for the first time on the first drag. Those deficiencies are a big part of why I don't think I've really gotten to the bottom of this whole thing...

If you decide to abandon pan, roll, and zoom, I wouldn't cry -- they are regular polygons anyway, so rotating is not too helpful and you can just make the vertices as spread out as possible so the image won't go off the canvas. But having made the switch to WebGL, if you want to keep it that way I am OK with it, too.

gwhitney:
I was glad to see that it Just Worked to take the sierpinski triangle gallery specimen and set the step formula to 1.5 -- produces a very nice picture. Not certain whether or not steps over 1 worked before.

gwhitney:
> I was glad to see that it Just Worked to take the sierpinski triangle gallery specimen and set the step formula to 1.5 -- produces a very nice picture. Not certain whether or not steps over 1 worked before.

Along those lines, this makes a pretty picture: http://localhost:5173/?name=Chaos+Game&viz=Chaos&corners=3&stepFormula=max%281.7-.0001n%2C+0.5%29&seq=Random&max=2&last=13999&length=14000

The length here is chosen so that it doesn't run too long with the "standard" step value of 0.5.

gwhitney:
I think in light of #541 we should just move ahead with this for alpha with an "all-formula" interface, and then after alpha, split it into its operational "pre-visualizer" and a skin that restores some of the options that have been blown away (but which behavior can be reproduced with the proper formulas), like the different coloring schemes. Question: do we need to document what options we want to restore other than via looking at the git history?

gwhitney:
I think once #542 is merged this can be rebased and you can pick work back up on it as your time permits, Kate. I think the ball is in your court on this; I didn't see any outstanding review tasks on my end at the moment, but if there is something you'd like me to do, let me know.

katestange:
Question for the Numberscope mind:  when a formula is out of bounds (e.g. the walker or corner formula outputs a walker or corner that doesn't exist), what should happen, ideally?  Right now it throws a popup but that seems annoying.  Is it possible to have it give an orange warning and carry on ignoring it?

katestange:
Another request for advice.  Right now formulas can break on bigints.  For example, the formula `mod(a,p)` where p is the number of corners, can return something that, when it goes into safeNumber, comes out larger than p.  Since this isn't the user's fault (in the sense that it is reasonable to expect mod(a,p) to be at most p-1), should this be handled in some other way than faulty formulas (users putting corner formula = 10 when there are only 3 corners)?  Or is this avoidable entirely?

katestange:
Another question.  Where are we about formulas?  If I have a formula that should give a corner or walker, should I be manually flooring the result?

gwhitney:
Ugh, this PR has sadly gotten into a state in which the changes GitHub is showing do not seem to correspond to the changes being made in this PR. They seem to include changes from a bunch of other PRs. That state will make it virtually impossible to review. Are you able to get it in sync, maybe with some kind of a rebase, or possibly make a new "clean" PR from ui2 with just the desired changes from this PR?

gwhitney:
> Question for the Numberscope mind: when a formula is out of bounds (e.g. the walker or corner formula outputs a walker or corner that doesn't exist), what should happen, ideally? Right now it throws a popup but that seems annoying. Is it possible to have it give an orange warning and carry on ignoring it?

Turtle in "Rule mode" simply ignores entries that are not one of the values in the Domain, just as if they had not occurred in the sequence. It displays an orange warning if this happens frequently. You could behave analogously in Chaos for the walker and corner formulas.

gwhitney:
And re your questions on Formulas: nailing down the behaviors of formulas to make them really airtight is pretty much waiting on upgrading mathjs, and that is in turn stuck on a number of bugs concerning bigint handling and what formulas should do when the result of an operation lies outside the number system of its operands (e.g. the log of rational number, etc.). So my suggestion would be just to note the offending cases in the "upgrade mathjs" issue #343 and not sweat them for alpha.

katestange:
Superseded by #561 since there were rebase issues as usual.

PR #548, opened 2025-03-18
refactor: Move ShiftCompare to workbench

gwhitney:
By submitting this PR, I am indicating to the Numberscope maintainers that I have read and understood the [contributing guidelines](https://numberscope.colorado.edu/doc/CONTRIBUTING/) and that this PR follows those guidelines to the best of my knowledge. I have also read the [pull request checklist](https://numberscope.colorado.edu/doc/doc/pull-request-checklist/) and followed the instructions therein.

<hr/>

About the smallest/shortest PR ever. Resolves the alpha portion of #429.

PR #549, opened 2025-03-19
refactor: Move ShowFactors to workbench.

### LINKED ISSUE #452: Remove or overhaul ShowFactors for alpha
Is this visualizer now obsolete, given the existence of the FactorFence visualizer just checked in? I would be comfortable with deleting it.

If it is to be kept, here are the overhaul items:

gwhitney:

<hr/>

This is the direct analogue of #548 but for ShowFactors. In this case, I also went through the "overhaul list," but it didn't really amount to anything in the case of ShowFactors because it has no parameters. Also, this time the removal of a visualizer changed the (pseudorandom) "transversal" test enough so that I have to update the CI snapshots before this is ready to review. I will add a message when that is done.

Resolves #452.


gwhitney:
OK, I think all is well here now.

katestange:
Works, passes tests.

PR #555, opened 2025-03-31
fix: check Differences levels early and warn if too many

### LINKED ISSUE #448: Differences visualizer overhaul for alpha

*    Does the visualizer handle/request canvas size/aspect ratio gracefully and appropriately for the new ui?
*    Is the visualizer pushing as much validation down to the per-parameter checks as possible?
*    Is it doing the appropriate level of resetting/recomputation in presketch vs setup?
*    Is it using descriptions, warnings, and errors appropriately, and not changing the user's input out from under their keystrokes? (Note that if a human interaction entails a change in another parameter -- for example, if you were binning terms and there were 10K terms then setting bin size to 10 might change number of bins to 1000, while changing the number of bins to 100 would change the bin size to100 -- that's OK; we just don't want to seem to be ignoring or contradicting their immediate keystrokes; instead, we inform the person what's at issue with what they entered.
*    Is it sufficiently and appropriately responsive to events and input? Does it handle resizes as gracefully as can be in the current scheme?

gwhitney:

<hr/>

This constitutes the Differences overhaul for alpha. So far as I could see, the only relevant point was that the interface was not warning if the actual number of levels used was fewer than the requested number of levels.  This PR fixes that, and it appears to me that otherwise the Differences visualizer conforms to all of the overhaul guidelines listed in #448.

Resolves #448.


katestange:
This appears to work and I agree it conforms.  But in the long run, don't we want to prevent or provide some accommodation for the overlap of numbers that you get when the sequence entries have lots of digits?  Maybe not an alpha issue?  I guess the main question for alpha is:  does the overlap issue make the visualizer seem broken?  If you'd like, I'd be happy to merge this and put off the discussion.

gwhitney:
OK, there wasn't any discussion of new features for alpha, but I agree, it does seem pretty broken on large entries.

Do you see any way to lay out the differences table without overlaps without measuring every difference at every level and doing some sort of recursive fitting thing? The difficulty is that if one row has `0 123456789 0` then you can't lay that row out until you calculate that the differences are `123456789 -123456789` which actually have way more digits than the previous row even though it's only two numbers instead of three, so you have to leave extra space between the two outer 0s  and the inner 12345689 in that first row. Etc., etc. 

And that case then begs the question of whether you want the spacing of the layout to be uniform, i.e. every entry takes up the same amount of horizontal space, or if you want to allow a more "ragged" layout that inserts (say) the minimal amount of extra horizontal space everywhere so that every difference can be centered on the two numbers that give rise to it and yet there end up being no overlaps? (I.e., if the triangle has a section that happens to be all one-digit numbers, can it be laid out more tightly within that section than some other part of the triangle that has all larger numbers?)

The "uniform" layout isn't _so_ bad to perform-- I think you just have to precompute _all_ of the entries of all of the rows, measure each one horizontally, take the largest measure, and then lay out each entry in a field of that width. But for some sequences, e.g. ones that grow slowly at first and then take off, that uniform layout will be fairly wasteful of horizontal space.

Another alternative is to use different font sizes for different magnitude numbers...

If you have a particular behavior/layout algorithm in mind, please let me know. If we can reach a decision of what to implement, happy to implement for alpha so that the product looks minimally "amateurish" on first impressions. I just feel like this triangle of numbers layout is a bit of a minefield, despite seeming simple at first glance. But perhaps you have a clearer vision here than I am mustering.

Note FormulaGrid at the moment does cut the font size down for larger numbers in the cells, but for very large numbers, it does nothing to prevent overlap of the numbers, that will overflow their cells despite the smaller font size.

katestange:
A possibility I can think of that wouldn't alter too much the attractive simplicity of the visualizer:
- make every number of fixed width, e.g. switch to 1.234 x 10^e or similar fixed width notation for long numbers
- when you mouse-over a number, it shows the full number (pop up? or at the bottom of the screen?)
What's the biggest number stored in the OEIS?  How big would the exponent e become?

gwhitney:
> A possibility I can think of that wouldn't alter too much the attractive simplicity of the visualizer:
> 
>     * make every number of fixed width, e.g. switch to 1.234 x 10^e or similar fixed width notation for long numbers
> 
>     * when you mouse-over a number, it shows the full number (pop up? or at the bottom of the screen?)
> 
> 
> What's the biggest number stored in the OEIS? How big would the exponent e become?

https://oeis.org/wiki/Large_numbers suggests that OEIS entries only go up to a thousand digits or so. Hence we could try `-1234e9999` as the format, but that's ten characters. With the current font, on my screen we'd only get about ten numbers across if we always allowed room for ten characters -- not enough to see much of the sequence. So I think we'd still want to compute the longest number that needs displaying and leave only enough room for that, switching to the e-notation at or over 10000000 and at or under -1000000, because they compact down to 1000e4 and -1000e3. This way you don't go over seven characters until 1000e100 or -1000e10. 

If we want to make the cutoffs symmetric as to sign, we could give one fewer digit of mantissa for negative numbers, e.g. -123e100 but 1234e100.  Or we could get esoteric and represent negative numbers like 1234ē100 (see the little minus sign above the e ? ;-)

You could put exponents in a smaller font and eliminate the `e` to save more horizontal space, e.g. `1234_55_` with the 55 in small digits (and no underlines) would be "a number that starts with 1234 and then has 55 more digits after the 4." In this scheme, you again have to decide how to deal with signs.

So I think we would still need to nail down a specific proposal here... 

There's also a weird phenomenon where for some sequences it could be that when you imagine you're going to fit 32 numbers across, the largest you will need is (say) 3 digits, but if you actually allow space for  three digits for every number, you can only fit (say) 25 numbers across, and then it turns if you are only looking at the first 25 numbers,
the largest you will need is two digits, and so then there's space left over, leaving you in a quandary: either you space out too much, or you leave blank space on the right, either of which would look weird.

This discussion also begs the question of scrolling sideways. Seems like ultimately that's a must. Is that something we want to pursue for alpha? Not for alpha but for beta? Later?  If we do go for sideways scrolling, then it seems like we are likely going to want to go irregular layout -- you don't know how many characters you'll need later, so why worry about it now? Just use the number of characters you need for what you are currently showing, and then allow wider numbers once you've scrolled some to the right. Irregular layout also gets rid of the weird phenomenon from the previous paragraph. So maybe we should implement irregular entry lengths now and leave side-scrolling for later? Unfortunately, it seems like the layout algorithm for differing entry lengths is kind of tricky.

All these details seem like a bit of a morass, sorry.

gwhitney:
Other possible features, for now or later:
* a row of sequence indices, just above the actual sequence entries. 
* a legend column on the left: index, entry, diffs, 2nd diffs, 3rd, 4th, etc.
* Color formula for the display of a number (of course, everybody else has got one ;-) Could do sign pattern, or by magnitude, etc.
* Option to control font size
* Option to substitute a dot or something for the number, so that the color carries all the information (and/or possibly the shape)

gwhitney:
We discussed these issues in meeting, and agreed that just the following two changes would be made in this PR for alpha; everything else has been entered in https://github.com/numberscope/frontscope/discussions/524#discussioncomment-12718053.

- [ ] Someone (hopefully @katestange) will add at least one featured specimen for Differences
- [ ] A "Digits displayed" parameter will be added, an integer with a minimum of 2, and every entry will be displayed centered in a rectangle of the same size with room for that many digits and a (short) "negative number sign" (perhaps `U+02D7 ˗ MODIFIER LETTER MINUS SIGN`). Entries that have more digits will be truncated to slightly fewer initial digits followed by an ellipsis (presumably `Unicode Character “…” (U+2026)`) so that they still fit in the allocated space.
- [ ] When the mouse is over an ellipsized number, the full difference will be displayed in a tooltip.

gwhitney:
OK, all this should need is one featured specimen, and then it should be ready for review.

katestange:
The error "Too few sequence entries for 41 rows..." seems incorrect.  If you try to use 41 rows with random integers set to max index Infinity, it still says this.  Is 40 just a hard limit?  If so, why?

katestange:
If you choose 2 digits for the number of digits displayed, there are a lot of "..." and "-..." i.e. we can end up not showing a single digit.  This looks kinda weird, so perhaps 3 should be the minimum.

katestange:
Oh, try this: http://localhost:5173/?name=Differences&viz=Differences&levels=41&digits=2&seq=Random&min=1&max=5
This URL shows both the above issues, and it also shows that we aren't using all Infinity available terms here for some reason.  Why is the space at the right?

katestange:
Ok, we have previously decided that Firefox is just bad at ModFill and Grid, as it sometimes doesn't plot the first few draw() loops worth of boxes.  We weren't able to debug it and it was intermittent, not appearing on a freshly loaded browser.  But it happens on my computer 99% of the time.  Ok... But now this visualizer gives only a completely blank screen in Firefox.  Reloading the tab doesn't help.  Clicking "play" will at least draw the numbers, but on a white background instead of black.  That's much more broken that just missing a few boxes, sadly.

katestange:
For a possible specimen, in keeping with the simplicity theme here, should we just use VF?  http://localhost:5173/?name=OEIS+A000045&viz=Differences&levels=40&seq=OEIS+A000045&last=40&length=41   
It actually has nice behaviour....

gwhitney:
> Ok, we have previously decided that Firefox is just bad at ModFill and Grid, as it sometimes doesn't plot the first few draw() loops worth of boxes. We weren't able to debug it and it was intermittent, not appearing on a freshly loaded browser. But it happens on my computer 99% of the time. Ok... But now this visualizer gives only a completely blank screen in Firefox. Reloading the tab doesn't help. Clicking "play" will at least draw the numbers, but on a white background instead of black. That's much more broken that just missing a few boxes, sadly.

Please post any one specific URL that displays this behavior and I will try again to debug it.

gwhitney:
> Oh, try this: http://localhost:5173/?name=Differences&viz=Differences&levels=41&digits=2&seq=Random&min=1&max=5
This URL shows both the above issues, and it also shows that we aren't using all Infinity available terms here for some reason. Why is the space at the right?

OK, have pushed a commit that should never limit the number of terms to less than will fill the screen, and always shows at least one digit, even if the digits displayed is 2 (it does this by reducing the ellipsis to two dots in the digits displayed = 2 case). How's that working for you?

gwhitney:
OK, and there's the VF differences featured specimen. I titled it "Deja vu Differences" -- we can change it if you'd like. I think this is ready for full/final review.

gwhitney:
Oops, hang on, I have to check in the CI snapshots.

gwhitney:
OK, should be good now.

katestange:
> Please post any one specific URL that displays this behavior and I will try again to debug it.

http://localhost:5173/?name=Differences&viz=Differences&seq=Random

Pretty much any URL will work.

katestange:
> How's that working for you?

That all looks lovely!

katestange:
> I titled it "Deja vu Differences"

I like the name!

src/visualizers/Differences.ts
```diff
@@ -27,7 +27,7 @@ difference of.
 const paramDesc = {
     /** md
 - **Number of rows:** How many rows to produce. _(Positive integer, no larger
-than the number of elements in the sequence')_
+  than the number of elements in the sequence')_
```
katestange, line 30: extra quote at the end should be removed

src/visualizers/Differences.ts
```diff
@@ -38,6 +38,28 @@ than the number of elements in the sequence')_
             if (l < 1n) stat.addError('Need at least one row')
         },
     },
+    /** md
+- **Digits displayed:** The maximum number of digits to display for each
+  sequence entry and difference displayed. If a number to be shown has more
```
katestange, line 43: I found that hard to parse.  How about "The maximum number of digits to show for each number displayed."

doc/visualizer-in-depth.md
```diff
@@ -494,6 +494,22 @@ browser, often with the JavaScript `Event` structure as shown.
...
+As a convenience and to prevent code duplication among Visualizers derived
+from P5Visualizer, the base P5Visualizer class provides a simple informational
+popup/tooltip facility, via the following instance method:
+
+`this.simplePopup('Display me', [atX, atY])`
+
+-   Shows the popup with the given text ('Display me' in this example) at the
+    given X and Y coordinates (here `atX` and `atY`). If the text to display
+    is empty (the default), hides the popup instead; thus, you can simply call
+    `this.simplePopup()` to hide the popup.
```
katestange, line 508: What's the default behaviour?  If the user doesn't call this.simplePopup() at all, what happens?

katestange (changes requested):

gwhitney:
> > Please post any one specific URL that displays this behavior and I will try again to debug it.
> 
> http://localhost:5173/?name=Differences&viz=Differences&seq=Random
> 
> Pretty much any URL will work.

Bummer. On Firefox 136.0.1 on Tumbleweed, I am simply not able to reproduce. Every way I can try to display this URL, it looks perfect: invoking a fresh Firefox on it, loading it into a Firefox that has been running for days, reloading it, opening multiple copies of it, changing the number of digits parameter and changing it back, hitting the re-run button, etc. I just can't get the symptom you are seeing. If you want to do a joint debugging session again at some point, I am game. I am sure there is some specific thing going on in your set up and that with enough time and effort, we could figure it out.

gwhitney:
OK, I think I've addressed the review comments. Thanks!

katestange:
Ok everything is good.  I'm going to merge.  We can leave Firefox for another issue.

PR #561, opened 2025-04-06
Chaos overhaul (new branch)

katestange:

This PR overhauls Chaos, changing it to P5GL and changing the parameters to mostly formulas.

This should eventually resolve https://github.com/numberscope/frontscope/issues/518 https://github.com/numberscope/frontscope/issues/447 https://github.com/numberscope/frontscope/issues/384 https://github.com/numberscope/frontscope/issues/317 https://github.com/numberscope/frontscope/issues/539

Still to do:

- [ ] update snapshots
- [ ] put images in the documentation
- [ ] featured specimens

### DIFF
--- a/src/visualizers/Chaos.ts
+++ b/src/visualizers/Chaos.ts
@@ -1,12 +1,21 @@
 import p5 from 'p5'
+import { markRaw } from 'vue'
 
-import {P5Visualizer, INVALID_COLOR} from './P5Visualizer'
-import {VisualizerExportModule} from './VisualizerInterface'
+import interFont from '@/assets/fonts/inter/Inter-VariableFont_slnt,wght.ttf'
+import { INVALID_COLOR } from './P5Visualizer'
+import { P5GLVisualizer } from './P5GLVisualizer'
+import { VisualizerExportModule } from './VisualizerInterface'
 
-import {math} from '@/shared/math'
-import type {GenericParamDescription, ParamValues} from '@/shared/Paramable'
-import {ParamType} from '@/shared/ParamType'
-import {ValidationStatus} from '@/shared/ValidationStatus'
+import { math, MathFormula, CachingError } from '@/shared/math'
+import type { GenericParamDescription, ParamValues } from '@/shared/Paramable'
+import { ParamType } from '@/shared/ParamType'
+import { ValidationStatus } from '@/shared/ValidationStatus'
+
+//[<img
+//  src="../../assets/img/Chaos/TODO.png"
+//  width=500
+//  style="margin-left: 1em; margin-right: 0.5em"
+///>](../assets/img/Chaos/TODO.png)
 
 /** md
 # Chaos Visualizer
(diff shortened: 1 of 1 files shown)


katestange:
See #540 for previous comments on the development of this.

katestange:
A possible featured specimen to try out:  http://localhost:5173/?name=Phoenix+Rising&viz=Chaos&corners=12&walkers=3&sizeFormula=0.6&colorFormula=rainbow%28w*360%2FW%2B65%29.brighten%281%29&pixelsPerFrame=100&seq=OEIS+A000005

gwhitney:
Three comments so far:
1) We discussed it earlier, but did you want to either add a parameter for the radius of the circle that the corners are on (so that you can conceptually zoom in without having to use the zoom) or start the camera farther out the z-axis and adjust everything else, so that you can zoom in further before you hit the "screen goes blank because I've flown through the plane containing the dots" problem?
2) Also/instead do something to prevent "flying through", such as switch to some geometric sequence of distances to the plane containing the dots, that never flies through?
3) In the screenshot below, you can see that when the corners are labeled, the relationship between the vertex location and where the number is placed does not seem uniform. I think it should be possible to place the numbers more accurately.

Screenshot_2025-04-06_14-35-44

gwhitney:
Ah, and `npm run doc:serve` and then visiting http://127.0.0.1:8000/doc/src/visualizers/Chaos/ is showing some garbled documentation.

gwhitney:
> 3\. In the screenshot below, you can see that when the corners are labeled, the relationship between the vertex location and where the number is placed does not seem uniform.

And when you resize to a very small screen height, numbers to the top are clipped, and the numbers seem much larger than they need to be:

Screenshot_2025-04-06_15-11-57

gwhitney:
The "Divisor Square" specimen does not seem to be updated for the revised visualizer. Perhaps that's implicit in the unchecked "featured specimen" box at the top.

gwhitney:
> http://localhost:5173/?name=Phoenix+Rising&viz=Chaos&corners=12&walkers=3&sizeFormula=0.6&colorFormula=rainbow%28w*360%2FW%2B65%29.brighten%281%29&pixelsPerFrame=100&seq=OEIS+A000005

Does the name refer to the embedded orange-red Sierpinski triangle? It is pretty, but not more so than Divisor Square and it's not clear to me it's worth having two featured specimens based on the same A000005 sequence... and I would say DivisorSquare has a somewhat better "Change Sequence" tableau. Isn't there some Chaos instance showing a bias in the distribution of primes among residue classes to some modulus? Would that make a good specimen?

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+It can depend on:
+
+`n` The index of the entry in the sequence being visualized.
+
+`a` The value of the entry.
+
+`s` The serial number of the step starting from one for the first dot.
+
+`m` The minimum index of the sequence being visualized. Note that the
+above definitions mean that `n`, `s`, and `m` are related by `n = m + s - 1`.
```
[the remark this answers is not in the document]
gwhitney: Oh, and given that you have the overall index, do you want to add a step number for the current walker?

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+            .textSize(textSize)
+
+        // Find locations for labels
+        const cornersLabels = this.chaosWindow(this.radius * this.labelOutset)
+
+        // Draw the labels
+        for (let c = 0; c < this.corners; c++) {
+            const label = cornersLabels[c]
+            // WebGL is warning here, but we HAVE loaded a font back in setup??
+            this.sketch.text(String(c), label.x, label.y)
```
gwhitney, line 466: the font loading is happening asynchronously, so I think it's possible that it hasn't completed by the first time that drawLabels is called,

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+                i++
+            ) {
+                const pos = this.dots[currWalker][i]
+                if (pos.x !== lastPos.x || pos.y !== lastPos.y) {
+                    sketch.fill(this.dotsColors[currWalker][i])
+                    const size = this.dotsSizes[currWalker][i]
+                    //sketch.push()
+                    //sketch.translate(lastPos.x, lastPos.y)
+                    //sketch.circle(pos.x, pos.y, this.dotsSizes[currWalker][i])
+                    //sketch.rect(pos.x, pos.y, size, size)
```
gwhitney, line 497: Maybe time to remove all the commented-out drawing code?

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+                }
+                lastPos = pos
             }
         }
+    }
+    // from p5.js docs examples
+    polygon(x: number, y: number, radius: number) {
+        const angle = this.sketch.TWO_PI / this.sides
+        this.sketch.beginShape()
+        for (let a = 0; a < this.sketch.TWO_PI; a += angle) {
```
gwhitney, line 509: might be nice to use this.sketch.TAU rather that TWO_PI since after all, that's exactly what it is.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+        }
+        return currentLength
     }
 
     draw() {
+        // check if we are zoom/panning, redraw if so
+        // p5 is doing something terrible right here
+        // its terribleness is proportional to the
+        // size of the stored dot data
+        // sometimes image degrades during this pause
```
gwhitney, line 535: Are all of the above comments still true? Or did the fix about creating geometry objects more or less take care of them?

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+                f: currentFrames,
+                c: lastCorner,
+                p: this.corners,
+                w: currWalker,
+                W: this.walkers,
+                x: position.x,
+                y: position.y,
+                A: (n: number | bigint) =>
+                    Number(this.seq.getElement(BigInt(n))),
+            }
```
gwhitney, line 717: Please modify the input object computations to re-use common values, rather than recompute them. That change will greatly aid maintenance of the code when we need to change the code used to compute one of the input variables, for example.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+            this.dots[currWalker].push(position.copy())
+            this.dotsSizes[currWalker].push(math.safeNumber(circSize))
+            this.dotsIndices[currWalker].push(i)
+            this.dotsCorners[currWalker].push(myCorner)
+            if (typeof clr === 'string') {
+                this.dotsColors[currWalker].push(this.sketch.color(clr))
+            } else if (math.isChroma(clr)) {
+                this.dotsColors[currWalker].push(this.sketch.color(clr.hex()))
+            } else
+                this.dotsColors[currWalker].push(this.sketch.color('white'))
```
gwhitney, line 800: You could make this whole thing unconditional by just always using `this.sketch.color(chroma(clr).hex())`. That should work regardless of what type/format `clr` is.

gwhitney (commented):
OK, here's a pass through the docs and code. Given there's a bunch of items, I'll presumably make another pass when these are resolved. (Some may not need any changes.)

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+    'p',
+    'W',
+] as const
 
 const paramDesc = {
+    /** md
+- **Corners**: the number of corners on the polygon.
+There must be at least two.
+If there are n corners, then the corners are numbered 0, 1,
+2, ..., n-1 (for use in the formulas via the `c' variable).
```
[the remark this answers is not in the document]
katestange: Hmm, I don't know if I agree, but I tried to rephrase to make what I meant clearer.

src/visualizers/Chaos.ts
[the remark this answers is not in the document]
katestange: Oh, sorry this was copied from other visualizers and not examined carefully.  This will take a bit more thought.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
+above definitions mean that `n`, `s`, and `m` are related by `n = m + s - 1`.
+
+`M` The Maximum index of the sequence being visualized. Note this value
+may be Infinity for sequences that are defined and can be calculated in
+principle for any index.
+
+'p' The number of corners.
+
+`f` The frame number of this drawing pass. If you use this variable, the
+visualization will be redrawn from the beginning on every frame,
```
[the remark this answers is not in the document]
katestange: Yes this is tangled up in the fade effect aspect.  Will think.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
-        },
+        level: 0,
     },
-    alpha: {
-        default: 0.9,
-        type: ParamType.NUMBER,
-        displayName: 'Alpha',
+
+    /** md
+- **Color formula**: an expression to compute the color of each dot.
```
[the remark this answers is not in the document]
katestange: I'm not sure what order we are likely to want to customize in.  It' snot clear to me colour is not even last.  But Step size feels first to me... will give it some thought.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
         required: false,
+        level: 0,
+    },
+    /** md
+- **Color chooser**: This color picker does not directly control the display.
+Instead, whenever you select a color with it, the corresponding color
+string is inserted in the **Color formula** box.
+**/
+    colorChooser: {
+        default: '#00d0d0',
```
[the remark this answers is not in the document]
katestange: yeah same is probably best

src/visualizers/Chaos.ts
katestange: yes this is definitely happening, do we just leave it as is and remove the comment?

src/visualizers/Chaos.ts
katestange: but wait, if this.sketch.TWO_PI = this.sketch.TAU, why are both defined?  why does it matter?  (for me, TWO_PI is less confusing)

src/visualizers/Chaos.ts
katestange: will make another comment soon

src/visualizers/Chaos.ts
katestange: I'm not sure I understand this comment.  Can you elaborate?  Or maybe it's just where you marked it that I'm finding it confusing.  Do you mean recomputing Number(i - this.seq.first + 1n)?  (This is marked next to the definition of A, which i don't see a way to improve.)

katestange:
I've pushed most of the suggested code tweaks, so you can resolve many of them (if I didn't respond, I expect you to resolve unless I misunderstood the guidance).

katestange:
Also lint passes on my machine, I wonder if my lint is a different version?

katestange:
But finally, I'm a bit frustrated with fade.  After your comment that my suggestion of putting a translucent rectangle over the picture to fade was not "artificial", I added such a feature.  Whether or not it is performant enough to keep it, it has highlighted some real issues going on.  

I'm too tired to keep at it now, so if in the meantime you want to look at this URL:
http://localhost:5173/?name=Random+integers+0+to+9&viz=Chaos&corners=8&walkers=8&sizeFormula=5&pixelsPerFrame=20&fadeEffect=0.05&seq=Random&last=1000&length=1001
then you will see what I mean. With fade, you can clearly see that the chunking is doing something weird, with a "fast initial" phase followed by a slower behaviour with intermittent spasms.  It's also eating up my computer, which should not be happening.

src/visualizers/Chaos.ts
gwhitney: Surely you've read the Tau Manifesto? I personally find it very convincing. I use tau in preference to pi everywhere possible. It is the angle measure of a full circle, and therefore the period of sin and cos. It's really the constant that comes up everywhere in math. The fact that we in effect write 2(tau/2) over and over and over again in math is one of those numerous things that ends up making it forbidding to many people. But you are right, this is not really a big deal. I just like to try to be the change I'd like to see and use the more fundamental constant, but of course TWO_PI is not incorrect, it's just slightly more complicated than necessary. Your code, your call, feel free to change or not. (But I do recommend you read the Tau Manifesto sometime -- maybe it will have some of the effect on you it did on me. But also maybe not -- I had spent my whole life independently wondering before I read it why "fraction of a circle" wasn't the most usual angle measure, so I was very primed for the Manifesto.)

src/visualizers/Chaos.ts
gwhitney: I just mean you compute three different objects for inputs to formulas, that all share most of their property names and values, so the shared parts should only be computed once (for efficiency) and written once (for readability and maintainability). In other words, the three highly overlapping lists are not DRY.

gwhitney:
Hmm, I can't even run now; I get
```
src/visualizers/Chaos.ts:828:64 - error TS18046: 'clr' is of type 'unknown'.

828             this.dotsColors[currWalker].push(this.sketch.color(clr.hex()))
                                                                   ~~~


Found 1 error in src/visualizers/Chaos.ts:828
```

I think you want to manually call `chroma()` on `clr` before calling .hex() on the result of that? That has a better chance of compiling, I think...

src/visualizers/Chaos.ts
katestange: I can't figure out how.

katestange:
I reverted the change that broke compiling.

src/visualizers/Chaos.ts
```diff
@@ -1,429 +1,804 @@
...
-    Walker,
-    Destination,
-    Index,
-    Highlighting_one_walker,
-}
+For non-random sequences, the distribution of dots can pick up information
+about local correlations and overall distribution.
+**/
+
+// the symbols available for the colour and step formulas
```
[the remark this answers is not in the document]
gwhitney: I put comments in the code and then extracted them into the documentation.

src/visualizers/Chaos.ts
gwhitney: (I'll leave this for you to decide which way to go...)

src/visualizers/Chaos.ts
gwhitney: Now each input object incorporates the previous one and then adds the new values.

src/visualizers/Chaos.ts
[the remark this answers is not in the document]
gwhitney: Documented actual behavior, and used `k` rather than `s` to conform with other visualizers.

src/visualizers/Chaos.ts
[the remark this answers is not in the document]
gwhitney: Yes, I ended up agreeing; step comes next, because you need it to compute the _new_ x and y coordinates, and I wanted to add those as variables in color and dot size.

src/visualizers/Chaos.ts
gwhitney: No, I just wait to draw the labels until we have the fonts. This may not interact correctly with fade, so we may still yet have to tweak it.

src/visualizers/Chaos.ts
gwhitney: Wow right yes typescript made this a big pain. I found a way to avoid taking cases. This should improve a lot when we have the "desired type" feature for formula parameters, after the mathjs update.

gwhitney:
> But finally, I'm a bit frustrated with fade. After your comment that my suggestion of putting a translucent rectangle over the picture to fade was not "artificial", I added such a feature. Whether or not it is performant enough to keep it, it has highlighted some real issues going on.
> 
> I'm too tired to keep at it now, so if in the meantime you want to look at this URL: http://localhost:5173/?name=Random+integers+0+to+9&viz=Chaos&corners=8&walkers=8&sizeFormula=5&pixelsPerFrame=20&fadeEffect=0.05&seq=Random&last=1000&length=1001 then you will see what I mean. With fade, you can clearly see that the chunking is doing something weird, with a "fast initial" phase followed by a slower behaviour with intermittent spasms. It's also eating up my computer, which should not be happening.

Yes, I see what you mean. I am tired at the moment too, so I will take another stab when I next get to this. I suspect the issue is that you are drawing the fade overlay every dot, as far as I can see. So it's adding a lot of polygons to the "chunks". Any chance you'd be content with just drawing one overlay per frame? That would "cluster" the fade levels, but I hypothesize it will reduce the processing burden significantly. I will give it a try sometime soon and if it looks OK to me, push it for you to try out and see what you think.

gwhitney:
OK, there were some indexing difficulties that caused the fast initial phase followed by the slow period -- basically once there was at least one chunk, it was redrawing all the dots on every frame.

Now the difficulty is that when it makes the chunks, all of the dots in the chunk are at the same unfaded intensity, and then the chunks are re-drawn and don't get faded. So I will need to

(a) when making a chunk, for each dot, figure out how many times it would be faded, then draw it in its dot color faded that many times, and 

(b) just before displaying each chunk, fade the number of times that would occur during the normal drawing of those dots.

Of course, all of these multiple fadings can be done by precomputing the proper derived color/opacity and fading once -- I just need to figure out the formulas.

gwhitney:
All right, I have put in code for (a) and (b). It turns out there was a complication -- I had implemented the Chroma overlay function incorrectly, and I had used a poor definition of scalar multiplication: if we are going to define `c + d` for colors c and d to mean overlaying d on top of c, then we had  better have `c + 2*d` = `c + d + d` (don't need parens, overlay is associative), or in other words, `2*d == d + d` and `3*d == d + d + d` etc. And then by the usual sorts of tricks like building up the reals from the naturals, you get what `r*d` has to mean for any real number r and color d. OK, so I switched to that definition for scalar multiplication. Then fading k times with color bg is just fading once with color `k*bg`, so it made that part very easy. And the situation has much improved. You can try your test URL above, it now looks totally fine. When you drag, it does temporarily brighten up because it now draws the entire thing in one frame, which means no fading for the last bunch of pixels, but I am not too fussed about that. The very old pixels are still differentially fainter, even when you do drag.

However, I have not quite gotten the fading completely smooth. For example, try this example of just watching three simultaneous more-or-less ordinary random walks on the integer lattice: http://localhost:5173/?name=ThreeStaggered&viz=Chaos&walkers=3&stepFormula=0.02&colorFormula=rainbow%28W*360%2Fh%29&pixelsPerFrame=3&fadeEffect=0.02&seq=Random&max=3  (I recommend you zoom in somewhat)

You will see that it fades out nicely, except that every now and then a bunch of the older stuff gets much lighter all at once. So on average, it seems to fade out about as rapidly as I'd expect, but it's jerky. I think it has to do with the chunking -- maybe I haven't quite gotten my counts of the number of fades in different circumstances consistent, or maybe there's still some issue in the fade formula -- possibly a numerical one, because the scalar multiplication formula does now involve raising the transparency ( = 1 - alpha) to the scalar power. So for a large scalar (lots of fades), we're visually comparing the result of that exponentiation done in floats, to the results of overdrawing that number of times on an eight-bit-per-channel graphics buffer. So maybe rounding issues in overdrawing the fade 30 times to 8-bit accuracy produces a significantly lesser fade than precomputing the effect of 30 fades in full 64-bit IEEE floating-point arithmetic. That would explain why I can't see the effect at larger fade rates, nor when there are more pixels per frame (= less frequent fades, since there is one fade per frame). The more I think about it and experiment, the more likely it seems that is what is going on. For example, if you take your sample URL above (not mine), and let it run to completion, and then drag it a bit but then stop moving the mouse with the button still down, you will see it fades out (because it keeps drawing more frames, and fades once each time, with nothing new being drawn because it has finished the sequence). But this fade out stops with a faint "ghost" of the pattern, and doesn't get any dimmer than that. I suspect because of 8-bit numerical issues in drawing the faint, just slightly opaque overlay over the very dim dots, it ends reaching a fixed point just shy of all channels equal to 0. 

So that would suggest that the occasional discrete "dimmenings" in my example URL are actually good -- the fadeout of really old points is being profitably recalculated with greater accuracy.

So should we call the current situation victory, or pursue even greater smoothness of fading? The only way I can see to achieve that smoothness is to choose a "visibility window" of step serial numbers (based on the fading rate and number of pixelsPerFrame). The idea is that steps earlier than the window have gotten so faint that they don't matter anymore. Then on every frame, you re-draw every dot in the visibility window, with its color pre-scaled using the full floating-point accuracy, so that they fade with greater fidelity. I don't see any way to use chunks in this scheme, so even once we've implemented it (which seems like a fair amount of work), the drawing might just not be performant enough to use practically. So I'd be inclined to leave well enough alone as far as fade goes, unless the lack of smoothness in some cases like the random walks link I gave really undermines  your goals for this feature.

Looking forward to your feedback and when it's next convenient we can regroup and decide what all else needs to happen with this PR to get it merged. Thanks!

katestange:
Random fun one!  http://localhost:5173/?name=OEIS+A007931&viz=Chaos&corners=9&stepFormula=0.16&sizeFormula=2&colorFormula=rainbow%28W*360%2Fh%29&pixelsPerFrame=3&fadeEffect=0.01&showLabels=true&seq=OEIS+A007931

katestange:
Another random fun one!  http://localhost:5173/?name=OEIS+A132131&viz=Chaos&corners=8&walkers=8&stepFormula=0.16&sizeFormula=3&colorFormula=rainbow%28W*360%2Fh%29&pixelsPerFrame=3&fadeEffect=0.02&showLabels=true&seq=OEIS+A132131

gwhitney:
From meeting on Aug 15: We believe that the state of play here is that Glen will attempt to reimplement the fading/drawing by using an invisible higher-precision image buffer and copying it to the display on every frame.

gwhitney:
> If we disable the lag-throttles, and let it lag out hugely, it can get to the point of looking like his results. Of course, he's using something totally different with no zoom and pan.

Since we missed the chance to chat this week, I thought I'd comment briefly in advance of the next get-together, in case its helpful for thought/progress in the meantime.

* We could have a checkbox that turns zoom/pan off and on, with the guidance that the number of points plotted will be limited when they are on. And then in the code, when there is no zoom/pan, we would never make chunks or redraw in any way, except when forced to start again from the beginning, and so just run indefinitely, as long as the visitor wants to wait/watch.
* But even if we do that, is the code fast enough/are the graphics fine-grained enough to get the beautiful images you want? Note that A W-H is not using p5, but rather the direct canvas drawing calls, which might certainly help with performance (on either the speed and/or the level of detail). It wouldn't be too hard to add a "RawCanvasVisualizer" to frontscope in addition to P5Visualizer, but then allowing zoom/pan on that might be a pain, and it would definitely be more trouble than it is worth to switch between p5 and raw canvas depending on an "allow zoom/pan" checkbox :-/
* And yes, I agree, if there are features/controls in A W-H's Chaos game that are missing from our Chaos visualizer, I would be happy to discuss/help getting them into frontscope, but I do think at this point filing post-alpha issues is the way to go, so we can get to an alpha release in a reasonable time frame.

See you soon.

gwhitney:
It occurs to me we could have a formula for the coordinates of the next point that has access to all of the other parameters and the coordinates of the current point and the target index, which defaults to "shrink the specified fraction toward the target indexth vertex". But i would definitely suggest that as a post-alpha issue unless it would help with A W-H images you'd like to be able to do now.

katestange:
Here's a variation that comes in fairly quick and has a nice detail -- a tradeoff.
http://localhost:5173/?name=Heptaflake&viz=Chaos&corners=7&bgColor=1D0E0E03&staticMode=true&cornerFormula=mod%28a%2Bc%2C7%29&stepFormula=%28g*T+%2B+%281-g%29*P%29*%5Bcos%28tau%2F5%0A%29%2C-sin%28tau%2F5%29%3Bsin%28tau%2F5%29%2Ccos%28tau%2F5%29%5D&sizeFormula=0.5&colorFormula=rainbow%28a*360%2F90%2B240%29&colorChooser=1B1896FF&pixelsPerFrame=700&fadeEffect=0.00&showLabels=true&seq=Formula&formula=pickRandom%28%5B1%2C++2%2C+6%2C+7%5D%29&last=100000&length=100001

katestange:
Oh, this is pretty:  http://localhost:5173/?name=Doily&viz=Chaos&corners=11&bgColor=1D0E0E03&staticMode=true&cornerFormula=mod%28a%2Bc%2Cp%29&sizeFormula=0.5&colorFormula=a%3F%235351BEFF%3A%231D5C98FF&colorChooser=5351BEFF&pixelsPerFrame=1000&fadeEffect=0.00&showLabels=true&seq=Formula&formula=pickRandom%28%5B+0%2C3%2C3%2C-3%2C-3%5D%29&last=1000000&length=1000001

katestange:
Hi @gwhitney do you know how to move the fern up to the center of the canvas?  I'd like to feature it.

katestange:
Something strange is happening with Fade.  Try this one:  http://localhost:5173/?name=OEIS+A007931&viz=Chaos&corners=9&walkers=2&bgColor=4E8DA4FF&staticMode=true&eagernessFormula=0.16&sizeFormula=1.5&colorFormula=rainbow%28W*360%2Fh%2B140%29&pixelsPerFrame=10&fadeEffect=0.003&seq=OEIS+A007931
I don't think the background colour should be changing with fade?

gwhitney:
This centers the Barnsley fern pretty well for me: http://localhost:5173/?name=Chaos+Game&viz=Chaos&stepFormula=%28%28%5B0%2C0%3B0%2C.16%5D*%28a%3D%3D0%29+%2B+%5B.85%2C.04%3B-.04%2C.85%5D+*%28a%3E2%29+%2B+%5B.2%2C-.26%3B.23%2C.22%5D*%28a%3D%3D1%29+%2B++%5B-.15%2C.28%3B.26%2C.24%5D*%28a%3D%3D2%29%29*%28P%2B%5B0%2C200%5D%29%2F40+%2B+%5B0%2C1.6%5D*%28a%3D1+or+a%3E2%29+%2B+%5B0%2C.44%5D*%28a%3D%3D2%29%29*40+-+%5B0%2C200%5D&sizeFormula=0.5&colorFormula=%2300D132&colorChooser=00D132FF&seq=Random&max=7

gwhitney (approved):
I think we are finally there :)

PR #568, opened 2025-11-05
FactorHistogram overhaul

### LINKED ISSUE #296: FactorHistogram does not distinguish numbers that have no factorization
Currently, FactorHistogram treats 0, +- 1, and numbers without known factorizations all the same, as having 0 factors. I don't mind 0 and +-1 all being lumped together, but numbers of absolute value 2 or more, the factorizations of which are not known, should be treated differently. They certainly have some number of factors, we just don't know how many. So they should either be removed from the count altogether, or they should be in a separate "unknown" bar or be considered to have "-1" factors, or something along those lines.

### LINKED ISSUE #449: FactorHistogram overhaul for alpha

### LINKED ISSUE #480: Mouse-over effects in Factor Histogram can be activated while dragging parameter tabs
I think the histogram bars' mouse-over effects should always be inactive when you're dragging a parameter tab. The current interface deviates from this in at least two cases.

### Drag-first case

1. View a Factor Histogram visualization.
2. Undock one parameter tab.
3. In the process of undocking the other parameter tab, move the cursor over a few histogram bars.
   * **Actual behavior:** *The histogram bars' mouse-over effects activate*
   * **Expected behavior:** *The histogram bars' mouse-over effects don't activate*

### Mouse-over-first case

1. View a Factor Histogram visualization.
2. Undock a parameter tab and drop it in a place where its title bar partially covers a histogram bar.
3. Mouse over the rectangle where the title bar and the histogram bar intersect
   * *The histogram bar's mouse-over effect activates*
4. Click and drag the parameter tab.
   * **Actual behavior:** *The histogram bar's mouse-over effect stays active*
   * **Expected behavior:** *The histogram bar's mouse-over effect deactivates when you click, and no other mouse-over effects activate while you drag*

gwhitney:

<hr/>
NOTE: This PR is actually on top of #561 (and so really just has a half-dozen commits or so). Therefore, I am marking it as a work in progress, since it should not be merged into ui2 until the Chaos overhaul has been and it has been rebased on the resulting ui2.

Changes in this PR:
* No longer posts a popup when the mouse button is down (#480)
* Correctly positions labels when zooming
* Performs factoring and bin setup in presketch to initiate this potentially lengthy process as early as possible
* Posts a warning about the number of terms used for an infinite sequence
* Avoids blocking the browser when the number of terms is huge
* Renames the Visualizer to FactorHistogram from just Historgram
* Minor documentation update

Resolves #296.
(Well, actually that was already resolved, just making sure we close it.)
Resolves #449.
Resolves #480.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,9 +49,9 @@ jobs:
             - name: End-to-end test
               run: npx playwright test -c e2e/playci.config.ts
 ### Uncomment when you need to re-extract snapshots
-            - name: Extract snapshots
-              if: always()
-              uses: actions/upload-artifact@v4
-              with:
-                  name: ci_actual_snapshots
-                  path: e2e/results/manual/output
+#            - name: Extract snapshots
+#              if: always()
+#              uses: actions/upload-artifact@v4
+#              with:
+#                  name: ci_actual_snapshots
+#                  path: e2e/results/manual/output
--- a/src/visualizers/FactorHistogram.ts
+++ b/src/visualizers/FactorHistogram.ts
@@ -99,7 +99,10 @@ class FactorHistogram extends P5GLVisualizer(paramDesc) {
         this.factorCount = []
         const last = this.endIndex()
         for (let i = this.seq.first; i <= last; i++) {
-            if (i % 10000n === 0n) await yieldExecution()
+            if (i % 10000n === 0n) {
+                this.precomputing = `Collecting values ... ${i} /`
+                await yieldExecution()
+            }
             let counter = 0
             const factors = this.seq.getFactors(i)
             if (factors) {
@@ -356,27 +359,24 @@ class FactorHistogram extends P5GLVisualizer(paramDesc) {
         nTicks = Math.floor(sketch.height / (height * tickHeight))
         const bigTick = nTicks * tickHeight
         const bigTickWidth = sketch.textWidth(bigTick.toString())
-        // Draws the markings on the Y-axis
+        // Draw the markings on the Y-axis
         const tickLeft = yAxisPosition - largeOffsetNumber / 5
         const tickRight = yAxisPosition + largeOffsetNumber / 5
         const rightJustify =
             bigTickWidth < tickLeft * scale - 2 * smallOffsetNumber
         for (let i = 1; i <= nTicks; i++) {
-            // Draws the tick marks
-            let tickY = xAxisHeight - tickHeight * height * i
-            sketch.line(tickLeft, tickY, tickRight, tickY)
-
-            const label = (tickHeight * i).toString()
-            let tickPos = tickRight + smallOffsetNumber
-            if (rightJustify) {
-                const labelWidth = sketch.textWidth(label)
-                tickPos = tickLeft - labelWidth - smallOffsetNumber
-            }
-
+            const tickY = xAxisHeight - tickHeight * height * i
+            const textY = tickY + textHeight / 2.5
             // Avoid placing text that will get cut off
-            tickY += textHeight / 2.5
-            if (tickY > sketch.textAscent()) {
-                this.write(label, tickPos, tickY)
+            if (textY > sketch.textAscent()) {
+                sketch.line(tickLeft, tickY, tickRight, tickY)
+                const label = (tickHeight * i).toString()
+                let tickPos = tickRight + smallOffsetNumber
+                if (rightJustify) {
+                    const labelWidth = sketch.textWidth(label)
+                    tickPos = tickLeft - labelWidth - smallOffsetNumber
+                }
+                this.write(label, tickPos, textY)
             }
         }
 


gwhitney:
OK, this PR is now rebased on ui2 for easier review, showing the correct number of commits and files changed as compared to ui2. I still need to take care of lint and CI tests,  which I will get to as soon as I can but it won't be before Monday afternoon. Nevertheless, Kate/Aaron you should be able to start looking over the files and trying the behavior, let me know if you see things that still need work.

gwhitney:
OK, all tests pass and I believe all of the alpha-level concerns I know of for FactorHistogram (including renaming it from Histogram) are resolved. Ready for final review, thanks!

gwhitney:
> Where is/is not Factor Histogram changed to Histogram? It's still Factor Histogram in the interface and in the URL.

No the change is in the other direction, Histogram -> FactorHistogram, to reflect the specialized nature of this visualizer and leave Histogram open for a general histogrammer.

Vectornaut:
In the documentation, I'd recommend changing "hashing a portion" to "hatching a portion".

gwhitney:
> In the documentation, I'd recommend changing "hashing a portion" to "hatching a portion".

Oh that's interesting. It somehow got into my vocabulary that's called hashing the region of the graph. But web searching I see that hatching is orders of magnitude more common. Will change.

gwhitney:
Changed to draft until I have a chance to change the text in the hover boxes and fix the documentation mistaken word choice (hashing/hatching)

gwhitney:
OK, I just pushed a commit with the modified bar labels. But I have encountered what seems like it might be a more serious issue. I can't quite seem to find a foolproof way to get it to happen, but generally speaking it seems to crop up like this: 

1) Start with a fresh window displaying frontscope.
2) Open the gallery and select Polyfactors.
3) Change the bin width to 2.
4)  Press the "Change Sequence" button
5) Here's typically the first sign of trouble: the sample thumbnails of the different sequence change options come up and display and look ok, but the UI freezes up for a few seconds. I can't scroll the thumbnails or select any.
6) After the UI unfreezes, select the A000045 Virahanka-Fibonacci numbers.
7) Try mousing over the various bars.. If that seems to work OK, press change sequence again, and if it's working ok, choose a different sequence. If that displays OK, switch back to A000045. Switch back and forth a few times.
8) Eventually (sometimes just on the first or second switch) I get bizarre visual behavior: black rectangles appearing in random places on the UI, ghost images of bars or hover boxes, the hover box "erasing" bars as I move the mouse around, and/or weird flickering slightly blurry images.

@katestange @Vectornaut I would be very interested to know if either of you can reproduces such odd behavior. If so and if you have any idea what might be going on, I am all ears. Thanks so much.

gwhitney:
All right, well, I am not able to reproduce the problem myself now, on any browser. Don't know what was going on. Maybe it was actually something running in a different tab. Anyhow, under the timing circumstances, I propose we move on (hard to debug a problem you can't make happen). I have just pushed a commit that fixes the documentation issue Aaron found, and removes some old crufty code that no longer seems to be needed, and fixes the hover box to be within the (very narrow) frustum that the latest version of zoom controls sets up. So I believe this is ready for final review, and merging once it is deemed OK. Will move on to next turtle items as soon as I am able.

katestange:
Zooming out on some examples reveals that there is occasionally one too many tick marks on the vertical axis.  The screenshot below and the link http://localhost:5050/?name=OEIS+A000045&viz=FactorHistogram&seq=OEIS+A000045&last=400&length=401 are an example.
Screenshot from 2025-11-28 21-18-15

.github/workflows/ci.yaml
```diff
@@ -49,9 +49,9 @@ jobs:
...
-#              uses: actions/upload-artifact@v4
-#              with:
-#                  name: ci_actual_snapshots
-#                  path: e2e/results/manual/output
+            - name: Extract snapshots
+              if: always()
+              uses: actions/upload-artifact@v4
+              with:
+                  name: ci_actual_snapshots
+                  path: e2e/results/manual/output
```
katestange: did you forget to comment this back out?

src/visualizers/FactorHistogram.ts
```diff
@@ -101,28 +114,27 @@ class FactorHistogram extends P5GLVisualizer(paramDesc) {
...
+    async presketch(seqChanged: boolean, sizeChanged: boolean) {
+        this.precomputing = 'Evaluating sequence entries ...'
+        await super.presketch(seqChanged, sizeChanged)
+        if (!seqChanged) {
+            this.precomputing = ''
+            return
+        }
+        this.precomputing = 'Calculating sequence values ...'
+        await this.seq.fill(this.endIndex())
+        this.precomputing = 'Factoring ...'
```
katestange: So sometimes it says "Factoring ...-1" on the screen.  For example, if you choose a random sequence and then choose a large upper bound on the random values, it will get into this state in Firefox for quite a while before it gets the histogram drawn.  What does the "-1" mean?

katestange (commented):
Looks good.

.github/workflows/ci.yaml
gwhitney: Yes, thanks for noticing.

PR #571, opened 2025-11-29
Numberglyph overhaul

### LINKED ISSUE #365: url storage colour palette bug with NumberGlyph
To reproduce this bug do:

1. open up a fresh scope and choose NumberGlyph, and Random, with random set between 0 and 4.
2. You'll see an array in grey, cyan and green.
3. copy the URL from the URL bar and open it in a new window
4. You'll see an array in grey, cyan and red.

In other words, the FIRST time you try this, you see green.  All subsequent times using the stored URL you see red.  The colour positions move about every time according to the random seed, but the colour palette seems to just depend on whether it was a fresh setup or a URL cut-n-paste.  Note that colour palettes are not part of the parameters of this visualizer -- they are not user settable.

### LINKED ISSUE #396: NumberGlyph range error
If you take a sequence with exactly 10000 terms (like A00040 at the moment) and run [NumberGlyph on it with the number of terms set to 10000](http://localhost:5173/?name=NumberGlyph&viz=NumberGlyph&n=10000&seq=OEIS+A000040), then right at the end it produces a RangeError that 10001 is not between 1 and 10000. Seems like there must be an OBOE (off-by-one error) somewhere...

### LINKED ISSUE #451: NumberGlyph overhaul for alpha

katestange:

<hr/>

This is the NumberGlyph overhaul PR.  Right now changes include:

- Now uses all screen real-estate.

Resolves #365 (a race condition)

Resolves #396 (a range error)

Resolves #451 (this is the overhaul issue)

Todo:

- [ ] add mouseover to show term values -- WON'T do now (for later sometime)
- [x] add Gallery items
- [x] update snapshots


katestange:
Please give feedback on this draft.  I'm planning to add a mouseover (if it's not a hassle) like in histogram, but otherwise I don't see too much else that needs doing besides gallery/snapshots.  But perhaps there are other issues for this overhaul?

gwhitney:
First impressions:
* I agree there needs to be at least one featured specimen using NumberGlyph, I think there are none at the moment.
* Would there be value to adding back in an option for a square layout (centered in the available space)? Consider that an entirely optional suggestion at your discretion.
* When the sequence length is infinite, the default number of terms used should fill the display area at "not too small" a glyph size. Currently, it seems a bit weird that the chosen length leads the visualization to just stop with clearly more room available for more glyphs.
* There is still something going on with #365: If I do a NumberGlyph on the sequence with formula `n`, and then in that same window switch to the Random sequence from 0 to 4, and then copy that url and paste it into a new window, I get different colors: just white, red, and orange in the original window, but white, red, green, blue in the new window. Is it possible that the palette is not getting cleared out on a sequence change, but just being reused? The visualizer lifecycle graph (http://127.0.0.1:8000/doc/doc/visualizer-in-depth/ when you are running `npm run doc:serve`) is very helpful for figuring out whether the proper resets are in the proper functions...
* Code all looks fine but I may give it one more read-over in the final review.

katestange:
The last commit 1ea1101 is a work-around because there had been some latex in the doc page that was being ignored during compilation.  It's probably best this way (it's in mathjs readable form instead of latex now), but did we used to have latex support that disappeared, and should we check there aren't problems elsewhere?

gwhitney:
Yes LaTeX support somehow died. Just pushed a commit that restores it. When you pull the commit, be sure to remember to `npm run install`.

katestange:
Ok!  Checks pass!  I did make the naturals start from 1.  I am putting off the option of making a perfect square for some later time (is there an issue about that)?  Feel free to review and/or merge.

PR #575, opened 2025-12-03
fix: Exception catching

### LINKED ISSUE #537: Sequence changer barfs a lot if current visualization can't deal with (e.g.) negative numbers
If your current visualizer has a formula in it like "log(n)" that can't deal with negative numbers, or in general throws on certain values, then when you go to change sequence, you get a litany of error popups that is very disruptive and feels "broken". We need some way for that sequence chooser to carry on as best it can when such errors are occurring. #534 may help to some extent, if it is quietly converting bad values to zeros or something like that.

### LINKED ISSUE #569: Exceptions in visualizations may not be caught
A visualizer's draw loop can throw errors that never get caught. To see an example, try the following reproduction procedure.

1. Launch Numberscope on a local development server.
   - To use the link below as-is, you'll need to serve it on port 5173.
2. Load [this specimen](http://localhost:5173/?name=Formula%3A+isPrime%28n%29+%3F+1+%3A+0&viz=Turtle&ruleMode=1&stepFormula=50*log%28a%29&seq=Formula&formula=isPrime%28n%29+%3F+1+%3A+0&first=2).
   - _Log entries that say "Uncaught RangeError" and "Uncaught (in promise) RangeError" appear in the browser console._
   - _This is because the visualizer uses `log(a)` in a formula, and the sequence has zeros among its terms._

gwhitney:

<hr/>

This PR attempts to catch errors in visualization `draw()` calls as well as `setup()`, and then handles such errors by placing a semitransparent overlay above the visualization canvas with a (hopefully helpful) message, rather than modal alert popup boxes. It should be tried with the known procedures for eliciting visualization errors described in issues #569 and #537.

Concerns about this PR:

1. It adds two new functions to src/shared/alertMessage.ts but does not add unit tests for them because I am totally unclear about how to arrange there to be an ambient webpage to manipulate in a vitest test. Should we let this slide because of the push to do the alpha release, or is that a bad precedent and therefore I should take the time to investigate testing DOM manipulation in vitest?
2. The issue #569 prescribes "We need to make sure that depart removes the error overlay as well as the visualizer." This PR makes no effort to perform such removal. On the other hand, I cannot see any negative consequences of not worrying about this removal. I think that is because in practice, any time a visualizer departs an element, in short order the element it was inhabiting is removed from the DOM as well, so the overlay child is eventually garbage collected (and in any case is no longer included in the actually-displayed webpage). Is it therefore OK to just leave the error overlay be as is currently happening, or should we proactively take steps to remove the errorOverlay from its containing element? I think it would be possible to do so; we could make `hasErrorOverlay` return the errorOverlay if one is present, and then in the `depart()` function we could do something like `this.within.removeChild(errorOverlay)`. I just didn't bother since I don't see any consequences for not bothering, but perhaps it's just good hygeine that we should proactively do. I don't really know how to judge.

Also, of course, I welcome feedback (@katestange , @Vectornaut) on how the messages look/are laid out/what their actual text is, etc. Thanks!

Resolves #537.
Resolves #569.

### DIFF
--- a/src/shared/alertMessage.ts
+++ b/src/shared/alertMessage.ts
@@ -24,7 +24,7 @@ export const alertMessage = (error: string | unknown) => {
     const errorMessage =
         'This visualization generated an error.\n\n' + `${error}\n\n`
     const suggestion =
-        'Please see the "Errors" section in the User Guide, '
+        'Please see User Guide > Introduction > Errors, '
         + 'accessible via the Help menu in the top bar. '
     return errorMessage + suggestion
 }
@@ -47,10 +47,12 @@ export function errorOverlay(error: string | unknown, where: HTMLElement) {
     const dangerSpan = document.createElement('span')
     dangerSpan.innerText = '⚠\n'
     dangerSpan.style.fontSize = '60px'
+    dangerSpan.style.paddingLeft = '1em'
     overlay.appendChild(dangerSpan)
-    const errorSpan = document.createElement('span')
-    errorSpan.innerText = alertMessage(error)
-    overlay.appendChild(errorSpan)
+    const errorDiv = document.createElement('div')
+    errorDiv.innerText = alertMessage(error)
+    errorDiv.style.paddingLeft = '1em'
+    overlay.appendChild(errorDiv)
     overlay.classList.add(eoClass)
     where.appendChild(overlay)
 }


katestange:
So, I loaded the URL from #569 and got the error.  Then I clicked on the sequence changer and got the attached screenshot.  Then I clicked on A000005 and despite the thumbnail working, the visualizer gave the same error.  So the thumbnail and scope contradict each other.  Which is correct?

Update:  A clue is provided by the reproduction procedure of @Vectornaut in #537 , where opening the sequence changer and choosing A00005 works fine, then going to a sequence producing an error and back to A00005, the error stays on screen.  So I think it isn't clearing.  In fact, although the circle arrow doesn't do anything, a full page refresh gets the visualizer working again.

Screenshot from 2025-12-03 13-37-17

gwhitney:
Indeed, somehow the current error catching can't recover once it's caught an error, even if the error is fixed by changing parameters. Will investigate and push a fix when I have one.

Vectornaut:
> It adds two new functions to src/shared/alertMessage.ts but does not add unit tests for them because I am totally unclear about how to arrange there to be an ambient webpage to manipulate in a vitest test.

It looks like our tests are [running in happy-dom](https://github.com/gwhitney/frontscope/blob/d30481d7a6c3b48ba7366fb28f1351754395f5c6/etc/vitest.config.ts#L11), so maybe we can use the happy-dom API to create an ambient page? Based on a quick skim of the [happy-dom documentation](https://github.com/capricorn86/happy-dom/wiki/Getting-started#usage), the `Window` class seems like a good candidate.

katestange:
Maybe instead of the last line about reloading and emailing, it could say "See Documentation > Errors for more information".  Then we could expand in the docs on the fact that this could be a result of some choice of parameters, give some examples of the types of things (e.g. log(0)), troubleshooting, and further instructions on how to contact us.

katestange:
Possible aesthetic suggestions:
1. Use numberscope colours of some type?  The orange is a new addition to the palette.  Maybe just white.
2. Make the triangle symbol (which I like) of a size that will generally fill a thumbnail.  That way the thumbnails are simple symbols, but when you go to the page, further text is below explaining the error.

katestange:
> (a) Grit our teeth, figure it out, and get it working for alpha
> (b) File an issue to do it later
> (c) Fuhgeddaboutit.

Haha. :)  I choose (b).  One can also successfully ignore issues, so it includes (c).  Although you could point out an easy cut-n-paste workaround in the "Errors" section of the docs.  You can just "CTRL-A" on the webpage and then paste it into a notepad, and the error message will be included along with all the text of the control panels.  (I was unable to reproduce your version of trying to grab it from outside the canvas area with the mouse, but CTRL-A worked well.)

gwhitney:
> > Did you want expanding the Errors section in the User Guide to be part of this PR?
> 
> A minimal version, yes, so at least there _is_ an Errors section -- it's kind of broken if it points to something not existing? I think a paragraph is fine for now, but then we can expand there if/when needed. What do you think? Maybe under "User Guide" after Visualizers and Sequences, so just "User Guide > Errors"?

There already _is_ such a section, which is why I referred to it. What aspects are missing from it that you would like to see in this PR? Thanks.

gwhitney:
> I choose (b).

Issue #576 filed.

katestange:
> Filling the whole thumbnail seems really jarring to me -- that is huge! I just put it at 60px, and on its own line. That way (at least on my screen), on a thumbnail, you see a nice big danger symbol and then just a text that there was an error, and the error message. The suggestion to see the help is then only visible by clicking or scrolling that thumbnail. Again, let me know what you think.

I like the aesthetic choices now!  I think your larger symbol is clearer what is going on intuitively in the thumbnails, and it isn't so much detail now.  I think it could help to have a small space at the left, so the text and symbol aren't up against the very left side of the canvas.

src/shared/alertMessage.ts
```diff
@@ -7,21 +7,75 @@ aren't working. Instead, we tell them about the error
...
-    const sendEmailDirective =
-        'If this issue persists, please send an '
-        + 'email to numberscope@colorado.edu with steps to reproduce '
-        + 'the error and the error message.'
-    return errorMessage + reloadDirective + sendEmailDirective
+        'This visualization generated an error.\n\n' + `${error}\n\n`
+    const suggestion =
+        'Please see the "Errors" section in the User Guide, '
+        + 'accessible via the Help menu in the top bar. '
+    return errorMessage + suggestion
```
katestange: The "Errors" section is under "Introduction" and I didn't know it was there because this was one more layer down so it isn't visible unless you expand Introduction.  Could we rephrase this?  "Please see User Guide > Introduction > Errors, accessible via the Help menu in the top bar." or something similarly detailed?

katestange:
> OK, edited all that into the existing section, let me know how it reads.

It's great.

katestange:
Looks great and passed all tests on my machine.  Merging.

PR #583, opened 2025-12-18
Featured updates

katestange:

<hr/>

Changes:
- gallery items can now specify a frame limit for thumbnails
- thumbnails stop after timeout or framelimit is reached, whichever is first
- frame limit doesn't apply to specimen itself
- gallery items reordered to create a pleasing mix
- tweaks to Resistance Is Futile, VFib Snowflake, others
- some gallery items removed

This is a draft, but please comment.


katestange:
>     1. Should we put a limit on the Pascal's Triangle so that it never shows any of the region where integer overflow starts to screw up the triangular pattern?

Yes, probably, but I wasn't sure how to address that without it looking kind of unfinished.

> in the color formula, and then the screen blanks with one of our error displays that says "Error: unknown format: 0". I think we should not release with a featured specimen that eventually errors out. 

Agreed.   Is that just limiting the number of terms to fix this?

3. We currently have the same lovely word in two titles: "Lattice Murmurations" and "The Murmurations of the Punctual Bird". Shouldn't we change one? Perhaps "The Meandering of the Punctual Bird" ? 

Agreed.  I like that.

5. Should we try to scale or translate Zeta Zero 10143 so that the path doesn't anywhere go off screen? 5. Same question for PEMDAS Symmetry.

Good idea, I agree.

> When these questions are decided/resolved, I for one am ready to merge this.

If you have some time to do these items, that would be great!

gwhitney:
> > in the color formula, and then the screen blanks with one of our error displays that says "Error: unknown format: 0". I think we should not release with a featured specimen that eventually errors out.
> 
> Agreed. Is that just limiting the number of terms to fix this?

No, you have to replace `A(n+8)` with `A(min(M, n+8))` to prevent an attempt to access a sequence entry that's out of bounds (`M` is set to the maximum index of the sequence). This change actually throws off the last eight points plotted, but I don't notice the difference.

OK, all of the fixes agreed on are committed. Going ahead with CI testing.

PR #585, opened 2026-01-11
fix: Ensure unmounting a thumbnail or a Scope view cleans up the visualizer

### LINKED ISSUE #581: Gallery drag after opening specimen
Right now, Gallery thumbs have a maximum runtime (timeout).

If you set the allowed runtime fairly high on the thumbnails in the gallery, then when you click on a specimen quickly, the remaining gallery items are still tying up CPU, and the specimen will run slowly, until the timer runs out.

Is it possible to make the thumbnails stop when you click on a specimen, so the gallery isn't tying up CPU anymore?

A good one to test this on is the raucus recaman, since it's at the top (easy to click quickly) and should run quickly if you load it from scratch (browser reload).

PS - this makes me worry the specimen continues to run in the background when you load the gallery!  some of them might run forever...

gwhitney:

<hr/>

Previously, references were being held to the HTML containers for visualizers. This would hinder garbage collection and allow visualizers to reawaken and continue to draw in a "zombie" state in which they were not connected to any visible element, draining cpu resources.

The PR removes all such references, preventing frontscope from bogging down as you switch back and forth from the gallery to individual visualizers.

It also fixes a typo in the definition of VFib Snowflake, and patches a bug which allowed some FormulaGrid visualizations to draw forever, rather than stopping once they had displayed everything available.

Resolves #581.

### DIFF
--- a/src/shared/defineFeatured.ts
+++ b/src/shared/defineFeatured.ts
@@ -98,10 +98,12 @@ const featuredSIMs = [
         sequenceKind: 'Formula',
         visualizerQuery:
             'dimensions=200&backgroundColor=000000FF&speed=1024'
-            + '&fillFormula=%7Bcircle%3A%0A++yellow'
-            + '.mix%28black%2C+1-gcd%28x%2Cy%29%2Fsqrt%28x+y%29%29%2C%0A'
-            + 'mouseover%3A+%5B%0A++++%27gcd%28%27%2C+x%2C+%27%2C+%27%2C'
-            + '+y%2C+%27%29+%3D+%27%2C+gcd%28x%2Cy%29%0A%5D%7D&inset=0.8',
+            + '&fillFormula=%7Bcircle%3A%0A++yellow.mix%28black%2C+1'
+            + '-gcd%28A%28x%29%2CA%28y%29%29%2Fsqrt%28A%28x%29A%28y%29%29%29'
+            + '%2C%0Amouseover%3A+%5B%0A++++%27%40%27%2C+x%2C+%27%2C+%27%2C'
+            + '+y%2C+%27%3A+gcd%28%27%2C+A%28x%29%2C+%27%2C+%27%2C+A%28y%29%2C'
+            + '+%27%29+%3D+%27%2C+gcd%28A%28x%29%2CA%28y%29%29%0A%5D%7D'
+            + '&inset=0.8',
         thumbFrames: 60,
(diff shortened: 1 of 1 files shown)


katestange:
First, I noticed a bug in the Integersteller gallery item which I'm hoping can be rolled into this too.  It's not strictly speaking a bug, but it's not the desired thing, either.  If you go to Integersteller and then try to change sequence, the picture doesn't actually change for any other sequence.  That's because the formula has gcd(x,y) in it instead of gcd(A(x),A(y)).  They are the same on the integers, but the latter would adapt the visualization to any other sequence, which is a natural thing the user would like to do when they click "change sequence".  Ooops!  Can we roll it in here?

katestange:
Next up, here's a real bug.  To reproduce:

1. Go to gallery, click on one of the turtle visualizations, say Raucus Recaman.
2. Click on change sequence and make sure you are set to thumbnails.  You see thumbnails for the *default* turtle settings, not for the current visualizer settings, as you should.

So apparently we removed too much of something.  This appears to be specific to Turtle.

As an apparently separate issue, Recaman seems to fail and give a sad face (not a proper error) on the canvas in the background if you leave the change sequence thumbnail modal up for a bit and/or switch windows a bit.  But this seems to happen in ui2 also.

gwhitney:
> Next up, here's a real bug. To reproduce:
> 
>     1. Go to gallery, click on one of the turtle visualizations, say Raucus Recaman.
> 
>     2. Click on change sequence and make sure you are set to thumbnails.  You see thumbnails for the _default_ turtle settings, not for the current visualizer settings, as you should.
> 
> 
> So apparently we removed too much of something. This appears to be specific to Turtle.

Should be fixed with the latest commit. Thanks for the great catch.

katestange:
> > Correct? Or do we want:
> 
> Yes, correct, the first of your two formulas seems the most correct and informative to me.

Oh wait, I misunderstood!  The second formula seems reasonable.  I'll try it.

katestange:
>     * Is there still a concern about Recaman erroring in the background that we want to track down before alpha? If so, I need to know how to reproduce that since I haven't observed it.

Can't reproduce either -- maybe my browser was in a weird state?

>     * Do we want any further tweaks to the new Integerstellar formula?

No, looks good now.

>     * Is there any further checking/review anyone wants to do before this is deployed?

It all looks good to me!!

PR #606, opened 2026-02-02
fix alignment of navbar links

ysulyma:
Before:

image

After:

image


gwhitney:
Yuri, the failures are to be expected with rearranging some layout of the UI; the change looks good and we will add commits with new snapshots as needed.

gwhitney (approved):
Straightforward CSS fix, thanks, Yuri!

gwhitney:
Change deployed at numberscope.colorado.edu, thanks again, Yuri.

PR #622, opened 2026-02-02
Dark mode support

ysulyma:
Basic dark mode support:

https://github.com/user-attachments/assets/f65191b7-d5ef-410d-b563-905117bebe28

- I replaced `--ns-color-white` with `--ns-color-bg` when used as a background color, and `ns-color-black` to `--ns-color-fg` when used as a text color. Other uses of these variables (e.g. borders) remain intact

- Uses the [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/light-dark) function. This has [87% browser support](https://caniuse.com/?search=light-dark) compared to [96% support](https://caniuse.com/?search=prefers-color-scheme) for [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme), but has the advantage of colocating the variants; if desired, I can change it to use `prefers-color-scheme` instead.

- the SVG text also changes color

- I didn't provide a dark variant for `--ns-color-primary` since it's the brand color and I don't know if it needs to align with other things


gwhitney:
Not a full review yet, but this feature looks reasonable. One preliminary question: what does `light-dark()` do when it's _not_ supported?

ysulyma:
@gwhitney good catch! I could do

```css
--ns-color-fg: #2f2f2f;
--ns-color-fg: light-dark(#2f2f2f, #ffffff);
```

for backwards compatibility, but I decided to just go with `prefers-color-scheme` for better compatibility. (There are only five values here so colocation isn't as much of an issue)

PR #623, opened 2026-02-02
Replace Prettier with Biome

ysulyma:
[Biome](https://biomejs.dev) is significantly faster: the current `npm run lint` script takes about 6 seconds to run, whereas `biome check` takes around 0.2 seconds. The slowness is more due to ESLint than Prettier, but swapping out ESLint is a much bigger task belonging in a separate PR.

Draft PR because I don't know how to update the scripts in `tools/`. I think you should basically be able to replace those scripts with `biome check --fix` though.

There was one minor formatting change I couldn't get it to ignore, otherwise I matched the current Prettier configuration.


gwhitney:
Thanks for bringing biome to our attention! It's not a formatter I, for one, had ever heard of. I don't know that the six seconds -> 0.2 seconds improvement is by itself enough of a motivation for us to switch formatters, since the whole/only point of a formatter is to avoid bikeshedding on formatting and the regular contributors are all pretty used to the status quo. So what I am going to do is close this PR, and create a Discussion about whether we would like to switch to Biome, which you're welcome to weigh in on. And if it appears that you are becoming a regular contributor to frontscope, that would up the weight given to your wanting to switch; whereas if it turns out you happen to end up being more of a brief visitor, then it may be less crucial whether we switch. (So of course your thoughts/inclinations as to which is more likely are relevant here, but we're not pushing you to figure that out in any rush. And of course if your feeling is you're likely to contribute semi-regularly if we use a formatter you can tolerate ;-) and unlikely if not, you can share that feeling and we'll take it into consideration, too.)

PR #625, opened 2026-02-03
fix help menu gap

ysulyma:
Fix the help menu gap introduced in #606 (see https://github.com/numberscope/frontscope/pull/620#issuecomment-3837436008)

Here I have highlighted the hover surface in red: the gap causes the help popup to vanish:

hover surface small

I expanded the hover surface to fix this (using negative margins to prevent any other layout shift); I also expanded the horizontal hover surface so that you don't fall off when moving diagonally (left+down) to the popup:

hover surface big

### DIFF
--- a/src/views/minor/NavBar.vue
+++ b/src/views/minor/NavBar.vue
@@ -161,7 +161,7 @@
                 padding-left: 0.5em;
 
                 padding-bottom: var(--popup-offset);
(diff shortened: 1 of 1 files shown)


gwhitney (approved):
With the tweak to slightly overlap the link and the popup to truly eliminate any danger zone in which the popup disappears, this looks good, thanks! Merging.

